PR #80, opened 2014-02-11
Don't determine python-format flag implicitly

JonathanRRogers:
This extends the extractor interface as discussed in https://github.com/mitsuhiko/babel/issues/35

### DIFF
--- a/babel/messages/catalog.py
+++ b/babel/messages/catalog.py
@@ -99,10 +99,7 @@ class Message(object):
         self.string = string
         self.locations = list(distinct(locations))
         self.flags = set(flags)
-        if id and self.python_format:
-            self.flags.add('python-format')
-        else:
-            self.flags.discard('python-format')
+
         self.auto_comments = list(distinct(auto_comments))
         self.user_comments = list(distinct(user_comments))
         if isinstance(previous_id, string_types):
@@ -111,6 +108,13 @@ class Message(object):
             self.previous_id = list(previous_id)
         self.lineno = lineno
         self.context = context
+
+    def determine_python_format(self):
+        """Sets python-format flag if message contains a format string"""
+        if self.id and self.python_format:
+            self.flags.add('python-format')
+        else:
+            self.flags.discard('python-format')
 
     def __repr__(self):
         return '<%s %r (flags: %r)>' % (type(self).__name__, self.id,
--- a/babel/messages/extract.py
+++ b/babel/messages/extract.py
@@ -67,7 +67,7 @@ def extract_from_dir(dirname=None, method_map=DEFAULT_MAPPING,
     """Extract messages from any source files found in the given directory.
 
     This function generates tuples of the form ``(filename, lineno, message,
-    comments, context)``.
+    comments, context, flags)``.
 
     Which extraction method is used per file is determined by the `method_map`
     parameter, which maps extended glob patterns to extraction method names.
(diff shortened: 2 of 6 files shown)


sils1297:
Commit message: please add reason. See http://coala.readthedocs.org/en/latest/Getting_Involved/Writing_Good_Commits/#how-to-write-good-commit-messages for more info.

Please rebase, should fix the tests.

sils1297:
@JonathanRRogers still around? Please let us know if you need any assistance!

sils1297:
@JonathanRRogers would you mind removing the merge commits and doing a rebase (`git rebase --interactive` will do the whole trick) instead? Also the test coverage is not sufficient for this to be merged. We only want to see your commits in the PR so we don't mix up anything making review complicated and error prone.

sils1297:
@JonathanRRogers from llooking shortly at the commits now, e.g. the first one needs an explanation in the commit message so it is clear what this is, also note that no commit should break the build because each of them needs to be revertible later individually. So you probably want to squash your "Fix tests" bug with the one where it belongs to.

sils1297:
oh and please ping me when you add new commits, I don't check all PRs regularly, a comment will summon me :)

JonathanRRogers:
Though I have been using git for years, I have never needed to rebase and I haven't used GitHub much. I looked around GitHub for documentation relating to rebase to no avail. I ran "git rebase --interactive" and it had no effect. Perhaps you can point me to an example of what you want exactly.

I have added a test to cover the extract function interface change I made and all the checks have passed.

sils1297:
@JonathanRRogers GitHub isn't really rebase friendly. If you need it depends on what workflow you are using. Using rebase has the advantage of making the commit history more maintenance friendly, you don't have all those merge commits cluttering it. This eases review and thus helps reducing bugs because a merge commit is close to non-reviewable.

You'll want to pull the current master first from this repository, then checkout your branch and do a `git rebase --interactive master`, git will show you a document in your editor where you can specify for each commit what you want to do with it (edit, squash, reword commit message) and if you remove a line you remove that commit.

If you need anything else, please contact me simply on https://gitter.im/python-babel/babel , I'm always happy to help :)

sils1297:
Test coverage looks good now, thanks.

sils1297:
and here's a tutorial: https://www.atlassian.com/git/tutorials/rewriting-history/git-rebase-i

JonathanRRogers:
If anyone's actually interested in this enhancement, let me know.

akx:
@JonathanRRogers Hey, sorry for the delay with the patch :crying_cat_face: 

This looks mostly like a nice patch, though adding an additional return value to extractors should probably be dealt with due care. In particular, all Babel-internal code paths should be checked to accept both 5-tuples and 6-tuples from extractors. (Looks like you've done that already, though?)

Even so, I'm afraid that external code that bypasses `extract_from_dir` might expect only 5-tuples from Babel-internal (or [other...](https://github.com/mitsuhiko/jinja2/blob/master/jinja2/ext.py#L448)) extractors, which _might_ mean that adding a sixth retval to extractors is a semver major change. That's not to say making a semver major change is bad; it's just another big step, and if we do something like that, it would be worth it to do other semver major changes in the same go.

For instance, formalizing the output interface for extractors from just regular tuples to namedtuples might be a worthy step to do at that point.

All that said, if you could rebase this patch on top of the current `master`, we could continue the discussion? :)

JonathanRRogers:
I'm sorry I've neglected this pull request for so long. I have rebased it again.

PR #84, opened 2014-03-05
Add support for loading data files from inside zips or wheels.

xlevus:
I've added some compatibility functions for loading the data files from inside zips or wheels.

This should make the package zipimportable, and easier to deploy as a zip to app-engine.

### DIFF
--- a/babel/core.py
+++ b/babel/core.py
@@ -53,7 +53,7 @@ def get_global(key):
         filename = os.path.join(dirname, 'global.dat')
         try:
             fileobj = zipopen(filename, 'rb')
-        except IOError, e:
+        except IOError as e:
             _raise_no_data_error()
         try:
             _global_data = pickle.load(fileobj)


sils1297:
@xlevus This needs to be rebased at least, also commit seperation doesn't seem sane, what's a merge commit doing there?

Ping me if you need any support.

babel/_compat.py
```diff
@@ -54,3 +56,74 @@
...
+
+
+def _zip_split(path):
+    for ext in ('zip', 'whl'):
+        if '.' + ext + os.sep in path:
+            return ext
+    return None
+
+
+def extract_nested_zipfile(path, parent_zip=None):
```
erickwilder: Can you add some unit tests for those new functions you're adding to the project?

erickwilder:
Any chance of this pull request conflict with the work done at https://github.com/mitsuhiko/babel/pull/198?

lipis:
yes.. maybe in August 2016 :)

sils1297:
@xlevus @lipis oh we didn't do that speech here: Babel was unmaintained for quite some time, we just took over maintenance of this module and there's a lot to work up - however I can more or less promise you that contributions from now on will be valued instead of ignored. Thus I think we can get this done within this year ;)

There seems to be some overlap with https://github.com/mitsuhiko/babel/pull/198 which seems a bit cleaner to me as pointed out by @erickwilder , maybe it makes sense get #198 merged and look into what functionality this PR provides on top eventually creating a new PR after the merge.

lipis:
:) I'm glad that the project is back on track.. didn't mean to insult anyone.. plus I understand that the whole project is based on someones free time.. so whenever :) The sooner the better though..

sils1297:
@lipis no offense taken :) Just making sure everybody understands what was and is going on.

sils1297:
@xlevus do you need any assistance?

joernhees:
i guess that merge is caused by me making a PR against his PR for py3 compatible exception syntax... from my point of view it can just be squashed away... JUST DO IT :heavy_check_mark:  ;)

xlevus:
@sils1297 I could use some help with some free time :stuck_out_tongue:

lipis:
and if @xlevus is not responsive.. maybe someone else could do a new PR with all the changes?

sils1297:
@xlevus heh, indeed I know that too. @lipis @joernhees if you want to take over this PR please say so explicitly and wait for @xlevus approval as he seems to be reading this stuff timely at least :)

lipis:
@xlevus is here :) So I guess all we need is to resolve conflicts and squash..

xlevus:
Ok, I've rebased it against master, Tests run, but I still haven't written any extra tests for the zip/wheel loading in `babel/_compat.py`.

lipis:
All green :)

sils1297:
> All green :)

Except the fact that codecov didn't submit it's status... I wonder why that is. Thing is this PR is not well tested, i.e. reduces code coverage which is not allowed so by that rule this would still need improvement.

xlevus:
I can't really put any more time into this I'm sorry as I currently don't use Babel in any of my projects.

sils1297:
@xlevus thanks for submitting. Anyone interested in taking over? Otherwise it might take some time until we actually find enough time to actually do code, we're all in our freetime here :)

sils1297:
> we're all in our freetime here

(I guess though)

lipis:
So we are almost in August 2016... :) @sils1297 or someone that is actually the owner of this project can fix or merge this thing?

sils1297:
@lipis feel free to take this over. Possibly @akx might be willing to help as well if there is specific demand for it.

lipis:
@sils1297 If I could take over I would take over no need to mention that in every comment.. I would guess the owners of this project should just fix the conflicts and actually take over!! The demand is obvious since 2014..

sils1297:
I cannot take this over personally right now. The original owner of this project is no longer actively around and we, the current maintainers if you want to call us that, are doing the best we can to keep it alive with very limited time available. We are open to suggestions but we can't fix everything there is demand for on our own. Believe it or not, before we were around it was even worse and you wouldn't even get a response for a year on anything.

If we're lucky @akx or @jtwang will pick this up.

lipis:
Everything is understandable and I know how the open source works.. the reason I want that is because I'm using it in another open source project and I just wanted to update to the latest one since it looks like at least there are quite a few recent releases! and I'm using the version from 2014 because I need the zip thingy..

akx:
@lipis There's a tentative plan to scrap the `.dat` files entirely and generate locale data as regular importable Python modules for Babel 3. Would that suit your scenario? (Actually, speaking of which, why do you need ZIP support?)

lipis:
Because I'm zipping all the libraries when uploading to Google App Engine..
https://github.com/gae-init/gae-init-babel/blob/master/requirements.txt#L8

The initial magic was done by @joernhees and maybe he can comment if that will help not.. (by guess is that, without the dat files it will work out of the box.. but I have no clue)

joernhees:
The connection between gae-init and this issue is that for size and file-count-limit (as in inodes) reasons the libs in appengine should be zipped. Luckily, python supports loading libs from folders as well as zip files (the zip file can contain a bundle of all libs you're using). Sadly, without this patch babel can't be loaded as a lib from such a bundle, because it assumes that it's loaded from a folder and not a zip file. It then tries to load its `.dat` files from the `.../lib.zip/.../*.dat` file, which the OS correctly refuses. So this patch just makes babel also support that kind of library loading.

The magic in gae-init actually happens here https://github.com/gae-init/gae-init-babel/blob/master/main/appengine_config.py#L7 by modifying the path to include the lib-bundle zip file...

https://github.com/gae-init/gae-init-babel/blob/master/requirements.txt#L8 is actually just directly importing this patch which i forked here: https://github.com/joernhees/babel/tree/zipcompat and built...

If the `.dat` files became proper submodules and are imported like other python files, then python's own module-loading-magic would take care of handling the fact that we're in a zip-file. As i'm not sure how big a change that would be and how much time will pass until then, i'd still greatly stand in favor of someone clicking "merge" here and removing the magic once it's solved in a different way.

Testing this properly would need a different environment in which babel as lib is loaded from a zip file while all its tests are run as usual. As the tests here don't fail it shows that the normal folder based lib loading isn't obstructed. As gae-init-babel works with this patch, i'm not sure if we really need to go that extra mile...

lipis:
Any chance for this PR?!

PR #95, opened 2014-05-09
Trying to improve odict performance

artms:
Odict __setitem__ method is linearly searching if item is present in keys list, which significantly reduces performance on big dict


sils1297:
@artms interesting, can you rebase the PR and add a little more information to the commit message?

artms:
I'm gonna create new pull request, I deleted babel some time ago from my list of forked repositories

PR #96, opened 2014-05-20
read in text mode, not binary

anru:
Read text files as binary it generally bad idea (and write too, but now using wb is shortest way for avoid problems while writing encoded text in python3). For example it affects regular expr. matching in third-party extractors for whitespace in windows platform.

### DIFF
--- a/babel/messages/frontend.py
+++ b/babel/messages/frontend.py
@@ -128,7 +128,7 @@ class compile_catalog(Command):
 
         for idx, (locale, po_file) in enumerate(po_files):
             mo_file = mo_files[idx]
-            infile = open(po_file, 'rb')
+            infile = open(po_file, 'r')
             try:
                 catalog = read_po(infile, locale)
             finally:
@@ -439,7 +439,7 @@ class init_catalog(Command):
         log.info('creating catalog %r based on %r', self.output_file,
                  self.input_file)
 
-        infile = open(self.input_file, 'rb')
+        infile = open(self.input_file, 'r')
         try:
             # Although reading from the catalog template, read_po must be fed
             # the locale in order to correctly calculate plurals
@@ -554,7 +554,7 @@ class update_catalog(Command):
         if not domain:
             domain = os.path.splitext(os.path.basename(self.input_file))[0]
 
-        infile = open(self.input_file, 'rb')
+        infile = open(self.input_file, 'r')
         try:
             template = read_po(infile)
         finally:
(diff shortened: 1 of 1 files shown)


anru:
@mitsuhiko what do you think?

sils1297:
@anru hey, could you rebase this PR onto current master? I think this should fix the tests.

sils1297:
And thanks for taking the time to submit a PR! Sorry that it took so long for you to get an answer. Maintenance is changing right now so this'll hopefully get better.

anru:
Rebased

sils1297:
@anru thanks, we'll review it shortly!

sils1297:
Marking as pending review because this needs maintainer attention.

jun66j5:
I think that changes would lead `UnicodeDecodeError` if default encoding isn't utf-8, e.g. Windows, Unix with LANG=C.

```
C:\usr\src\trac\trac.git>C:\venv\py33-amd64\Scripts\python.exe setup.py compile_catalog -l ja
running compile_catalog
  File "setup.py", line 167, in <module>
    **extra
  File "C:\usr\apps\python33-amd64\Lib\distutils\core.py", line 148, in setup
    dist.run_commands()
  File "C:\usr\apps\python33-amd64\Lib\distutils\dist.py", line 930, in run_commands
    self.run_command(cmd)
  File "C:\usr\apps\python33-amd64\Lib\distutils\dist.py", line 949, in run_command
    cmd_obj.run()
  File "C:\venv\py33-amd64\lib\site-packages\babel\messages\frontend.py", line 133, in run
    catalog = read_po(infile, locale)
  File "C:\venv\py33-amd64\lib\site-packages\babel\messages\pofile.py", line 204, in read_po
    for lineno, line in enumerate(fileobj.readlines()):
UnicodeDecodeError: 'cp932' codec can't decode byte 0x83 in position 968: illegal multibyte sequence
```

Makman2:
We had the same problems in coala, if you output messages to console you should decode them with `sys.stdout.encoding` with setting the `errors="replace"` flag.
`open` has a builtin parameter where you can specify encoding and error handling, if you need to output strings you need to do `the_string.encode(sys.stdout.encoding, errors="replace").decode(sys.stdout.encoding)`.

akx:
Closing this as it hasn't been worked upon in 2ish years and the current behavior doesn't seem to cause issues as far as I'm aware.

PR #100, opened 2014-06-12
Add __copy__ and __deepcopy__ to LazyProxy.

astaric:
When copy.copy or copy.deepcopy is used to copy LazyProxy, __init__ is never called and self._value is never set to None. Accessing it in self.value (which is used in __getattr__) triggers another call to __getattr__, resulting in endless recursion.

### DIFF
--- a/babel/support.py
+++ b/babel/support.py
@@ -263,6 +263,23 @@ class LazyProxy(object):
     def __setitem__(self, key, value):
         self.value[key] = value
 
+    def __copy__(self):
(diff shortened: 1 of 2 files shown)


sils1297:
cool, can you rebase? This should fix the tests.

astaric:
import copy on top and then use copy.deepcopy in method?

sils1297:
Yes please :)

sils1297:
@astaric ok, no need to change then :)

PR #103, opened 2014-07-04
Fix incorrect test_fractional_seconds

afedotov89:
In this test were not considered significant zeros in microseconds


sils1297:
was this broken on master? If it was it probably isn't anymore because master builds now. Is this PR still needed? If yes please rebase.

sils1297:
and reopen in that case

afedotov89:
https://github.com/mitsuhiko/babel/pull/191

PR #117, opened 2014-11-05
Add support for parent locale exceptions

etanol:
The method to generate the name for the parent locale has some exceptions.  Notably the latin america regions (``es_MX``, ``es_CO``, ``es_AR``, etc...) which should inherit from ``es_419`` instead of ``es``.  Fortunately, these exceptions are described in the <localeParents> element of the supplemental data document.

This pull request tries to implement the proper locale inheritance for such cases.

There is an ugly hack because one of the commits introduces a circular dependency between ``babel.core`` and ``babel.locale``.

### DIFF
--- a/babel/core.py
+++ b/babel/core.py
@@ -60,6 +60,8 @@ def get_global(key):
             fileobj.close()
     return _global_data.get(key, {})
 
+localedata.get_global = get_global
+
 
 LOCALE_ALIASES = {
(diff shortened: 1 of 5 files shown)


sils1297:
@etanol reopen if you have a special reason for keeping this one alive.

etanol:
Superseded by #180 since GitHub's UI does not allow to change the base/origin branch.

eduardolujan:
Hi @etanol about the problem with es_MX what is the stable version for (#117) that solves that problem.


Best regards,
Eduardo

etanol:
[2.1.1](https://github.com/python-babel/babel/releases/tag/2.1.1) onwards

PR #119, opened 2014-11-12
Fix odict's pop method

regisb:
`odict.pop` method does not remove the key from the `_keys` attribute if
the default argument is not specified.

### DIFF
--- a/babel/util.py
+++ b/babel/util.py
@@ -199,12 +199,15 @@ class odict(dict):
         return self._keys[:]
 
     def pop(self, key, default=missing):
-        if default is missing:
-            return dict.pop(self, key)
-        elif key not in self:
-            return default
-        self._keys.remove(key)
-        return dict.pop(self, key, default)
+        try:
+            value = dict.pop(self, key)
+            self._keys.remove(key)
+            return value
+        except KeyError as e:
+            if default == missing:
+                raise e
+            else:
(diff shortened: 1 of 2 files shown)


sils1297:
This needs a rebase which will probably fix the tests.

Can you adjust your commit message so it explains the change and not the fix? Is there any issue? Please place a reference to the issue describing the problem. If there is no issue, could you create one?

More information about a good commit message style is available on http://coala.readthedocs.org/en/latest/Getting_Involved/Writing_Good_Commits/#how-to-write-good-commit-messages .

regisb:
I rebased this commit and amended the message. The corresponding issue was created: https://github.com/mitsuhiko/babel/issues/196
Please don't hesitate to comment on the wording, both of the issue and the commit message.

erickwilder:
Hi @regisb 
Tests are failing on both Python 3.3 and 3.4. Can you review you code and avoid breaking on those versions? Using [pyenv][#pyenv] can be very helpful to test against all required Python versions at once.

Let me know if you need some help here. Thank you for your contribution.

[#pyenv]: https://github.com/yyuu/pyenv

regisb:
@erickwilder,

I just updated this PR to make tests pass with python > 3 and improve the code coverage. Let me know if there is any way I can still improve this PR.

PR #122, opened 2014-11-22
Added territory-language import from CLDR.

erikr:
Available in the global data onder the 'territory_language' key. For example, for NL this looks like:

```
{'en': {'official_status': None, 'population_percent': 74.0},
 'fy': {'official_status': 'official_regional', 'population_percent': 4.3},
 'gos': {'official_status': None, 'population_percent': 3.6},
 'id': {'official_status': None, 'population_percent': 1.8},
 'li': {'official_status': None, 'population_percent': 5.5},
 'nl': {'official_status': 'official', 'population_percent': 100.0},
 'rif': {'official_status': None, 'population_percent': 1.2},
 'tr': {'official_status': None, 'population_percent': 1.2},
 'zea': {'official_status': None, 'population_percent': 1.4}
}
```

Source from the CLDR is: http://www.unicode.org/cldr/charts/latest/supplemental/territory_language_information.html

### DIFF
--- a/scripts/import_cldr.py
+++ b/scripts/import_cldr.py
@@ -145,6 +145,7 @@ def main():
         territory_currencies = global_data.setdefault('territory_currencies', {})
         parent_exceptions = global_data.setdefault('parent_exceptions', {})
         currency_fractions = global_data.setdefault('currency_fractions', {})
+        territory_languages = global_data.setdefault('territory_languages', {})
 
         # create auxiliary zone->territory map from the windows zones (we don't set
         # the 'zones_territories' map directly here, because there are some zones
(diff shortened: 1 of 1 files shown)


sils1297:
looks good on first sight. Rebase to make tests pass and then we'll see the coverage.

sils1297:
assigning @etanol for review of CLDR related PR :)

etanol:
Wouldn't this need some kind of API functionality?  I am not sure if the `global.dat` file is intended to be directly exposed, because its structure is not guaranteed to be stable.  Any future changes to it would then break users' code.

akx:
Well, `get_global()` is a public API as it is...

akx:
This was closed via #315.

PR #131, opened 2014-12-25
FIX: pybabel update

iho:
FIX: pybabel update -i translations/messages.pot -d translations


sils1297:
tests pass - maybe a rebase already helps against that

akx:
@iho If you're still around, mind explaining why this is necessary? In what situation (platform, input file(s)) can one encounter the problematic behavior that this fixes?

akx:
Closing as expired.

PR #140, opened 2015-01-26
Allow multiple arguments for same keyword

st4lk:
[xgettext](https://www.gnu.org/savannah-checkouts/gnu/gettext/manual/html_node/xgettext-Invocation.html#Language-specific-options) allows to have different arguments for same keyword, whereas babel doesn't. This pull request fix it.

Example
-----------
some_file.py contents:
```
print _("Hello")
print _("Event", "Events", 5)
```

- output of `xgettext some_file.py --language=Python --output-dir=locale --keyword=_ --keyword=_:1,2`:    
```
#: some_file.py:1
msgid "Hello"
msgstr ""

#: some_file.py:2
msgid "Event"
msgid_plural "Events"
msgstr[0] ""
msgstr[1] ""
```
- output of `pybabel extract ./ --output=locale/messages.pot --keyword=_ --keyword=_:1,2`    
```
#: some_file.py:1
msgid "Hello"
msgstr ""

#: some_file.py:2
msgid "Event"
msgstr ""
```

As we can see, xgettext correctly found plural argument list for `_("Event", "Events", 5)`. But babel treats function `_` to have only one argument, ignoring option `--keyword=_:1,2` and therefore assumes, that "Event" has no plural forms.

P.S.
In [tornado](http://www.tornadoweb.org/en/stable/locale.html?highlight=locale#module-tornado.locale) web framework, function `_` can take both the single string and a string with plural form and with count. I found it useful, as we don't need to have another translate function name for such case.

### DIFF
--- a/babel/messages/extract.py
+++ b/babel/messages/extract.py
@@ -21,7 +21,7 @@ import os
 import sys
 from tokenize import generate_tokens, COMMENT, NAME, OP, STRING
 
-from babel.util import parse_encoding, pathmatch, relpath
+from babel.util import parse_encoding, pathmatch, relpath, len_recurse
 from babel._compat import PY2, text_type
 from textwrap import dedent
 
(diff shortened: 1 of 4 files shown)


sils1297:
hey @st4lk can you rebase this? Should fix the tests.

st4lk:
I've done the rebase now.

babel/messages/extract.py
```diff
@@ -29,15 +29,15 @@
 GROUP_NAME = 'babel.extractors'
 
 DEFAULT_KEYWORDS = {
```
akx: Changing the format of the keywords dict should be documented.

In addition, it's a backward-incompatible change for client apps that pass `keywords` as code, so it should probably support the old format as well as the new one.

PR #143, opened 2015-02-16
Respect currency digit settings.

karlspalding:
The supplemental currency data should be respected when
formatting currencies. In particular this means some
currencies should always have a certain number of digits
after the decimal point.

http://www.unicode.org/reports/tr35/tr35-numbers.html#Supplemental_Currency_Data

Conflicts:
	scripts/import_cldr.py
	tests/test_numbers.py

Conflicts:
	scripts/import_cldr.py
	tests/test_numbers.py

### DIFF
--- a/babel/numbers.py
+++ b/babel/numbers.py
@@ -278,7 +278,13 @@ def format_currency(number, currency, format=None, locale=LC_NUMERIC):
     locale = Locale.parse(locale)
     if not format:
         format = locale.currency_formats.get(format)
+
+    fractions = get_global('currency_fractions')
+    digits = fractions.get(currency, fractions['DEFAULT'])['digits']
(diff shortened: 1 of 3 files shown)


dahlia:
+1 this.

sils1297:
@karlspalding hey, thanks for submiting this PR! Looks good on first sight, maybe a rebase onto current master solves the failing CI, could you do that?

Second, could you remove the "Conflict" paragraphs from your commit message?

Ping me if you need any help.

etanol:
Duplicates #127.  Even though this implementation is not fully backwards compatible, the idea is exactly the same.

etanol:
I'm going to close this, since #127 addresses the same issue.

PR #146, opened 2015-03-05
Introduce PyInstaller support

wodo:
Direct access to `__file__` is not possible inside a PyInstaller  bundle.

### DIFF
--- a/babel/core.py
+++ b/babel/core.py
@@ -10,6 +10,7 @@
 """
 
 import os
+import sys
 
 from babel import localedata
 from babel._compat import pickle, string_types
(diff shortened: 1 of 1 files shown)


sils1297:
@wodo can you rebase this PR? This should fix the tests.

wodo:
Two years later. It is still a problem to use python-babel in a PyInstaller environment. Direct access to __file__ is not possible inside a PyInstaller bundle. The current version of python-babel has now two locations for this access.

* babel/localedata.py
* babel/core.py

I will provide a better pull request.

PR #153, opened 2015-03-31
Add __hash__ to Locale.

thanatos:
The `Locale` class has `__eq__`, but not `__hash__`. The absence of `__hash__` results in some rather maddening results when trying to use it as a key to `dict`s, and there's no reason not to `__hash__`…

### DIFF
--- a/babel/core.py
+++ b/babel/core.py
@@ -326,6 +326,9 @@ class Locale(object):
     def __ne__(self, other):
         return not self.__eq__(other)
 
+    def __hash__(self):
+        return hash((self.language, self.territory, self.script, self.variant))
+
     def __repr__(self):
         parameters = ['']
         for key in ('territory', 'script', 'variant'):
--- a/tests/test_core.py
+++ b/tests/test_core.py
@@ -57,6 +57,11 @@ def test_get_global():
 
 class TestLocaleClass:
 
+    def test_hash(self):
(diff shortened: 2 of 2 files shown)


sils1297:
@thanatos hey, thanks for submitting this PR! Could you rebase? This will probably fix CI tests.

thanatos:
@sils1297 Rebased!

sils1297:
@thanatos many thanks! I dropped you a comment right on the commit, should be easy to correct. In addition I'd rewrite the commit message to `Locale: Add __hash__`, without trailing period and with the tag, see http://coala.readthedocs.org/en/latest/Getting_Involved/Writing_Good_Commits/#how-to-write-good-commit-messages for reasons. (Not that your commit message is bad or unacceptable, we're not *that* picky here :))

akx:
@sils1297: I think this is fine for merging; we can add the negative test case in a subsequent PR.

Better to merge than to leave this hanging in a PR limbo :)

sils1297:
@akx would you mind amending this commit, adding the testcase, making a new PR? Be sure to keep the original author and mention your co-authorship in the commit message, I think that's the best way to handle attribution here.

akx:
@sils1297 Done as #303.

sils1297:
awesome, will merge that!

PR #164, opened 2015-06-19
Don't set tag_date by default, to avoid the date ending up in the version number

ColdrickSotK:
I want to be able to build and install Babel with bit-for-bit reproducibility, as part of [the Baserock project's](http://wiki.baserock.org/projects/deterministic-builds/) work on the subject. Including the build date in the version number was preventing Babel from being bit-for-bit reproducible.

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -1,6 +1,3 @@
-[egg_info]
-tag_date = true
-
 [aliases]
 release = egg_info -RDb ''
 


erickwilder:
@ColdrickSotK I was looking at the Baserock project to better understand what you're looking for. I'll make some tests without the date in the building process and see if we have no regressions.
Can you update your pull request to the latest `master` and/or `2.x-maintenance` branches? We still have some failing tests here.

sils1297:
Will be done by https://github.com/python-babel/babel/pull/259 , many thanks for submission though!

PR #167, opened 2015-07-28
Update locale.rst

jeremydw:
Fixes a minor typo.

### DIFF
--- a/docs/locale.rst
+++ b/docs/locale.rst
@@ -109,7 +109,7 @@ want.  You can also ask for the information in parts:
     u'Alemanha'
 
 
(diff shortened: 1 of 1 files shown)


sils1297:
@erickwilder Since we have not dicussed about merging and reviews I'm not doing merges on my own, your call.

erickwilder:
:+1:
@sils1297 is always better to have as much input as we can, but this can slow things. 

I think we can merge pull requests like this. I'll just keep an eye on PR with lots of changes and if not sure about merging, ask for @mitsuhiko's opinion.

Issue #109, opened 2014-09-03
ImportWarning: Not importing directory '.../babel/localedata': missing __init__.py


quinox:
(This is the same as [#156](http://babel.edgewall.org/ticket/156) from the old tracker.)

Babel shows an import-warning when you've configured your warning settings. It can be fixed by renaming the folder ''localedata'' or renaming the file ''localedata.py'', this way the code is explicit in what to import.

To reproduce:

```python
cd /tmp

virtualenv virtual_156

. virtual_156/bin/activate

pip install babel
Downloading/unpacking babel
  Downloading Babel-1.3.tar.gz (3.4MB): 3.4MB downloaded
  ...

python -W error

Python 2.7.6 (default, Mar 22 2014, 22:59:56) 
[GCC 4.8.2] on linux2
Type "help", "copyright", "credits" or "license" for more information.
>>> import babel
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/tmp/virtual_156/local/lib/python2.7/site-packages/babel/__init__.py", line 20, in <module>
    from babel.core import UnknownLocaleError, Locale, default_locale, \
  File "/tmp/virtual_156/local/lib/python2.7/site-packages/babel/core.py", line 14, in <module>
    from babel import localedata
ImportWarning: Not importing directory '/tmp/virtual_156/local/lib/python2.7/site-packages/babel/localedata': missing __init__.py
>>> 
```

rasca:
+1 same happens here

hector:
+1 happening to me also
If you guys want to preserve the name for the localedata.py and the localedata folder, just:
```mv localedata.py localedata/__init__.py```

sils1297:
This was fixed on master with commit https://github.com/mitsuhiko/babel/commit/7d387f95f4f41ed646bcb2b4fa39dae4a601b6ce .

We're currently adressing some maintenance relevant CI, backporting and release issues so it might take a few weeks until we can make some kind of release out of that. This situation will hopefully improve in the future with automated development releases done on merge, see https://github.com/mitsuhiko/babel/issues/194 for more information.

PR #160, opened 2015-06-02
Fix of ImportWarning. Issue #109

philiptzou:
Current version will throw out an ImportWarning since some script automatically created `localedata/` folder. This fixed the problem.

PR #177, opened 2015-08-01
Fixed issue #109: `ImportWarning` warned when `import babel`

philiptzou:
`localedata.py` uses the same name with folder `localedata/`, which
Python will try to load `localdata/__init__.py` and warn `ImportWarning`
if the file not found.

If try to filter the `ImportWarning` as an error like this:

```python
import warnings
warnings.filterwarnings('error', category=ImportWarning)
import babel
```

An `ImportWarning` exception will be raised.


philiptzou:
@sils1297: I create this pull request instead of updating #160 since I can't modify voleg/babel.

philiptzou:
@sils1297 The missed newline fixed

sils1297:
@philiptzou cool, thanks! I'll get this reviewed by tommorrow.

sils1297:
@philiptzou looks good to me on first sight. If I might suggest a new commit message:

```
localedata.py: Move to localedata/__init__.py

`localedata.py` uses the same name with folder `localedata/`, which
Python will try to load `localdata/__init__.py` and warn `ImportWarning`
if the file not found.

Fixes https://github.com/mitsuhiko/babel/issues/109
```

@erickwilder can you take a look if this violates anything existent in this project? I'm not that much into it yet. Also we need to agree on commit message guidelines at some point.

sils1297:
@philiptzou also, if you have a few minutes left - could you add a test for this?

sils1297:
@philiptzou sorry for this inconvenience: we have added major changes to the continuous integration services. Can you please rebase so they get triggered?

erickwilder:
@sils1297
It's ok for me. Whenever possible we need to add tests. @philiptzou can you add some?

sils1297:
Ok @philiptzou following stuff needs to be done:

1. Rebase onto current master.
2. Change commit message. (Describe change, not only bug number.)
3. Add a test (in an own commit) that would have failed before your commit.
4. Push it up.
5. We'll take a last look and merge it :)

philiptzou:
To be honest I don't think a unit test is required. This pull-request is just moved `babel/localedata.py` → `babel/localedata/__init__.py` and correspondingly changed the `_dirname` value. If there's an error or failure, the `babel.localedata` module would not be imported properly, and an `ImportError` would be likely raised. But this doesn't happen to every exist test. So we can say this changes has already been covered by other exist tests.

erickwilder:
@philiptzou ok. I'm fine with this. Can we merge @sils1297?

PR #180, opened 2015-08-03
Last attempt to support parent locale exceptions

### LINKED ISSUE #97: es_MX/MXN: format_currency using comma instead dot for decimal separator
Hello. I did various tests and found that format_currency is formatting MXN currency (from Mexico, es_MX locale) wrong. There is a small code that compare it to python locale, that use posix data:

```python
In [1]: import locale

In [2]: from babel.numbers import format_currency, Locale

In [3]: locale.setlocale(locale.LC_ALL, 'es_MX.UTF-8')
Out[3]: 'es_MX.UTF-8'

In [4]: locale.localeconv()['int_curr_symbol']
Out[4]: 'MXN '

In [5]: locale.localeconv()['currency_symbol']
Out[5]: '$'

In [6]: locale.localeconv()['decimal_point']
Out[6]: '.'

In [7]: locale.currency(50)
Out[7]: '$ 50.00'

In [8]: format_currency(50, currency='MXN', locale=Locale('es_MX'))
Out[8]: u'50,00\xa0$'
```
In my tests I updated CLDR data, from 23.1 to 25 and same issue happens.

This can be a issue in CLDR data?

etanol:
The method to generate the name for the parent locale has some exceptions. Notably the latin america regions (es_MX, es_CO, es_AR, etc...) which should inherit from es_419 instead of es. Fortunately, these exceptions are described in the element of the supplemental data document.

This pull request tries to implement the proper locale inheritance for such cases.

~~There is an ugly hack because one of the commits introduces a circular dependency between `babel.core` and `babel.locale`.~~

**NOTE:** This is a new pull request that should supersede #117 and fix #97.

### DIFF
--- a/babel/localedata/__init__.py
+++ b/babel/localedata/__init__.py
@@ -81,11 +81,14 @@ def load(name, merge_inherited=True):
             if name == 'root' or not merge_inherited:
                 data = {}
             else:
-                parts = name.split('_')
-                if len(parts) == 1:
-                    parent = 'root'
-                else:
-                    parent = '_'.join(parts[:-1])
+                from babel.core import get_global
+                parent = get_global('parent_exceptions').get(name)
+                if not parent:
+                    parts = name.split('_')
+                    if len(parts) == 1:
+                        parent = 'root'
+                    else:
+                        parent = '_'.join(parts[:-1])
                 data = load(parent).copy()
             filename = os.path.join(_dirname, '%s.dat' % name)
             fileobj = open(filename, 'rb')
--- a/babel/numbers.py
+++ b/babel/numbers.py
@@ -258,7 +258,7 @@ def format_currency(number, currency, format=None, locale=LC_NUMERIC):
     >>> format_currency(1099.98, 'USD', locale='en_US')
     u'$1,099.98'
     >>> format_currency(1099.98, 'USD', locale='es_CO')
-    u'1.099,98\\xa0US$'
+    u'US$1.099,98'
     >>> format_currency(1099.98, 'EUR', locale='de_DE')
     u'1.099,98\\xa0\\u20ac'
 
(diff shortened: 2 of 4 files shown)


sils1297:
hey, could you add `Fixes <full bug URL>` to the commit message where appropriate?

etanol:
Changed.

As commented, I'm not really sure about how to break the circle dependency between `babel.core` and `babel.localedata` without some heavy modifications.

erickwilder:
I'll review it soon

sils1297:
@etanol sorry for this inconvenience: we have added major changes to the continuous integration services. Can you please rebase so they get triggered?

etanol:
Done.

babel/core.py
```diff
@@ -60,6 +60,9 @@ def get_global(key):
             fileobj.close()
     return _global_data.get(key, {})
 
+# FIXME: Very ugly hack in order to break cyclic import dependency
+localedata.get_global = get_global
```
erickwilder: Could we add the import statement inside `babel.localedate.load()` instead?

PS: I didn't tested it, just an idea.

sils1297: Hey, tbh a cyclic import smells. I don't have time to look into details of this one right now but usually it shows that a refactoring is needed.

sils1297: with that I'm saying we probably want to file a bug about this as this is a code architectural issue and not in scope for this PR.

etanol: Yes, there is an architectural issue.  But @erickwilder is right anyway, using an import inside `babel.localedata.load` is less magic and works just fine for this case.  I'll push a correction right away.

erickwilder:
@etanol 
Can you update you branch with our `master`, there's some conflicts here.

erickwilder:
It's ok for me. 
@sils1297 any addition to this?

sils1297:
For me too.

@rultor merge please

sils1297:
just thinking... rultor has no access yet... merging manually then.

etanol:
Thanks :smiley:

PR #182, opened 2015-08-04
travis: Submit coverage to codecov

sils1297:
This will make us able to monitor test coverage more closely to make babel more stable.

@mitsuhiko I can't activate codecov for this repository, you need to do that with the admin access. You should be able to log in at https://codecov.io/ and activate it for this repository, the scripts in this PR will do the rest. Takes 30 seconds.

@erickwilder as we agreed on we probably want to enforce a certain patch coverage or so. That needs to be discussed and @mitsuhiko needs to put it in the UI then which is easy.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,13 +9,15 @@ python:
 
 install:
   - pip install --upgrade pip
-  - pip install pytest
+  - pip install pytest pytest-cov
   - pip install --editable .
 
 # Use travis docker infrastructure for greater speed
 sudo: false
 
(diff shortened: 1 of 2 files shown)


sils1297:
ah that already seems to work: https://codecov.io/github/mitsuhiko/babel?ref=bfecb0ef6d457fe52423d86732350dadcf8dfe74

sils1297:
will be superseeded soon by https://github.com/mitsuhiko/babel/pull/183

PR #184, opened 2015-08-04
gitignore: Add temporary editor files

sils1297:
.idea stores settings for PyCharm, ~ files are temporary files e.g. used
by gedit. .swp is used by vim. We don't want any of those to be
accidentally committed in the repo.


erickwilder:
Ok to merge

PR #186, opened 2015-08-04
Add rultor configuration

sils1297:
This allows merging via github comments. The comment `@rultor merge`
will execute the script (which currently doesn't do anything) and if it
succeeds will perform the merge.

I plan to use rultor later for:
 * Automatic deployment to PyPI (development releases directly from
   master, we do this already in coala, see
   https://github.com/coala-analyzer/coala/blob/master/.rultor.yml)
 * Automatic releasing with deployment to PyPI.
 * Veryfy that all CI services pass before merging (see
   https://github.com/yegor256/rultor/issues/869)

With this any manual pushes to master are disallowed, all pushes to
master have to be validated by continuous integration and reviewed by a
non-committer.

### DIFF
--- a/.rultor.yml
+++ b/.rultor.yml
@@ -1,0 +1,2 @@
+merge:
+  script: echo "Nothing to do (yet)."


erickwilder:
@sils1297 
Travis builds are taking so long to complete. Do you know if there's some configuration to be fixed?

sils1297:
@erickwilder travis guys have problems with their mac infrastructure, all mac builds (also coala and coala/PyPrint) are affected. See https://www.traviscistatus.com/ for more info.

.rultor.yml
```diff
@@ -0,0 +1,2 @@
+merge:
+  script: echo "Nothing to do (yet)."
```
erickwilder: It worth adding a configuration that does nothing? Maybe you can add a simple (but useful) script here?
This is not a blocking, just a suggestion.

sils1297: planning to do that later as described, no time right now. We could try
getting the tests to work in rultor which is probably easy and a good pre
merge validation as the CI services don't revalidate a PR when master
changes.

This has btw. the nice side effect that you can do rultor merges from your
phone through github comments ;)

2015-08-05 15:51 GMT+02:00 Erick Wilder <notifications@github.com>:

PR #197, opened 2015-08-06
Backport CI settings to 2.x

erickwilder:
This pull request adds near the same CI settings found at the `master` branch. The only difference is that I removed any Python 3.4 related settings. Official 3.4 support was added (and I think it should remain) in the upcoming 3.0 release.

### DIFF
--- a/.ci/appveyor.yml
+++ b/.ci/appveyor.yml
@@ -1,0 +1,45 @@
+# From https://github.com/ogrisel/python-appveyor-demo/blob/master/appveyor.yml
+
+environment:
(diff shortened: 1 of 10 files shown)


erickwilder:
@sils1297 
Can you take a quick look here?

sils1297:
@erickwilder you got your review. Its a bit nitpicking but those comments would be addressed to suit my personal quality expectations. I know they may be higher than average. We should have a talk and a document about all that IMO to make sure we both have the same expectations to commits.

PR #200, opened 2015-08-07
Improve odict performance by making key search O(1)

artms:
Odict setitem method is linearly searching if item is present in keys list, which significantly reduces performance on big dict. Make it use hash.

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -8,6 +8,8 @@ Version 3.0
 
 - Upgraded data to CLDR 26
 - Add official support for Python 3.4
+- Improved odict performance which is used during localization file
+  build, should improve compilation time for large projects
 
 Version 2.0
 -----------
--- a/babel/util.py
+++ b/babel/util.py
@@ -172,8 +172,9 @@ class odict(dict):
         self._keys.remove(key)
 
     def __setitem__(self, key, item):
+        new_key = key not in self
         dict.__setitem__(self, key, item)
-        if key not in self._keys:
+        if new_key:
             self._keys.append(key)
 
     def __iter__(self):


etanol:
Although the rough performance improvement of this change is around 28% for key insertion, I think it is a good chance to define what Python versions Babel is going to support.

If it is going to be 2.7 onwards, I think you could drop this class entirely and use the standard `collections.OrderedDict` which behaves much more similarly to a standard `dict`.  It's key insertion performance is a little bit worse than now, but I don't think it matters because, AFAIK, it is not being used in hot code paths.

artms:
Well python 2.7 will stay for long and extra code to strip out odict might not be worth the effort...

etanol:
@artms Precisely, 2.7 *has* `OrderedDict` in its standard library.  So if Babel is *not* going to support older Python versions such as 2.6, 2.5 or 2.4, then ripping of `odict` *is* worth the effort by being forward compatible and more idiomatic.

artms:
oh, (facepalm) about OrderedDict, yeap - then decision is needed which python is lowest babel supports.
@sils1297  there was a big performance difference for us, babel was compiling dictionaries for ~10mins before change and it went down to less than a minute after adding that small optimization.

artms:
@sils1297 - it is fast, non scientific measurement, but as much as `list.__contains__` is optimized it is still significantly slower than hash at lookup:
List:
```
l=range(100000)
for i in xrange(10000):
    l.__contains(99999)
```
30 seconds
vs
```
l=range(100000)
h={}
for i in l:
    h[i]=True

for i in xrange(10000):
    h.__contains__(99999)
```
100ms
It is very biased because it tries to lookup worst case key (the one at the end of the list), but it shows how much performance is degrading and time difference is an order of magnitude

sils1297:
@artms from what I see in the time I have available now (I'm sorry but I don't have the time to fully dive into this until september) we'll want to merge your patch now. Many thanks for your time and your little study! Please know that it s greatly appreciated. Only needed change from my side is the commit message as indicated because git will cut the end off and it will be rather useless then.

@erickwilder can you please review this one and merge if it is fine? I can't review it properly but it *looks* good to me especially as @artms has gone to great lengths to prove its usefulness. This is a performance fix with no functional change and should thus be backported to the 2.0 maintenance branch as well if applying cleanly, just cherry-pick it in.

@erickwilder @mitsuhiko we might want to think about dropping python 2.6 support because of the lack of OrderedDict. I'm neither the one to make this call alone not fully informed about how many people still use 2.6 but my guess would be that this number is rather low. If this grows into a longer discussion we should use a new github issue for that. If we decide to drop this we'll want to open a bug about replacing this with the OrderedDict object reducing the maintenance effort.

erickwilder:
It's good to me too. Ready to merge

@sils1297 
I do think that discussing the minimum Python version to be supported must be done in another issue and any decision should be fully supported by @mitsuhiko

erickwilder:
@artms 
Can you add an entry to the `CHANGES` file to shortly describe what you've done? 
This way we can keep track of what's the contents of our next release easily.

Thank's again for you contribution. :smile:

artms:
Updated changelog and cut commit message

sils1297:
@erickwilder lgtm, partial approved :)

PR #210, opened 2015-09-10
Add format_timedelta(format='narrow') support

moreati:
e.g.
```python
>>> format_timedelta(timedelta(hours=3), format='short', locale='en')
u'3 hrs'
>>> format_timedelta(timedelta(hours=3), format='narrow', locale='en')
u'3h'
>>> print format_timedelta(timedelta(hours=3), format='short', locale='gu_IN')
3 કલાક
>>> print format_timedelta(timedelta(hours=3), format='narrow', locale='gu_IN')
3 ક
```

### DIFF
--- a/babel/dates.py
+++ b/babel/dates.py
@@ -739,6 +739,13 @@ def format_timedelta(delta, granularity='second', threshold=.85,
     >>> format_timedelta(timedelta(hours=-1), add_direction=True, locale='en')
     u'1 hour ago'
 
+    The format parameter controls how compact or wide the presentation is:
+
+    >>> format_timedelta(timedelta(hours=3), format='short', locale='en')
+    u'3 hrs'
+    >>> format_timedelta(timedelta(hours=3), format='narrow', locale='en')
+    u'3h'
+
     :param delta: a ``timedelta`` object representing the time difference to
                   format, or the delta in seconds as an `int` value
     :param granularity: determines the smallest unit that should be displayed,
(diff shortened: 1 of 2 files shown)


sils1297:
@moreati could you see to it that our tests pass (and then that your code is covered by it)? The checks have to pass in order for us to review and merge this PR.

moreati:
Sorry for that. Last minute, late night tweak. Thought I'd run it through tox.

sils1297:
@moreati could you squash the commits? Both represent one change together and are easier to maintain (blame, revert, bisect...) when squashed.

sils1297:
@moreati ping me if you need any assistance.

sils1297:
@moreati many thanks! Will merge after CI approval. This won't be backported as it's a new feature.

moreati:
> I'm not sure if `narrow` is the right expression for that.

It's the term used by [CLDR](http://cldr.unicode.org/translation/units) - ditto for the existing values: `short`, and `long`

moreati:
That codecov regression isn't strictly related to having format='narrow' - the innards of that if block already weren't covered. Do you want it fixed in this pull request? If so do you want the commit squashed?

sils1297:
> It's the term used by CLDR - ditto for the existing values: short, and long

alright

sils1297:
@moreati yes please add a check but in an own commit as this change makes sense without the other and they're orthogonal and atomic IMO.

sils1297:
that sentence got chaotic... this test is not explicitly bound to your change and thus it could be considered an atomic change and therefore is fine in another commit :)

PR #220, opened 2015-09-10
Backport: FixedOffsetTimezone: fix display of negative offsets

### LINKED ISSUE #214: FixedOffsetTimezone renders negative offsets incorrectly
```python
>>> babel.util.FixedOffsetTimezone(-60).zone
Etc/GMT+-60
```

moreati:
fixes #214 in 2.x-maintenance

### DIFF
--- a/babel/util.py
+++ b/babel/util.py
@@ -255,7 +255,7 @@ class FixedOffsetTimezone(tzinfo):
     def __init__(self, offset, name=None):
         self._offset = timedelta(minutes=offset)
         if name is None:
-            name = 'Etc/GMT+%d' % offset
+            name = 'Etc/GMT%+d' % offset
         self.zone = name
 
     def __str__(self):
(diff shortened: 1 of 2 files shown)


sils1297:
@moreati FWIW normally you don't mention the "backport" in the commit message, you usually just do two PRs from the same branch to master and the maintenance branch, so both get the same commit.

moreati:
> just do two PRs from the same branch to master and the maintenance branch, so both get the same commit

Ah, sorry - I'm still unlearning subversion habits that are bad git habits, and when (not) to use things like cherry-pick.

sils1297:
@moreati we'd use cherry-pick in coala (my open source project, coala-analyzer.org) because we want a linear history and it's a bit cleaner, code quality wise. But here we use merges so there's no need. No need to be sorry, where do we stay if not every single one of us learns a bit out of this :)

PR #223, opened 2015-09-17
Remove duplicate `sudo: false` entry.

rjollos:


sils1297:
@rjollos many thanks! (FWIW usually commit messages don't have a trailing period in the shortlog, just a common convention.)

PR #236, opened 2015-09-21
Enforce Python version in `setup.py`

rjollos:
Print error message and exit if Python version requirement not satisfied.


sils1297:
@rjollos hey, could you adress my comment? Also I'd love it if you could rebase the branch instead of merging master into it (what GitHub does), this clutters the history with merge commits and we do have quite enough of those IMO.

PR #238, opened 2015-09-21
setup.cfg: Use universal wheel

sils1297:
@etanol this one too

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -6,3 +6,6 @@ release = egg_info -RDb ''
 
 [pytest]
 norecursedirs = .* _* scripts {args}
+
+[bdist_wheel]
+universal = 1


etanol:
Sorry, I took a bit long to see this.  As I mentioned, I'm in the middle of a relocation.

@sils1297  I think you were too quick to merge this one.  The generated wheel is missing the `localedata/__init__.py` due to the data files definition in `setup.py`.  I think the easy fix would be to move the data files to a different subdirectory, or to refactor the code and rename the `babel.localedata` module/package.

I'm checking the 2.x branch but that one will probably work if it doesn't contain the `babel.localedata` module is still in the `localedata.py` file.

sils1297:
> Sorry, I took a bit long to see this.

Ah, I merged because the other one got approved and they're essentially the same. We also need review policies...

etanol:
Related commit: 7d387f95f4f41ed646bcb2b4fa39dae4a601b6ce

sils1297:
@etanol would you mind joining https://gitter.im/python-babel/babel for a talk? We need to organize a bit.

PR #245, opened 2015-09-22
Add CONTRIBUTING

sils1297:
This finally fixes our review process and gives users a bug template.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,0 +1,52 @@
+# Babel Contribution Guidelines
+
+Welcome to Babel! These guidelines will give you a short overview over how we
+handle issues and PRs in this repository. Note that they are preliminary and
+still need proper phrasing - if you'd like to help - be sure to make a PR.
+
+Please know that we do appreciate all contributions - bug reports as well as
+Pull Requests.
+
+## Filing Issues
+
+When filing an issue, please use this template:
+
+```
+# Overview Description
+
+# Steps to Reproduce
+
+1.
+2.
+3.
+
+# Actual Results
+
+# Expected Results
+
+# Reproducibility
+
+# Additional Information:
+
+```
+
+## PR Merge Criteria
+
+For a PR to be merged, the following statements must hold true:
+
+- All CI services pass. (Windows build, linux build, sufficient test coverage.)
+- All commits must have been reviewed and approved by a babel maintainer who is
+  not the author of the PR. Commits shall comply to the "Good Commits" standards
+  outlined below.
+
+## Correcting PRs
+
+Rebasing PRs is preferred over merging master into the source branches again
+and again cluttering our history. If a reviewer has suggestions, the commit
+shall be amended so the history is not cluttered by "fixup commits".
+
+## Writing Good Commits
+
+Please see
+http://coala.readthedocs.org/en/latest/Getting_Involved/Writing_Good_Commits/
+for guidelines on how to write good commits and proper commit messages.


sils1297:
@etanol @erickwilder whoever comes first, could you please self assign and review this? If you're fine with the PR, please try using dropping a plain `@rultor merge` command and see if this works.

erickwilder:
I think that the right home of this file is inside of the `docs/` directory (as an `.rst` not `.md` file), this way the contributing guidelines will be available at `babel.pocoo.org` too.

sils1297:
This way they will be picked up by github, see
https://github.com/blog/1184-contributing-guidelines

2015-09-23 17:15 GMT+02:00 Erick Wilder <notifications@github.com>:

erickwilder:
How about not using markdown, use RST without the `.rst` extension, the same way that `CHANGES` is used?
https://github.com/python-babel/babel/blob/master/docs/changelog.rst#L1

This way we address 2 issues: keep the main source of the documentation (babel.pocoo.org) updated with the contributing guidelines too.

sils1297:
That's actually a good idea though I don't like RST. Still it's a good idea and thus I'll do it :)

sils1297:
Hey, what you desire is done here: https://github.com/python-babel/babel/blob/sils/contributing_rst/CONTRIBUTING

However, I actually find the initial solution to be better, *because*:

- Markdown is way better readable in GitHub. RST isn't very good in plain text, links are weird and most importantly one cannot simply copy the bug template without hassle (on github)!
- Users will rather read this file on GitHub than in the official docs. The official docs could link to this but I think it makes sense to optimize readability for Markdown/Plain Text.

erickwilder:
I understand your point. But using RST is not a big deal, it's already used in the project (and most of the Python community is comfortable with it) and readability of plain RST files are not so different from Markdown. 
Also, we shouldn't took any project decision based on the fact of being hosted by Github. A comment I've seen before:
>  "Github could be the next Sourceforge"
And that's true.

sils1297:
> Also, we shouldn't took any project decision based on the fact of being hosted by Github. A comment I've seen before:
> >  "Github could be the next Sourceforge"
> 
> And that's true.

Heh what a comment. :) Good catch. However, changing the document to whatever one might switch to in the future is not a problem. I did this comment when dealing with commit messages which are "meant to be forever" because they're by definition our history and we want our history to be browsable at all times even if we switch in the future. This does not apply to this document which should be tailored to our current needs. As of now we *are* using GitHub and people checking out this repository will IMO most likely view this document on GitHub. Having it in the documentation *at all* is not needed IMO because the docs are for users of Babel.

> readability of plain RST files are not so different from Markdown.

So my main concern is the code block with is way better to copy in plain text with the three backticks. Secondly for me personally (just one statistical data point here) markdown is way better readable because it seems to be a little more leightweight but still very expressive. As mentioned befre I'm not a sphinx expert *at all* so I might be missing an alternative syntax or so.

erickwilder:
If you want to use Markdown, so be it. Just for the record, it's not what I'd do

erickwilder:
@sils1297 Could you rebase this one, after that we can merge.

sils1297:
> If you want to use Markdown, so be it. Just for the record, it's not what I'd do

Ok, this seems to be really an opinion thing, fwiw I have nothing against being convinced towards the better solution.

> Could you rebase this one, after that we can merge.

So I'd like to retry rultor on this one because the merging of the maintenance branch does not work presumably because of a rultor bug which I suspect doesn't apply here - if that works than I wouldn't need the additional rebase as rultor would do that for me. So I'll try that out :)

sils1297:
@rultor merge

sils1297:
Ok, that's what I wanted to know - GitHub is a pain in the ass... but I can work around that.

sils1297:
@rultor merge again please, this time it might work.

PR #246, opened 2015-09-22
Sils/wheel

sils1297:
@etanol can you check if this fixes your issue? I didn't test this because of time and I'd want you to try to reproduce it anyways as you found it. :)


etanol:
May I suggest using a hyphen `-` instead of an underscore `_` in the new name of the locale data subdirectory?  This way, it will convert it in an illegal package/module name, and it will be clearer that it is not meant to be imported.

Otherwise, it looks good to me.

sils1297:
I dislike hyphens in filenames. Plus I'm not sure everyone knows that hypens in module names aren't allowed in python, tbh this is new to me and wouldn't make it more obvious to me. The more important and obvious indicator is IMO that there are no python files in it...

erickwilder:
Actually @etanol's idea is simple but effective. I like it. :+1: 
Every programming language that I've learned so far has some kind of "valid identifiers" section in their docs and Python is not different. As far as I can remember, this was something right in the beginning of the [Python Tutorial](#python-tutor). So, no excuses to "not knowing hyphens are not valid package/module name" in this case.

[#python-tutor]: https://docs.python.org/3/tutorial/

sils1297:
@etanol @erickwilder thanks for convincing me of the right thing. I updated the PR, waiting for acknowledgement from one of you.

etanol:
Travis is failing with PyPy due to some environment problem, not the actual test run.  @sils1297 check [the Travis log](https://travis-ci.org/python-babel/babel/jobs/82212244), do you know how to amend the configuration?

setup.py
```diff
@@ -62,7 +62,7 @@ def run(self):
         'Topic :: Software Development :: Libraries :: Python Modules',
     ],
     packages=['babel', 'babel.messages', 'babel.localtime'],
-    package_data={'babel': ['global.dat', 'localedata/*.dat']},
+    package_data={'babel': ['global.dat', 'locale-data/*.dat']},
```
etanol: Since you already included those in `MANIFEST.in`, and we are using *setuptools*, you can save some work by using `include_package_data=True` instead.  [Reference](http://pythonhosted.org/setuptools/setuptools.html#new-and-changed-setup-keywords).

sils1297: I filed https://github.com/python-babel/babel/issues/260 because this is another change and should be done in a nown commit.

sils1297:
> Travis is failing with PyPy due to some environment problem, not the actual test run

Well looks like codecov failed again... I restarted the build for now, if this fails again I'll ping the codecov maintainer about this.

sils1297:
@etanol whatever this was, I guess he fixed it. FWIW build passes now.

sils1297:
@etanol waiting for your approval or disapproval here.

sils1297:
ah you labelled it, ok merging.

PR #249, opened 2015-09-23
Fix zh plural selection

mbirtwell:
Provide only one option for chinese in the plural data. The 3 previous options were all the same and I've checked with a chinese colleague she thinks that applies to all variants on the chinese language. 

Before this change things like ```get_plurals('zh_CN')``` would return the default plural rules because that string would get expanded out to ```Locale('zh', 'CN', 'Hans')``` with string representation ```"zh_Hans_CH"```, which isn't in the table. 

I feel like this is side stepping the underlying issue that the way that this table is searched isn't very flexible, but I thought it might be a pragmatic improvement.

### DIFF
--- a/babel/messages/plurals.py
+++ b/babel/messages/plurals.py
@@ -192,10 +192,8 @@ PLURALS = {
     'vi': (1, '0'),
     # Xhosa - From Pootle's PO's
     'xh': (2, '(n != 1)'),
-    # Chinese - From Pootle's PO's
-    'zh_CN': (1, '0'),
-    'zh_HK': (1, '0'),
-    'zh_TW': (1, '0'),
+    # Chinese - From Pootle's PO's (modified)
+    'zh': (1, '0'),
 }
 
 
--- a/tests/messages/test_plurals.py
+++ b/tests/messages/test_plurals.py
@@ -10,17 +10,34 @@
 # This software consists of voluntary contributions made by many
 # individuals. For the exact contribution history, see the revision
 # history and logs, available at http://babel.edgewall.org/log/.
+import pytest
 
-import doctest
-import unittest
-
+from babel import Locale
 from babel.messages import plurals
 
 
-def test_get_plural():
-    assert plurals.get_plural(locale='en') == (2, '(n != 1)')
+@pytest.mark.parametrize(('locale', 'num_plurals', 'plural_expr'), [
+    (Locale('en'), 2, '(n != 1)'),
+    (Locale('en', 'US'), 2, '(n != 1)'),
+    (Locale('zh'), 1, '0'),
+    (Locale('zh', script='Hans'), 1, '0'),
+    (Locale('zh', script='Hant'), 1, '0'),
+    (Locale('zh', 'CN', 'Hans'), 1, '0'),
+    (Locale('zh', 'TW', 'Hant'), 1, '0'),
+])
+def test_get_plural_selection(locale, num_plurals, plural_expr):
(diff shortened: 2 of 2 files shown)


sils1297:
@mbirtwell hey, many thanks for this PR! I'll review this ASAP. One minor thing, could you please remove this huge merge commit which is totally unreviewable and rebase on the current master instead? It's just a simple `git fetch origin && git reset --hard HEAD~1 && git rebase --interactive origin/master && git push --force`, assuming that `origin` is this repository.

sils1297:
/me wonders why travis didn't pick up this PR...

sils1297:
> I feel like this is side stepping the underlying issue

I'm not sure if this is the right approach. I'd rather strive for the cleaner solution where possible. @etanol maybe you can help here and provide some insight as you have more technical experience?

Marking this `in progress` again as I think we'll want to discuss this first.

mbirtwell:
If travis does pick it up I think it will be sad. The tests all pass based on 2.0 where I did the work, but I get a bunch of failures when on master

mbirtwell:
Hmm looking harder at those failures it might be because I've got the wrong CLDR data for that branch.

sils1297:
@mbirtwell ok, waiting for your investigation then, thanks!

mbirtwell:
Yes it was just the wrong cldr data. That caused my failures. 

As to whether there is a need for a cleaner solution. I think that this change is good either way. I've just had a look at the list of plural rules and the only other one that is not just a plain language is pt_BR because it seems to differ from non-Brazilian Portuguese. That one appears to work because correctly because Locale will always omit the script for Portuguese.  

i.e.
```
In [2]: get_plural('pt_BR')
Out[2]: (2, '(n > 1)')

In [3]: get_plural('pt_Latn_BR')
Out[3]: (2, '(n > 1)')

In [5]: get_plural('pt')
Out[5]: (2, '(n != 1)')

```

So pragmatically I think this works although the system might be a little fragile.

sils1297:
@etanol can you help out here? You should have a much easier time making this decision and performing the review.

etanol:
There seems to be an underlying issue indeed.  Apparently, when we parse any `zh` locale we assume the `Hans` script if it hasn't been specified.  Therefore, when representing the locale as string, the scripts get automatically added.

Your solution is very pragmatic and it doesn't change the meaning since all plural rules for Chinese were the same.

However, you need to move the test code to the `tests.messages.test_plurals` module instead.  Also, could you rework the test code **without** the *pytest* decorator?

mbirtwell:
I'm happy to move the test code. In retrospect that is obviously where it should have been. 

I can also change the test to loop over the examples in the test function, but I'm not quite so happy about it. I think the pytest parametrize approach is a cleaner way of writing these kind of tests. Why do prefer doing without it?

etanol:
It's just a matter of symmetry.  If every contributor composes test cases freely, then overall code is harder to understand.  I'm not against using a more declarative approach to testing, but a uniform code/testing style is more valuable from the maintainer perspective.

mbirtwell:
mark and parametrize are both built-in parts of py test, the best documentation I know of on parametrize is here https://pytest.org/latest/parametrize.html

etanol:
@mbirtwell Never mind my comment about the use of `pytest.mark.parametrize`.  It has been used before so there's no need to amend that specific part.

mbirtwell:
I've reworked the tests and squashed everything.

I made more changes to the tests than I originally planned because it looked a bit duplicated when I had the parametrize version. I've tried to break the tests up so that they each have a clear purpose.

Issue #247, opened 2015-09-22
Use tox when running tests with rultor


erickwilder:
We're using plain `pytest` invokation when using `@rultor` to merge pull requests. This can lead to broken behaviour in some of our supported Python versions. To avoid that we **must** use `tox` command to be able to test on all supported versions when using `@rultor` commands.

PR #252, opened 2015-09-24
rultor: Remove test execution

sils1297:
Something's wrong here and we do need to debug that first. So in order
for us  being able to do rultor merges, let's just not execute tests for
now.

@etanol @erickwilder could you please review and merge this before others as this blocks all rultor based merges?

### DIFF
--- a/.rultor.yml
+++ b/.rultor.yml
@@ -9,6 +9,4 @@ merge:
   fast-forward: only
   rebase: true
   script:
-    - pip install .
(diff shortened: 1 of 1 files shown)


erickwilder:
Ok to me. Should I merge with rultor or the merge button?

erickwilder:
The #247 issue should address this problem. I'm working on that

sils1297:
As rultor fails we need to do the manual merge here.

PR #265, opened 2015-09-28
Add support for date-time skeletons

mbirtwell:
The skeletons for dates and times are described on
http://cldr.unicode.org/translation/date-time-patterns under
Additional Date-Time Formats. And are useful when you want to some more
control over formatting dates and times but don't want to force all
locales to use the same pattern.

### DIFF
--- a/babel/core.py
+++ b/babel/core.py
@@ -732,6 +732,19 @@ class Locale(object):
         return self._data['datetime_formats']
 
     @property
+    def datetime_skeletons(self):
+        """Locale patterns for formatting parts of a datetime.
+
+        >>> Locale('en').datetime_skeletons['MEd']
+        <DateTimePattern u'E, M/d'>
+        >>> Locale('fr').datetime_skeletons['MEd']
+        <DateTimePattern u'E dd/MM'>
+        >>> Locale('fr').datetime_skeletons['H']
+        <DateTimePattern u"HH 'h'">
+        """
+        return self._data['datetime_skeletons']
+
+    @property
     def plural_form(self):
         """Plural rules for the locale.
 
--- a/babel/dates.py
+++ b/babel/dates.py
@@ -694,6 +694,34 @@ def format_time(time=None, format='medium', tzinfo=None, locale=LC_TIME):
     return parse_pattern(format).apply(time, locale)
 
 
+def format_skeleton(skeleton, datetime=None, tzinfo=None, locale=LC_TIME):
+    r"""Return a time and/or date formatted according to the given pattern.
+
+    The skeletons are defined in the CLDR data and provide more flexibility
+    than the simple short/long/medium formats, but are a bit harder to use.
+    The are defined using the date/time symbols without order or punctuation
+    and map to a suitable format for the given locale.
+
+    >>> t = datetime(2007, 4, 1, 15, 30)
+    >>> format_skeleton('MMMEd', t, locale='fr')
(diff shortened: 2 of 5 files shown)


etanol:
Interesting.  However, dedicated test code is missing.  I know, doctests should cover it but I have a feeling that, if not me, somebody else will ask you for it.

xmo-odoo:
I found this while looking into intervals (as they use skeletons), this PR implements basic support for skeletons, but doesn't seem to support [skeleton matching](http://www.unicode.org/reports/tr35/tr35-dates.html#availableFormats_appendItems) (see "matching skeletons" sub-section) and match adjustment, is that normal/intentional?

I also am not sure about the inheritance support: the spec states
> dateFormatItems inherit from their parent locale, so the inherited items need to be considered when processing.

which I read as each skeleton (`dateFormatItem/@id`) not found on the current locale being looked up in the parent locale, is that automatically resolved by babel?

sils1297:
@etanol uhm I'mnot quite sure I agree with you here, that would mean that we have redundant tests which is nothing but a maintenance overhead, am I wrong?

etanol:
@sils1297 Yes, they are an overhead.  However, if you check the test suite you will find that there are quite a few tests that are identical to the corresponding docstring.  Since I don't know the whole history of Babel, I'm not clear on how this redundancy originated.

sils1297:
@etanol so we probably want to avoid redundancy from now on and approve PRs that remove identical tests, agreed? Code duplication is bad and doctest is a great invention to reduce testing overhead, we should use that advantage IMO.

mbirtwell:
@sils1297 I've removed the commit which added the extra tests from this PR.

@xmo-odoo I'll be honest I did the minimal here that I needed for my use-case. Which didn't include skeleton-matching although that looks nice I think this is still useful without that. Inheritance however seems to have been handled for me by babel's CLDR import mechanism. e.g. there is no 'hm' skeleton in the en_GB.xml file but the following works:

```
In [1]: import babel

In [2]: l = babel.Locale.parse('en_GB')

In [3]: l.datetime_skeletons['hm']
Out[3]: <DateTimePattern 'h:mm a'>
```

masklinn:
@mbirtwell But there's no skeletons at all in en_GB in recent CLDR right? I was wondering more about a partial list in the sub-locale, does Babel just use this partial list or does it merge with the parent locale's?

mbirtwell:
In CLDR 26, which babel is currently using there is a partial list of skeletons that appears to get correctly merged with the parent locale (en_US I think in CLDR 26, although I believe that is also changing).

xmo-odoo:
@mbirtwell ok, good to know then (I only checked 23.1 — used by the last release — and 28 — latest CLDR — and neither has "partial" skeletons lists)

akx:
@mbirtwell Could you please rebase on top of 2.2.0? One doctest will also need updating based on the current CLDR data.

```python
================================== FAILURES ===================================
_______________ [doctest] babel.core.Locale.datetime_skeletons ________________
EXAMPLE LOCATION UNKNOWN, not showing all tests of that example
??? >>> Locale('fr').datetime_skeletons['MEd']
Expected:
    <DateTimePattern u'E d/M'>
Got:
    <DateTimePattern 'E dd/MM'>

==================== 1 failed, 520 passed in 5.28 seconds =====================
```

akx:
Thanks @mbirtwell! :tada:

PR #267, opened 2015-09-29
Extract a format_date_with_time helper function

mbirtwell:
The exposes some of the behaviour of format_datetime which might be useful
on it's own as it's own function.

Although this doesn't require the changes I proposed in #265 it is complementary to them. 

It allows you to format the date and time using different or explicit formats and then combine them in a locale aware way.

### DIFF
--- a/babel/dates.py
+++ b/babel/dates.py
@@ -569,6 +569,23 @@ def format_date(date=None, format='medium', locale=LC_TIME):
     return pattern.apply(date, locale)
 
 
+def format_date_with_time(date, time, format='medium', locale=LC_TIME):
+    r"""Combines a pre-formatted date and time in to a single string
+
+    >>> format_date_with_time("20/8/15", "14:40", locale='en')
+    u'20/8/15, 14:40'
+    >>> format_date_with_time("20/8/15", "14:40", 'long', 'en')
+    u'20/8/15 at 14:40'
+    >>> format_date_with_time("20/8/15", "14:40", 'long', 'de')
+    u'20/8/15 um 14:40'
+
+    """
+    return get_datetime_format(format, locale=locale) \
+        .replace("'", "") \
+        .replace('{0}', time) \
+        .replace('{1}', date)
+
+
 def format_datetime(datetime=None, format='medium', tzinfo=None,
                     locale=LC_TIME):
     r"""Return a date formatted according to the given pattern.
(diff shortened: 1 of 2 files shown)


etanol:
I'm sorry, but no.  This implementation cannot handle timezone support at all.  Trying to concatenate `date` and `time` on the wild is just asking for trouble.  That is why we have a `datetime` type.

mbirtwell:
Well with out a function like this you are making life hard for someone who wants a long date but to only give the time to the nearest minute. Very reasonable for example if you are showing the start time of a meeting. It could be implemented with my function like:

```
def format_mydatetime(ts, locale):
    return format_date_with_time(
        format_date(ts, 'full', locale=locale),
        format_time(ts, 'short', locale=locale),
        'long',
        locale,
    )
```

format_date and format_time are both tz aware and assuming that ts is a datetime with the appropriate time zone info this will just work. I admit it requires the user to do things properly so maybe it should come with a health warning, but it's not too hard to do so and if you don't provide this function people will just have to implement potentially buggy copies of it instead.

etanol:
It's just bad API design.  Formatting functions should accept native datatypes and convert them to localized strings.  I don't see the justification in adding a function that can accept arbitrary strings and returns another (partially localized) string.

I'm very sorry about this because I've been in your position several times.  But this one is not going to happen.

mbirtwell:
How would you feel about a pull request that augmented format_datetime instead so that it provided options for using different formats for the date and the time?

etanol:
That is certainly more reasonable.  For example, optionally accepting a tuple or list from the `format` parameter.  It would keep compatibility, being a very subtle API change, and you would get your desired functionality with almost no code duplication.

mbirtwell:
So I was thinking prototype would be something like
```
format_datetime(datetime=None, format='medium', tzinfo=None, locale=LC_TIME, date_fmt=None, time_fmt=None)
```
where date_fmt and time_fmt would be ignored if format is not one of ('full', 'long', 'medium', 'short')
date_fmt and time_fmt would default to format if not set
date_fmt and time_fmt would be passed to format_date and format_time appropriately. 

date_fmt and time_fmt could the each independantly be one of the standard format names or a custom format. The one thing that it seems to me this misses is support for skeletons (which you might have noticed I'm keen on). But it's easy enough to look up the skeleton from the locale object and pass this in as a custom format.

Issue #90, opened 2014-04-24
Bypass number rounding


kdeldycke:
`babel.numbers.format_decimal()` and `babel.numbers.format_currency()` have a built-in banker rounding implemented. See: https://github.com/python-babel/babel/blob/3ec7bb104e808d1fda16fbf9fec4b9d7efccaa3f/babel/numbers.py#L648-L649

If we need to localize numbers without rounding, we can't use these methods.

The minimal default decimal precision per method are:

  * `babel.numbers.format_decimal()` => 3 digits after the dot
  * `babel.numbers.format_currency()` => 2 digits after the dot

Proof:

```python
>>> import babel
>>> set([babel.Locale.parse(l).decimal_formats._data[None].frac_prec
...      for l in babel.localedata.locale_identifiers()])
set([(0, 6), (0, 3)])
>>>
```

```python
>>> import babel
>>> set([babel.Locale.parse(l).currency_formats._data[None].frac_prec
...      for l in babel.localedata.locale_identifiers()])
set([(2, 2)])
>>>
```

So if you have monetary amounts to localize, and they're already rounded to 2 trailing digits, then you're lucky. Using any of the `format_*()` methods will have no side effects.

But for other numbers with higher precision, using `format_*()` as-is is dangerous. It will introduce unwanted rounding. For example, calling `format_currency(0.9999, 'EUR', locale='fr')` will return `1,00 €`. I expect here to get the pristine `0,9999 €` string.

I think there must be a clean and documented way to bypass arbitrary rounding when localizing numbers.

etanol:
The CLDR specification section about [number rounding](http://unicode.org/reports/tr35/tr35-numbers.html#Rounding) states that *half-even* should be the default algorithm.  However, it seems that there is room to alternative rounding modes as an implementor decision.

This can be easily implemented once I manage to land my new code that relies on the `decimal` standard package to perform the rounding.

kdeldycke:
Good ! Didn't know there was a refactor in progress. I especially welcome `decimal`-based code, as I might finally get rid of some dirty workarounds ! :)

akx:
@etanol, @kdeldycke: Is this issue still valid?

etanol:
Yes, in fact, now that #272 has been merged, it is *easier* to implement.  Although I have a design dilemma:

1. Adding a parameter to optionally specify the rounding mode
2. Forcing the use of [decimal contexts](https://docs.python.org/3/library/decimal.html#context-objects) to override decimal parameters

[...]

PR #272, opened 2015-10-09
Faster rounding

etanol:
At work, we had some performance issues when trying to format almost 100,000 prices in a single report.  Some profiling revelaed inefficiencies in the number rounding functions.  Such number rounding operations can be implemented making use of the `decimal` standard Python module.  Which probably means that the number rounding code in Babel was meant to support Python versions below 2.4.

Since the oldest supported version now is 2.6, it is probably time to remove the rust and make more use of the standard library (when compatibility is not an issue).

Moreover, this pull request also contains some additional code to make use of the `cdecimal` implementation when available (already shipped with Python since 3.3).  Which speeds up some operations considerably.  Here are some dumb microbenchmarks taken with CPython 2.7.

Original rounding implementation:

```
In [4]: %timeit format_currency(1234.5678, 'USD', locale='en_US')
1000 loops, best of 3: 284 µs per loop
```

New rounding implementation based on `decimal`:

```
In [4]: %timeit format_currency(1234.5678, 'USD', locale='en_US')
1000 loops, best of 3: 229 µs per loop
```

New rounding implementation based on `cdecimal`:

```
In [4]: %timeit format_currency(1234.5678, 'USD', locale='en_US')
10000 loops, best of 3: 100 µs per loop
```

This will also make it easier to provide a solution to #90.  However, the API design for that will have to be discussed a bit.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -12,7 +12,15 @@ matrix:
     - os: linux
       python: 2.6
     - os: linux
+      python: 2.6
+      env:
+      - CDECIMAL=cdecimal
+    - os: linux
       python: 2.7
+    - os: linux
+      python: 2.7
+      env:
+      - CDECIMAL=cdecimal
     - os: linux
       python: pypy
     - os: linux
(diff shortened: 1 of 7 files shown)


erickwilder:
Well, honestly I do like pull requests that [removes more code rather than add](https://www.youtube.com/watch?v=oHcToNCn4K8). But I see that there's only a unused test case being removed here (`bankersround()` function is gone) - but you've chaged `parse_grouping()` function a lot. I'm wondering if it's not worth adding some unit tests just to `parse_grouping()` function, since the `bankersround()` function had tests dedicated just to it.

I know that you're indirectly testing it by showing that the behaviour of the library hasn't changed - but I see that adding those tests could prevent some regressions in the future.

Nice work here!

PR #274, opened 2015-10-13
Fix UnicodeEncodeError on file encoding detection

imankulov:
If the first line of a python file is not a valid latin-1 string,
parse_encoding dies with "UnicodeDecodeError". These strings nonetheless can be
valid in some scenarios (for example, Mako extractor uses
babel.messages.extract.extract_python), and it makes more sense to ignore this
exception and return None.

### DIFF
--- a/babel/util.py
+++ b/babel/util.py
@@ -65,7 +65,7 @@ def parse_encoding(fp):
             try:
                 import parser
                 parser.suite(line1.decode('latin-1'))
-            except (ImportError, SyntaxError):
+            except (ImportError, SyntaxError, UnicodeEncodeError):
                 # Either it's a real syntax error, in which case the source is
                 # not valid python source, or line2 is a continuation of line1,
                 # in which case we don't want to scan line2 for a magic
--- a/tests/test_util.py
+++ b/tests/test_util.py
@@ -14,6 +14,7 @@
 import unittest
 
 from babel import util
+from babel._compat import BytesIO
 
 
 def test_distinct():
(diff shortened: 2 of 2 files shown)


etanol:
Very nice.

etanol:
Do you have a working example where this happens?  Because any byte sequence can be decoded as Latin1.  It doesn't mean that the decoded result should make sense, but decoding Latin1 to `unicode` usually never fails.

imankulov:
Oh, to be frank, I misunderstood the source of the problem initially. Of course, when you have a byte sequence (Python 2.x string) and you *decode* it, you cannot end up with Unicode**Encode**Error.

The exception is thrown later on, when we try to parse the resulting object with `parser.suite `. The thing is, in Python 2.x parser.suite expects a byte sequence (py2.x str), and in Python 3.x it expects a unicode object (py3.x str).

**Python 2.x example**

```python
$ python
>>> import parser
>>> parser.suite('print("я")')
<parser.st object at 0x7f3150d18090>
>>> parser.suite(u'print("я")')
UnicodeEncodeError: 'ascii' codec can't encode character u'\u044f' in position 7: ordinal not in range(128)
```

**Python 3.x example**

```python
$ python3
>>> import parser
>>> parser.suite('print("я")')
<parser.st object at 0x7f1f222e3f90>
>>> parser.suite('print("я")'.encode('utf8'))
TypeError: suite() argument 1 must be str, not bytes
```

As it turned out, to support both 2.x and 3.x properly we have to switch between py3.x str and py2.x str, depending on the interpreter version. Current solution is to "die quietly" if first string contains non-ascii symbols. As I feel it,  it is "good enough" to cover almost everything. Non-ascii first string theoretically can contain encoding information, like `# coding: utf-8 😈`, but do we really have to properly process this case?

akx:
@imankulov Can you rebase this, please?

imankulov:
@akx Done.

akx:
Thank you, this looks good!

imankulov:
@akx Awesome! Thanks a lot for merging it.

PR #278, opened 2015-10-31
Sils/coala

sils1297:
Hey guys, just a little code checking addition, doesn't do much yet obviously.

This is at the same time the configuration for GitMate, a little service I'm just writing, online at http://gitmate-coalas.rhcloud.com . It's in a shaky state (prototype, I'm currently redeploying a new version more than daily) but it can check pull requests using coala and drop comments for any issues right in GitHub - wanna try? I'm currently working on getting PyLint analysis in properly, shouldn't take long to get that. I think it's pretty cool, I'd appreciate any feedback concerning usability, no bug reports though.

### DIFF
--- a/babel/messages/catalog.py
+++ b/babel/messages/catalog.py
@@ -93,10 +93,10 @@ class Message(object):
                        PO file, if any
         :param context: the message context
         """
-        self.id = id #: The message ID
+        self.id = id
         if not string and self.pluralizable:
             string = (u'', u'')
-        self.string = string #: The message translation
+        self.string = string
         self.locations = list(distinct(locations))
         self.flags = set(flags)
         if id and self.python_format:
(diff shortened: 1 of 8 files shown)


sils1297:
GitMate moved to app.gitmate.com and gets more mature every day.

sils1297:
This is how it'll deal with PRs if we merge this and I turn it on: https://github.com/coala-analyzer/gitmate-test/pull/3

babel/messages/catalog.py
```diff
@@ -93,10 +93,10 @@ def __init__(self, id, string=u'', locations=(), flags=(), auto_comments=(),
                        PO file, if any
         :param context: the message context
         """
-        self.id = id #: The message ID
```
akx: Hmm, I think `#:` comments should be left as `#:` (and not turned into `# :`), because some tools (as far as I know) use them for some sort of automated documentation generation.

sils1297:
Note that this is an aesthetic PR and this does not have the 90% coverage hit.

babel/messages/catalog.py
sils1297: hm, would be interesting to know what they're used for here. E.g. in this case the id param already has a documentation.

sils1297: I was not able to find anything regarding to that in the net, the comments were introduced in 7ab115ce the first time with no comment or reasoning in the commit message. I've pushed a commit on top that removes them.

akx:
LGTM.

Seems like harmless bears, not scary bears. :bear:

sils1297:
@akx thanks, merging after CI approval.

Issue #263, opened 2015-09-26
Updating the CHANGES file - new concept


sils1297:
@erickwilder @etanol the changes file is written to at almost every PR and is the main source for conflicts during merge/rebase. I think we should come up with a new strategy here.

How about: whenever some maintainer merges a PR, he changes the CHANGES file and commits that to a `feature/changes` branch which has a Pull Request open which we merge when we do the release? This could be even done via the GitHub UI if a merge is done while lacking a proper terminal.

erickwilder:
Sounds good. We can give a try. :+1:

techtonik:
Alternative workflow ifs to create `CHANGES.techtonik` files that are being merged into CHANGES during release.

sils1297:
We could also just collect them on a wiki page *or* simply update the changes file when we're doing a release by crawling the commit history. I'm collecting changes here now: https://github.com/python-babel/babel/wiki/CHANGES

akx:
The wiki page approach seems to work fine. Closing this as resolved with that method :)

PR #279, opened 2015-11-04
Allow file locations without line numbers.

fschulze:
I'm proposing that locations without line numbers should work. Currently they are just swallowed. Other tools like msgmerge from gettext keep them as they are. Locations without line numbers are useful to reduce the number of differences when templates and code are edited without changing any of the translations.
The current implementation might break backward compatibility for other code that uses pofile from Babel.
Another possible solution would be to just keep unparsed locations as regular comments, but that is quite a bit more complex to handle correctly in my opinion.

### DIFF
--- a/babel/messages/pofile.py
+++ b/babel/messages/pofile.py
@@ -218,6 +218,8 @@ def read_po(fileobj, locale=None, domain=None, ignore_obsolete=False, charset=No
                         except ValueError:
                             continue
                         locations.append((location[:pos], lineno))
+                    else:
+                        locations.append((location, None))
             elif line[1:].startswith(','):
                 for flag in line[2:].lstrip().split(','):
                     flags.append(flag.strip())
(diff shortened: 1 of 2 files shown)


sils1297:
@fschulze hey, thanks for submitting this PR! I hope I'll have time to take a look at this within a reasonable delay (I'm off for GSoC mentor summit from tomorrow), if not please ping me. This project is lacking review manpower, if you'd like to join we'd appreciate any help properly reviewing other PRs!

babel/messages/pofile.py
```diff
@@ -449,9 +451,13 @@ def _write_message(message, prefix=''):
             _write_comment(comment, prefix='.')
 
         if not no_location:
-            locs = u' '.join([u'%s:%d' % (filename.replace(os.sep, '/'), lineno)
-                              for filename, lineno in message.locations])
-            _write_comment(locs, prefix=':')
+            locs = []
+            for filename, lineno in message.locations:
+                if lineno:
+                    locs.append(u'%s:%d' % (filename.replace(os.sep, '/'), lineno))
```
akx: The `filename.replace(os.sep, '/')` duplication could be moved to a `filename = filename.replace(os.sep, '/')` statement.

akx:
:+1: for this; allowing omitting line numbers altogether makes diffs to .po files much nicer to read. 

(Setting `lineno` to `0` works too, but `filename:0` isn't as elegant as no lineno altogether.)

sils1297:
@akx thanks!

sils1297:
added note to CHANGES file: https://github.com/python-babel/babel/wiki/CHANGES

fschulze:
@akx and @sils1297 : thanks for the review and merge!

fschulze:
Just curious, shouldn't the CHANGES file be updated and not only the wiki entry?

akx:
@fschulze: As I understand the CHANGES file in the repo will get updated at release, see https://github.com/python-babel/babel/issues/263

sils1297:
@fschulze updating CHANGES in each PR leads to lots of conflicts, in some cases maintainers check the commit log and compile the file from that but we have lots of merge commits cluttering the history here and no time to make every contributor rebase his PR so the wiki page is right now the most convenient option :/

fschulze:
Yeah, I know the pain of conflicts for CHANGES. Thanks for the clarification.

PR #294, opened 2015-12-21
Test/CI improvements

akx:


akx:
I think `codecov/patch` is a little confused.

sils1297:
@akx seems correct to me, you seem to have touched 4 non comment lines, 3 of them are covered in tests which makes a patch coverage of 75%. I've deactivated that feature of codecov now so it'll only look for coverage raising or staying.

PR #295, opened 2015-12-21
Compile noninteger TR35 operands to zeroes when emitting Gettext

### LINKED ISSUE #287: CLDR parser is probably broken (invalid plural forms)
While trying to import data from CLDR with babel I encountered some issues which I would like to share.

- Version 2.1.1 installed with pip returns invalid number of plurals at least for Turkish (tr) and Hungarian (hu). According to babel, both languages have only one form, while [in fact they have two](http://www.unicode.org/cldr/charts/28/supplemental/language_plural_rules.html#hu): `one` and `other`.
- Version installed from git fixes the issue with Hungarian and Turkish, but gettext strings returned for at least Russian and Polish, contain forbidden variables. While gettext only accepts `n` as a variable, Babel rules contain `i` and `v` vars.

Attaching the gist with my testing script and my output: https://gist.github.com/imankulov/866660acdd90cb4004c7

Tested with following interpreter

```
Python 2.7.10 (default, Sep 23 2015, 04:34:14)
[GCC 4.2.1 Compatible Apple LLVM 7.0.0 (clang-700.0.72)] on darwin
```

I would appreciate any input. Thanks!

akx:
See http://www.unicode.org/reports/tr35/tr35-numbers.html#Operands
See https://www.gnu.org/software/gettext/manual/html_node/Plural-forms.html

Fixes #287

### DIFF
--- a/babel/plural.py
+++ b/babel/plural.py
@@ -485,6 +485,9 @@ def _unary_compiler(tmpl):
     return lambda self, x: tmpl % self.compile(x)
 
 
+compile_zero = lambda x: '0'
+
+
 class _Compiler(object):
     """The compilers are able to transform the expressions into multiple
     output formats.
@@ -557,10 +560,10 @@ class _JavaScriptCompiler(_GettextCompiler):
     # XXX: presently javascript does not support any of the
     # fraction support and basically only deals with integers.
     compile_i = lambda x: 'parseInt(n, 10)'
-    compile_v = lambda x: '0'
-    compile_w = lambda x: '0'
-    compile_f = lambda x: '0'
-    compile_t = lambda x: '0'
+    compile_v = compile_zero
+    compile_w = compile_zero
+    compile_f = compile_zero
+    compile_t = compile_zero
 
     def compile_relation(self, method, expr, range_list):
         code = _GettextCompiler.compile_relation(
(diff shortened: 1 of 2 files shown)


akx:
@imankulov: If you're around, could you give this a review?

imankulov:
@akx I'm sorry for delay.

I finally managed to test it, and it works as expected! Rules for Russian and other languages involving variables other than `n`, can be simplified, of course, but I guess this is not the goal of this PR, and they are technically correct.

That's how Russian rule looks like, for example:

    nplurals=4; plural=(((((0 == 0)) && (((n % 10) >= 2 && (n % 10) <= 4))) && (!(((n % 100) >= 12 && (n % 100) <= 14)))) ? 1 : (((((0 == 0)) && (((n % 10) == 0))) || (((0 == 0)) && (((n % 10) >= 5 && (n % 10) <= 9)))) || (((0 == 0)) && (((n % 100) >= 11 && (n % 100) <= 14)))) ? 2 : ((((0 == 0)) && (((n % 10) == 1))) && (!(((n % 100) == 11)))) ? 0 : 3)

Have nothing against merging it. Thank you for the fix!

akx:
Great! Yeah, expression simplification is probably a little out of scope (though I have to admit it would be interesting...)

Merging this, thank you!

PR #298, opened 2015-12-23
Javascript: Support dotted names as keywords

### LINKED ISSUE #148: Keywords containing dots never match in JS
In our project we use `$t.gettext` in JavaScript files. However, `--no-default-keywords -k '$t.gettext'` results in an empty dictionary after extractions.

I believe this happens because the tokenizer only uses the `gettext` function name so the `$t.` is never part of it.

It would be nice if such cases were handled properly or if that's too complex, then defining a keyword containing a dot should result in an error instead.

akx:
Fixes #148


akx:
Superseded by #332.

PR #300, opened 2015-12-23
Embetter pofile sorting

akx:
I think sorting message locations irrespective of the sort flags is alright.

The only UC I can think of where that wouldn't be desired might be trying to read a .po and emitting an identical copy of it, but I don't think Babel does that as it is anyway?

### DIFF
--- a/babel/messages/pofile.py
+++ b/babel/messages/pofile.py
@@ -453,7 +453,7 @@ def write_po(fileobj, catalog, width=76, no_location=False, omit_header=False,
 
         if not no_location:
             locs = []
-            for filename, lineno in message.locations:
+            for filename, lineno in sorted(message.locations):
                 if lineno:
                     locs.append(u'%s:%d' % (filename.replace(os.sep, '/'), lineno))
                 else:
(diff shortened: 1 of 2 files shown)


akx:
@jtwang DRYed! Good point.

babel/messages/pofile.py
```diff
@@ -474,8 +472,21 @@ def _write_message(message, prefix=''):
...
+            sort_by_message=sort_output,
+            sort_by_file=sort_by_file
+        ):
             for comment in message.user_comments:
                 _write_comment(comment)
             _write_message(message, prefix='#~ ')
             _write('\n')
+
+
+def _sort_messages(messages, sort_by_message, sort_by_file):
```
jtwang: Test please and docs. No default values for sort flags?

akx:
@jtwang: I improved the signature of the helper function a little more -- no good reason to have two mutually exclusive arguments!

Also, I don't think it warrants a separate test; that function's coverage should already be 100% by way of the other tests.

jtwang:
ack 4f60b3ebde4a462ab686129982dc13f3d840617a, bc59c9c64ed285888a1bd929d55abbb63212e95b, edce8eeaa594e277d3abb29fb419cf6e4fefb52a, 65ce160b6915011e8c99e6175607e5fb329ce7fb

@akx totally disagree about testing, but I'll defer to your familiarity with this project. :)

Single arg change looks good (bah enums only in 3.4). Good to merge!

akx:
Cheers @jtwang! :)

PR #92, opened 2014-04-29
merge method fails on Python 3

dpgaspar:
Could you please apply this merge or correct this problem, i'm having the following error on Python 3.3

File "XXXXXX/lib/python3.3/site-packages/Babel-1.3-py3.3.egg/babel/support.py", line 590, in merge
    self.files.extend(translations.files)
AttributeError: 'filter' object has no attribute 'extend'

Filter in python 3 return iterators, so there is no extend method.

Thanks!

PR #301, opened 2015-12-23
Flatten NullTranslations.files into a list

### LINKED ISSUE #162: A bug in support.Translations class prevent merge method to work with python 3
This code sample works well in python 2:

    from babel.support import Translations
    translations = Translations.load("path/to/translations", ["en"])
    other_translations = Translations.load("other/path/to/translations", ["en"])
    translations.merge(other_translations)

Switching to python 3 make it to throw an AttributeError ('filter' object has no attribute 'extend') at babel/support.py:590:

    self.files.extend(translations.files)

It turns out that self.files is set at babel/support.py:284:

    self.files = filter(None, [getattr(fp, 'name', None)])

The filter method returns a list in python 2 but python 3 version returns a kind of generator, which has not extend method. Make a list from the filter fix the issue, but I'm not sure what side effects it could have:

    self.files = list(filter(None, [getattr(fp, 'name', None)]))

akx:
`filter` is special in Python 3, and latter usages would error out in
strange ways.

Fixes #92
Fixes #162

### DIFF
--- a/babel/support.py
+++ b/babel/support.py
@@ -298,7 +298,7 @@ class NullTranslations(gettext.NullTranslations, object):
         self._catalog = {}
         self.plural = lambda n: int(n != 1)
         super(NullTranslations, self).__init__(fp=fp)
-        self.files = filter(None, [getattr(fp, 'name', None)])
+        self.files = list(filter(None, [getattr(fp, 'name', None)]))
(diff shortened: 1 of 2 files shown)


sils1297:
all is green here, merging

PR #302, opened 2015-12-28
Try to fix the broken builds

etanol:
The cdecimal extension doesn't seem to install anymore.  It looks like it has been renamed to m3-cdecimal.  However, in my system doesn't install properly either.  Let's try Travis to see if it has better luck.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -16,13 +16,13 @@ matrix:
     - os: linux
       python: 2.6
       env:
-      - CDECIMAL=cdecimal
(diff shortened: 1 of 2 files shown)


akx:
The CI system doesn't use the tox config, so just edit .travis.yml?

etanol:
@akx Good catch, thanks!

sils1297:
ah you implemented it in a fork, GitMate won't pick up the comments in this case.

PR #303, opened 2015-12-29
Add hash to locale class

akx:
This PR replaces PR #153.


sils1297:
@akx does this need a changelog entry?

akx:
@sils1297 Will add.

PR #304, opened 2015-12-29
Intern the CLDR's Windows Zone Map

akx:
This avoids `babel.util` roundaboutly importing `babel.core` (and calling `get_global()`!) during import time on Windows systems.

I spotted this circular import during initial development for #31.


akx:
Probably not a good idea to intern the data after all.

PR #305, opened 2015-12-29
Make Locale objects immutable and cache them

akx:
Refs #31


etanol:
Looks good.  Although I can't risk my neck over the `Memoized` meta class, since I have played very little with them.

akx:
@mitsuhiko: Would you have time to give this a look? :)

babel/_memoized.py
```diff
@@ -0,0 +1,36 @@
+# TODO: This can't live in .util until the circular import of
+#       core -> util -> localtime -> win32 -> core is resolved.
+
+
+class Memoized(type):
+    """
+    Metaclass for memoization based on __init__ args/kwargs.
```
xmo-odoo: Should probably document that it assumes the constructor&initializer are pure since the cache-fetch is not locked, you could have two identical memoized instances being constructed.

babel/_memoized.py
```diff
@@ -0,0 +1,36 @@
...
+            key = cls._get_memo_key(args, kwargs)
+        else:
+            key = (args or None, frozenset(kwargs.items()) or None)
+
+        try:
+            return cls._cache[key]
+        except KeyError:
+            try:
+                if lock:
+                    lock.acquire()
```
xmo-odoo: Shouldn't that be outside the `try`? If lock acquisition fails for some reason, it's going to try and release it anyway, which might blow up with a different error or do odd things.

babel/_memoized.py
```diff
@@ -0,0 +1,36 @@
...
+        if "_cache_lock" not in dict:
+            dict["_cache_lock"] = None
+        return type.__new__(mcs, name, bases, dict)
+
+    def __memoized_init__(cls, *args, **kwargs):
+        lock = cls._cache_lock
+        if hasattr(cls, "_get_memo_key"):
+            key = cls._get_memo_key(args, kwargs)
+        else:
+            key = (args or None, frozenset(kwargs.items()) or None)
```
xmo-odoo: Maybe memoization could provide both a metaclass and a base class (using the metaclass), the latter providing a default _get_memo_key method? That would obviate the need for this conditional? (though not the `__new__` override)

akx:
This will probably need some rethinking wrt fuzzy locale support...

PR #54, opened 2013-09-05
Updates a catalog's header to from the template

dstanek:
A language specific catalog's initial header is based on the template
when it is created.  Updates to the template's header were previously
not getting into the catalog during the update_catalog process.

This came from trying to update the copyright holder in the 
[Keystone project](https://git.openstack.org/cgit/openstack/keystone).  I had to edit the generated .po files by hand since the headers were not being updated.

PR #307, opened 2015-12-30
Update catalog headers from the template

akx:
Rebased version of #54 with added test.

Closes #54.


sils1297:
bug fix, won't mention this in CHANGES

PR #309, opened 2015-12-30
Validate Locale parameter formats while constructing

### LINKED ISSUE #283: Locale __init__/parse inconsistencies
Babel 2.1.1, Python 2.7.5
<hr>
`Locale.parse('en_US')` returns `Locale('en', territory='US')`
but
`Locale('en_US')` returns `Locale('en_US')`
when I would expect
`Locale('en_US')` returns `Locale('en', territory='US')` or `UnknownLocaleError`
<hr>
`Locale.parse(12)` intentionally raises `TypeError: Unxpected value for identifier: 12`
but
`Locale(12)` causes an uncaught `TypeError`
```
babel/core.pyc, line 150, __init__
    identifier = str(self)
babel/core.pyc, line 339, __str__
    return get_locale_identifier((self.language, self.territory, self.script, self.variant))
babel/core.pyc, line 943, get_locale_identifier
    return sep.join(filter(None, (lang, script, territory, variant)))
TypeError: sequence item 0: expected string, int found
```

akx:
This avoids being able to construct mongrel locales with `language = 'en_US', territory = None`.

Fixes #283


akx:
Come to think of it, this paves the way nicely for https://github.com/python-babel/babel/issues/30; we can now derive a `FuzzyLocale` class or similar that doesn't validate as strictly. :)

dan-may:
> akx:
> Come to think of it, this paves the way nicely for #30; we can now derive a FuzzyLocale class or similar > that doesn't validate as strictly. :)

Is there some kind of roadmap as to when fuzzy locales will be implemented? I have a requirement that needs the ability to define custom/fuzzy locales, so it would be great to know when I can implement this.

akx:
Moved this out of the next point release due to the warning above.

PR #62, opened 2013-10-14
Refactored frontend module and removed duplicate code

bacher09:
Frontend module repeats  functions for distutils and cli interface, I tried to merge this parts of code.

PR #311, opened 2015-12-30
Refactor frontend.py (remove duplicate code)

akx:
This PR is directly inspired by #62.

A little bit of intricate, fine magic is used to create an Optparse parser out of the Distutils commands, a little bit of fine-tuning here and there and then everything . . . just works, with a third less code.

The `,fuzzy` changes in the tests are actually an oversight that this PR fixes as a side effect: The commit e85e71d22f4ef6c8f117cbe2d16758b20125cc0d had added `catalog.fuzzy = False` to the distutils command code, but the same change had never made it into the CLI code.

This should do wonders for test coverage statistics, which also means the frontend code will be more easily maintained.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -16,6 +16,6 @@ test-env
 babel/global.dat
 babel/global.dat.json
 tests/messages/data/project/i18n/long_messages.pot
-tests/messages/data/project/i18n/temp.pot
+tests/messages/data/project/i18n/temp*
 tests/messages/data/project/i18n/en_US
 /venv*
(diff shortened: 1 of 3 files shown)


akx:
@sils1297 Feel free to merge this..?

sils1297:
@akx I don't have time to do a full review :/ it seems well tested and a good improvement, if you're confident, feel free to merge

akx:
I'm pretty confident. Let's see how it goes :)

PR #123, opened 2014-11-23
Documented all valid keys for get_global()

erikr:
This adds a simple list of all valid keys for `get_global()` to the documentation. I've already included the `territory_languages` key, which is introduced in PR #122.

PR #315, opened 2016-01-03
Territory languages

akx:
This PR is a rebase of and supersedes #122 and #123.

### DIFF
--- a/babel/languages.py
+++ b/babel/languages.py
@@ -12,6 +12,9 @@ def get_official_languages(territory, regional=False, de_facto=False):
 
     If the `de_facto` flag is set, then languages which are "de facto" official are also returned.
 
+    .. warning:: Note that the data is as up to date as the current version of the CLDR used
+                 by Babel.  If you need scientifically accurate information, use another source!
+
     :param territory: Territory code
     :type territory: str
     :param regional: Whether to return regionally official languages too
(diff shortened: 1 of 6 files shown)


etanol:
Even though the population percents may be informative, I have the feeling that it will become outdated very quickly.  Therefore, unless we can keep up with the bi-yearly CLDR release schedule, we may be left behind again (like when we supported CLDR 23) and, then, such data will not be very useful.

On the other hand, the supported languages by territory will hopefully change in a less dramatic way.

One solution could be to add a **big fat disclaimer** note regarding the possible lack of accuracy for the population percent information.

akx:
@etanol: I'll make an issue of adding that big fat disclaimer.

If you think this is good to merge otherwise, go ahead and do it :)

babel/languages.py
```diff
@@ -0,0 +1,61 @@
...
+    if regional:
+        allowed_stati.add("official_regional")
+    if de_facto:
+        allowed_stati.add("de_facto_official")
+
+    pairs = []
+    for language, info in get_global("territory_languages").get(territory, {}).items():
+        if info["official_status"] in allowed_stati:
+            pairs.append((info["population_percent"], language))
+    pairs.sort(reverse=True)
```
etanol: I would be inclined to write:

```python
languages = get_global("territory_languages").get(territory, {})
pairs = [(info['population_percent'], language)
             for language, info in langs.iteritems()
                 if info['official_status'] in allowed_stati]
pairs.sort(reverse=True)
```

It should be a little faster and a bit more Pythonic to make more extensive use of comprehensions.

akx:
Switched to an admittedly more pythonic and more or less just as readable list comprehension as suggested by @etanol.

Since patch coverage is 100% and overall coverage is increasing, I'll just go ahead and merge this :)

PR #316, opened 2016-01-04
Implement date interval formatting

### LINKED ISSUE #276: Add support for formatting intervals
The LDML has defined an `intervalFormats` for locale-aware interval formatting since [UTS#35 revision 10](http://unicode.org/reports/tr35/tr35-10.html). Since Babel has added support for formatting durations (timedeltas) it would be convenient to support intervals as well.

akx:
Nb: There are some plumbing commits at the top -- the changes there are required by the actual interval code.

Fixes #276

### DIFF
--- a/babel/dates.py
+++ b/babel/dates.py
@@ -1041,6 +1041,25 @@ class DateTimeFormat(object):
         else:
             raise KeyError('Unsupported date/time field %r' % char)
 
+    def extract(self, char):
+        char = str(char)[0]
+        if char == 'y':
+            return self.value.year
+        elif char == 'M':
+            return self.value.month
+        elif char == 'd':
+            return self.value.day
+        elif char == 'H':
+            return self.value.hour
+        elif char == 'h':
+            return (self.value.hour % 12 or 12)
+        elif char == 'm':
+            return self.value.minute
+        elif char == 'a':
+            return int(self.value.hour >= 12)  # 0 for am, 1 for pm
+        else:
+            raise NotImplementedError("Not implemented: extracting %r from %r" % (char, self.value))
+
     def format_era(self, char, num):
         width = {3: 'abbreviated', 4: 'wide', 5: 'narrow'}[max(3, num)]
         era = int(self.value.year >= 0)
--- a/babel/dates.py
+++ b/babel/dates.py
@@ -1215,27 +1215,57 @@ def parse_pattern(pattern):
         return pattern
 
     result = []
+
+    for tok_type, tok_value in tokenize_pattern(pattern):
+        if tok_type == "chars":
+            result.append(tok_value.replace('%', '%%'))
+        elif tok_type == "field":
+            fieldchar, fieldnum = tok_value
+            limit = PATTERN_CHARS[fieldchar]
+            if limit and fieldnum not in limit:
+                raise ValueError('Invalid length for field: %r'
+                                 % (fieldchar * fieldnum))
+            result.append('%%(%s)s' % (fieldchar * fieldnum))
+        else:
+            raise NotImplementedError("Unknown token type: %s" % tok_type)
+
+    return DateTimePattern(pattern, u''.join(result))
+
+
+def tokenize_pattern(pattern):
+    """
+    Tokenize date format patterns.
+
+    Returns a list of (token_type, token_value) tuples.
+
+    ``token_type`` may be either "chars" or "field".
+
+    For "chars" tokens, the value is the literal value.
+
+    For "field" tokens, the value is a tuple of (field character, repetition count).
+
+    :param pattern: Pattern string
+    :type pattern: str
+    :rtype: list[tuple]
+    """
+    result = []
     quotebuf = None
     charbuf = []
     fieldchar = ['']
     fieldnum = [0]
 
     def append_chars():
-        result.append(''.join(charbuf).replace('%', '%%'))
+        result.append(('chars', ''.join(charbuf).replace('\0', "'")))
         del charbuf[:]
 
     def append_field():
-        limit = PATTERN_CHARS[fieldchar[0]]
-        if limit and fieldnum[0] not in limit:
-            raise ValueError('Invalid length for field: %r'
-                             % (fieldchar[0] * fieldnum[0]))
-        result.append('%%(%s)s' % (fieldchar[0] * fieldnum[0]))
+        result.append(('field', (fieldchar[0], fieldnum[0])))
         fieldchar[0] = ''
         fieldnum[0] = 0
 
     for idx, char in enumerate(pattern.replace("''", '\0')):
         if quotebuf is None:
-            if char == "'": # quote started
+            if char == "'":  # quote started
                 if fieldchar[0]:
                     append_field()
                 elif charbuf:
@@ -1257,10 +1287,10 @@ def parse_pattern(pattern):
                 charbuf.append(char)
 
         elif quotebuf is not None:
-            if char == "'": # end of quote
+            if char == "'":  # end of quote
                 charbuf.extend(quotebuf)
                 quotebuf = None
-            else: # inside quote
+            else:  # inside quote
                 quotebuf.append(char)
 
     if fieldchar[0]:
@@ -1268,4 +1298,4 @@ def parse_pattern(pattern):
     elif charbuf:
         append_chars()
 
-    return DateTimePattern(pattern, u''.join(result).replace('\0', "'"))
+    return result
--- a/babel/dates.py
+++ b/babel/dates.py
@@ -1299,3 +1299,62 @@ def tokenize_pattern(pattern):
         append_chars()
 
     return result
+
+
+def untokenize_pattern(tokens):
+    """
+    Turn a date format pattern token stream back into a string.
+
+    This is the reverse operation of ``tokenize_pattern``.
+
+    :type tokens: Iterable[tuple]
+    :rtype: str
+    """
+    output = []
+    for tok_type, tok_value in tokens:
+        if tok_type == "field":
+            output.append(tok_value[0] * tok_value[1])
+        elif tok_type == "chars":
+            if not any(ch in PATTERN_CHARS for ch in tok_value):  # No need to quote
+                output.append(tok_value)
+            else:
+                output.append("'%s'" % tok_value.replace("'", "''"))
+    return "".join(output)
+
+
+def split_interval_pattern(pattern):
+    """
+    Split an interval-describing datetime pattern into multiple pieces.
+
+    > The pattern is then designed to be broken up into two pieces by determining the first repeating field.
+    - http://www.unicode.org/reports/tr35/tr35-dates.html#intervalFormats
+
+    >>> split_interval_pattern(u'E d.M. \u2013 E d.M.')
+    [u'E d.M. \u2013 ', 'E d.M.']
+    >>> split_interval_pattern("Y 'text' Y 'more text'")
+    ["Y 'text '", "Y 'more text'"]
+    >>> split_interval_pattern(u"E, MMM d \u2013 E")
+    [u'E, MMM d \u2013 ', u'E']
+    >>> split_interval_pattern("MMM d")
+    ['MMM d']
+    >>> split_interval_pattern("y G")
+    ['y G']
+    >>> split_interval_pattern(u"MMM d \u2013 d")
+    [u'MMM d \u2013 ', u'd']
+
+    :param pattern: Interval pattern string
+    :return: list of "subpatterns"
+    """
+
+    seen_fields = set()
+    parts = [[]]
+
+    for tok_type, tok_value in tokenize_pattern(pattern):
+        if tok_type == "field":
+            if tok_value[0] in seen_fields:  # Repeated field
+                parts.append([])
+                seen_fields.clear()
+            seen_fields.add(tok_value[0])
+        parts[-1].append((tok_type, tok_value))
+
+    return [untokenize_pattern(tokens) for tokens in parts]
--- a/babel/dates.py
+++ b/babel/dates.py
@@ -882,6 +882,107 @@ def format_timedelta(delta, granularity='second', threshold=.85,
     return u''
 
 
+def _format_fallback_interval(start, end, skeleton, tzinfo, locale):
+    if skeleton in locale.datetime_skeletons:  # Use the given skeleton
+        format = lambda dt: format_skeleton(skeleton, dt, tzinfo, locale=locale)
+    elif all((isinstance(d, date) and not isinstance(d, datetime)) for d in (start, end)):  # Both are just dates
+        format = lambda dt: format_date(dt, locale=locale)
+    elif all((isinstance(d, time) and not isinstance(d, date)) for d in (start, end)):  # Both are times
+        format = lambda dt: format_time(dt, tzinfo=tzinfo, locale=locale)
+    else:
+        format = lambda dt: format_datetime(dt, tzinfo=tzinfo, locale=locale)
+
+    formatted_start = format(start)
+    formatted_end = format(end)
+
+    if formatted_start == formatted_end:
+        return format(start)
+
+    return (
+        locale.interval_formats.get(None, "{0}-{1}").
+        replace("{0}", formatted_start).
+        replace("{1}", formatted_end)
+    )
+
+
+def format_interval(start, end, skeleton, tzinfo=None, locale=LC_TIME):
+    """
+    Format an interval between two instants according to the locale's rules.
+
+    >>> format_interval(date(2016, 1, 15), date(2016, 1, 17), "yMd", locale="fi")
+    u'15.\u201317.1.2016'
+
+    >>> format_interval(time(12, 12), time(16, 16), "Hm", locale="en_GB")
+    '12:12 \u2013 16:16'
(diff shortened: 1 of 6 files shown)


scripts/import_cldr.py
```diff
@@ -16,6 +16,9 @@
 import os
 import re
 import sys
+
+from babel.dates import split_interval_pattern
```
etanol: This must go below the `sys.path` manipulation, i.e., together with the other `babel` imputs.  Otherwise, you may import a **different** Babel version.

babel/dates.py
```diff
@@ -694,7 +718,7 @@ def format_time(time=None, format='medium', tzinfo=None, locale=LC_TIME):
     return parse_pattern(format).apply(time, locale)
 
 
-def format_skeleton(skeleton, datetime=None, tzinfo=None, locale=LC_TIME):
+def format_skeleton(skeleton, datetime=None, tzinfo=None, fuzzy=True, locale=LC_TIME):
```
etanol: More tests needed, e.g., with `fuzzy == False` and `tzinfo != None`.  Use the `tests` package if you don't want to pollute the *docstring*.

babel/dates.py
```diff
@@ -835,6 +866,100 @@ def _iter_patterns(a_unit):
...
+        format = lambda dt: format_datetime(dt, tzinfo=tzinfo, locale=locale)
+
+    return (
+        locale.interval_formats.get(None, "{0}-{1}").
+        replace("{0}", format(start)).
+        replace("{1}", format(end))
+    )
+
+
+def format_interval(start, end, skeleton, tzinfo=None, fuzzy=True, locale=LC_TIME):
```
etanol: Again, more tests needed, at least to cover the `tzinfo` and `fuzzy` parameters.

etanol:
Looks good.  However, for the amount of newly added code, I see few test cases.

scripts/import_cldr.py
akx: Oop, good catch. I just used PyCharm's automatic import generation, so I didn't even notice...

akx:
Thanks for prodding me to add more tests, @etanol. Found a few corner-case bugs too, which are now fixed.

babel/dates.py
akx: Those are now added in a new test case file.

babel/dates.py
```diff
@@ -1221,4 +1426,123 @@ def append_field():
...
+    This uses the rules outlined in the TR35 document.
+
+    >>> match_skeleton('yMMd', ('yMd', 'yMMMd'))
+    'yMd'
+
+    >>> match_skeleton('yMMd', ('jyMMd',), allow_different_fields=True)
+    'jyMMd'
+
+    >>> bool(match_skeleton('yMMd', ('qyMMd',), allow_different_fields=False))
+    False
```
etanol: If this returns `None` just change `False` to `None` in the docstring.  The `bool()` wrapping doesn't add any real value here.

tests/test_date_intervals.py
```diff
@@ -0,0 +1,35 @@
...
+def test_format_interval_invalid_skeleton():
+    t1 = TEST_DATE
+    t2 = TEST_DATE + datetime.timedelta(days=1)
+    assert dates.format_interval(t1, t2, "mumumu", fuzzy=False, locale="fi") == u"8.1.2016\u20139.1.2016"
+
+
+def test_format_interval_no_difference():
+    t1 = TEST_DT
+    t2 = t1 + datetime.timedelta(minutes=8)
+    assert dates.format_interval(t1, t2, "yMd", fuzzy=False, locale="fi") == "8.1.2016"
```
etanol: I still don't see any tests for the `tzinfo` parameter of the `format_interval` function.

babel/dates.py
akx: An explicit `None` there won't work:

```python
>>> match_skeleton('yMMd', ('qyMMd',), allow_different_fields=False)
Expected: None
Got nothing
```

but leaving the retval out entirely works, so I've done that in a forthcoming revision of the PR.

tests/test_date_intervals.py
akx: Test added (though I don't think any Finn would ever format a timezoned time like that, but that's not the point here) :).

akx:
Comments heeded, fixes folded in and branch rebased.

Since Github predictably threw away the commit-related comments, here:

image

akx:
@sils1297, @etanol: This is ready for review again.

babel/dates.py
```diff
@@ -78,10 +178,7 @@ def get_next_timezone_transition(zone=None, dt=None):
                If not given the current time is assumed.
     """
     zone = get_timezone(zone)
-    if dt is None:
-        dt = datetime.utcnow()
-    else:
-        dt = dt.replace(tzinfo=None)
+    dt = _get_datetime(dt).replace(tzinfo=None)
```
etanol: Why does Codecov show this line as **not** covered?  It should be

babel/dates.py
```diff
@@ -994,6 +1161,25 @@ def __getitem__(self, name):
...
+        if char == 'y':
+            return self.value.year
+        elif char == 'M':
+            return self.value.month
+        elif char == 'd':
+            return self.value.day
+        elif char == 'H':
+            return self.value.hour
+        elif char == 'h':
+            return (self.value.hour % 12 or 12)
```
etanol: Case not covered.

babel/dates.py
```diff
@@ -1221,4 +1448,122 @@ def append_field():
...
+    # TODO: maybe implement pattern expansion?
+
+    # Based on the implementation in
+    # http://source.icu-project.org/repos/icu/icu4j/trunk/main/classes/core/src/com/ibm/icu/text/DateIntervalInfo.java
+
+    # Filter out falsy values and sort for stability; when `interval_formats` is passed in, there may be a None key.
+    options = sorted(option for option in options if option)
+
+    if 'z' in skeleton and not any('z' in option for option in options):
+        skeleton = skeleton.replace('z', 'v')
```
etanol: This is also not covered.  Would it be interesting to add a case for this line?

akx:
Coverage added. :)

babel/dates.py
akx: Added doctest: `match_skeleton('hmz', ('hmv',))`

babel/dates.py
akx: Added `test_format_interval_12_hour()`.

babel/dates.py
```diff
@@ -38,6 +38,106 @@
 time_ = time
 
 
+def _get_dt_and_tzinfo(dt_or_tzinfo):
```
jtwang: This refactor is great!

babel/dates.py
```diff
@@ -707,6 +754,9 @@ def format_skeleton(skeleton, datetime=None, tzinfo=None, locale=LC_TIME):
     u'dim. 1 avr.'
     >>> format_skeleton('MMMEd', t, locale='en')
     u'Sun, Apr 1'
+    >>> format_skeleton('yMMd', t, locale='fi')  # yMMd is not in the Finnish locale; yMd gets used
```
jtwang: For demonstration purposes, I think it would be good to also have this call again, except with fuzzy=False

babel/dates.py
```diff
@@ -835,6 +889,119 @@ def _iter_patterns(a_unit):
...
+            skeleton = match_skeleton(skeleton, interval_formats)
+        else:
+            skeleton = None
+        if not skeleton:  # Still no match whatsoever?
+            # > Otherwise, format the start and end datetime using the fallback pattern.
+            return _format_fallback_interval(start, end, skeleton, tzinfo, locale)
+
+    skel_formats = interval_formats[skeleton]
+
+    if start == end:
```
jtwang: This case should be documented. Bonus points for a doctest. :)

babel/dates.py
```diff
@@ -835,6 +889,119 @@ def _iter_patterns(a_unit):
...
+    >>> format_interval(time(12, 12), time(16, 16), "Hm", locale="en_GB")
+    '12:12 \u2013 16:16'
+
+    >>> format_interval(time(5, 12), time(16, 16), "hm", locale="en_US")
+    '5:12 AM \u2013 4:16 PM'
+
+    >>> format_interval(time(16, 18), time(16, 24), "Hm", locale="it")
+    '16:18\u201316:24'
+
+    Unknown skeletons fall back to "default" formatting.
```
jtwang: Example with raw format please. Eg. 'MMM d'

tests/test_date_intervals.py
```diff
@@ -0,0 +1,53 @@
...
+
+def test_format_interval_same_instant_3():
+    assert dates.format_interval(TEST_TIME, TEST_TIME, "xxx", fuzzy=False, locale="fi") == "11.46.15"
+
+
+def test_format_interval_same_instant_4():
+    assert dates.format_interval(TEST_DATE, TEST_DATE, "xxx", fuzzy=False, locale="fi") == "8.1.2016"
+
+
+def test_format_interval_invalid_skeleton():
```
jtwang: Please add a test for skeleton == None

babel/dates.py
```diff
@@ -835,6 +889,119 @@ def _iter_patterns(a_unit):
...
+        # > is, this allows for adjusting the string value field's width,
+        # > including adjusting between "MMM" and "MMMM", and using different
+        # > variants of the same field, such as 'v' and 'z'.
+        if skeleton and fuzzy:
+            skeleton = match_skeleton(skeleton, interval_formats)
+        else:
+            skeleton = None
+        if not skeleton:  # Still no match whatsoever?
+            # > Otherwise, format the start and end datetime using the fallback pattern.
+            return _format_fallback_interval(start, end, skeleton, tzinfo, locale)
```
jtwang: If I'm reading this right...
- if skeleton is invalid and start == end, this method will return an interval string.
- if skeleton is valid and start == end, this method will return a single datetime string. (from l:975)

This seems a bit odd to me.

Also, please add test cases to pin down the behaviour. :)

babel/dates.py
```diff
@@ -835,6 +889,119 @@ def _iter_patterns(a_unit):
...
+
+    skel_formats = interval_formats[skeleton]
+
+    if start == end:
+        return format_skeleton(skeleton, start, tzinfo, fuzzy=fuzzy, locale=locale)
+
+    start = _ensure_datetime_tzinfo(_get_datetime(start), tzinfo=tzinfo)
```
jtwang: I wonder if we could use a decorator for these tzinfo kwargs.

Just an idle thought, feel free to ignore. :)

babel/dates.py
```diff
@@ -835,6 +889,119 @@ def _iter_patterns(a_unit):
...
+
+    start_fmt = DateTimeFormat(start, locale=locale)
+    end_fmt = DateTimeFormat(end, locale=locale)
+
+    # > If a match is found from previous steps, compute the calendar field
+    # > with the greatest difference between start and end datetime. If there
+    # > is no difference among any of the fields in the pattern, format as a
+    # > single date using availableFormats, and return.
+
+    for field in PATTERN_CHAR_ORDER:  # These are in largest-to-smallest order
```
jtwang: This logic is weird enough that it might be worth pulling out into a separate method with lots of unit tests.

babel/dates.py
jtwang: It would be kind of neat to see a test method specifically for `extract`, but no big deal, feel free to ignore.

babel/dates.py
```diff
@@ -1142,6 +1328,13 @@ def get_week_number(self, day_of_period, day_of_week=None):
     'z': [1, 2, 3, 4], 'Z': [1, 2, 3, 4], 'v': [1, 4], 'V': [1, 4]  # zone
 }
 
+#: The pattern characters declared in the Date Field Symbol Table
+#: (http://www.unicode.org/reports/tr35/tr35-dates.html#Date_Field_Symbol_Table)
+#: in order of decreasing magnitude.
+PATTERN_CHAR_ORDER = "GyYuUQqMLlwWdDFgEecabBChHKkjJmsSAzZvV"
+
+_pattern_cache = {}
```
jtwang: Consider creating a memoization decorator for this. The decorator could come in useful later when someone picks up caching parsed Locales. :)

babel/dates.py
```diff
@@ -1221,4 +1448,125 @@ def append_field():
...
+    """
+    Split an interval-describing datetime pattern into multiple pieces.
+
+    > The pattern is then designed to be broken up into two pieces by determining the first repeating field.
+    - http://www.unicode.org/reports/tr35/tr35-dates.html#intervalFormats
+
+    >>> split_interval_pattern(u'E d.M. \u2013 E d.M.')
+    [u'E d.M. \u2013 ', 'E d.M.']
+    >>> split_interval_pattern("Y 'text' Y 'more text'")
+    ["Y 'text '", "Y 'more text'"]
```
jtwang: I'm surprised there's a space in the first `'text '`

jtwang: Would you please add a few more examples?

Eg. `E, MMM d – E, MMM d, y G`, `MMM d – d` (thanks for putting up with my en-centric examples <3)

scripts/import_cldr.py
```diff
@@ -693,6 +697,21 @@ def main():
...
 
 
+def parse_interval_formats(data, tree):
+    # http://www.unicode.org/reports/tr35/tr35-dates.html#intervalFormats
+    interval_formats = data.setdefault("interval_formats", {})
+    for elem in tree.findall("dateTimeFormats/intervalFormats/*"):
+        if elem.tag == "intervalFormatFallback":
+            interval_formats[None] = elem.text
+        elif elem.tag == "intervalFormatItem":
+            skel_data = interval_formats.setdefault(elem.attrib["id"], {})
```
jtwang: Should we filter out draft elements? Babel currently ignores them for a lot of settings, although I haven't checked if it ignores them for all settings.

babel/dates.py
akx: #305... :) The solution there is a metaclass that overrides the class ctor, though, so it's not directly applicable.

babel/dates.py
akx: Re the surprising space: That's because the pattern is tokenized, split and untokenized, and tokenization joins spans of character literals into one character literal token. The result is the same though -- `' ' ' '` (quoted space, unquoted space, quoted space) === `'   '` (3 quoted spaces) === `   ` (3 unquoted spaces), etc.

babel/dates.py
akx: That sounds like it would make static analysis/code completion unnecessarily harder.

babel/dates.py
akx: Hmm, currently no, we don't.

If there is no skeleton match, the `medium` format is the fallback.

babel/dates.py
akx: `_format_fallback_interval` has an equality check too.

* if skeleton is valid and the instants are the same, the skeleton's format is used to format a single datetime string
* if skeleton is valid and the instants are not the same, the TR35 greatest-difference algorithm is used
* if skeleton is not valid and the instants are not the same, `_format_fallback_interval` formats both instants as is suitable for their types and joins them with the fallback format (usually `{0} - {1}`).
* if skeleton is not valid and the instants are the same, `_format_fallback_interval` formats the  instant as is suitable for its type and returns it.

I think the `test_format_interval_same_instant_*` tests cover this.

babel/dates.py
akx: I don't find it that weird, but OTOH I'm used to us Finns formatting intervals taking that into account:

> `format_interval(date(2016, 1, 15), date(2016, 1, 17), "yMd", locale="fi")`
> '15.-17.1.2016'

Also, I'm not sure it's really worth it, considering how awkward it might make the `format_interval` function. Could be done later if we find corner cases where intervals are rendered wrong?

scripts/import_cldr.py
akx: Sure. We're doing that for currency formats anyway.

akx:
Thanks for the round of reviews, @jtwang :)

CLDR import and null-skeleton behaviors were folded in to the respective commits.
The new tests live in the last "more tests" commit.

tests/test_date_intervals.py
akx: Test added.

babel/dates.py
akx: Doctested!

babel/dates.py
akx: Done.

babel/dates.py
jtwang: Sounds good, thanks for the explanation!

babel/dates.py
jtwang: That is true. Tradeoffs...

babel/dates.py
jtwang: Yep, sounds reasonable to me!

babel/dates.py
jtwang: As I posted in gitter: "also idk what i was smoking, CLDR doesn't have raw formats as interval IDs. sorry about that!" :P

babel/dates.py
jtwang: oh boy, metaclasses... ::grabs an aspirin::

Issue #318, opened 2016-01-07
update_catalog removes translators


ajaeger:
Since the release of babel 2.2.0, we see in OpenStack that the header comments get modified and the list of translators gets removed:

```patch
--- a/neutron/locale/tr_TR/LC_MESSAGES/neutron.po
+++ b/neutron/locale/tr_TR/LC_MESSAGES/neutron.po
@@ -1,15 +1,11 @@
 # Turkish (Turkey) translations for neutron.
 # Copyright (C) 2015 ORGANIZATION
 # This file is distributed under the same license as the neutron project.
+# FIRST AUTHOR <EMAIL@ADDRESS>, 2015.
 #
-# Translators:
-# ADİL REŞİT DURSUN <ardursun@deltanoc.com>, 2015
-# Alper Çiftçi <alprciftci@gmail.com>, 2015
-# Andreas Jaeger <jaegerandi@gmail.com>, 2015. #zanata
-# OpenStack Infra <zanata@openstack.org>, 2015. #zanata
 msgid ""
 msgstr ""
```

neutron.pot file starts with, it's created with extract_messages:

```po
# Translations template for neutron.
# Copyright (C) 2015 ORGANIZATION
# This file is distributed under the same license as the neutron project.
# FIRST AUTHOR <EMAIL@ADDRESS>, 2015.
#
#, fuzzy
msgid ""
msgstr ""
"Project-Id-Version: neutron 7.0.1.dev41\n"
"Report-Msgid-Bugs-To: EMAIL@ADDRESS\n"
```

How can we keep the translators in here?

Reference:
https://review.openstack.org/#/c/263140/4/neutron/locale/tr_TR/LC_MESSAGES/neutron.po

akx:
This is probably introduced by the otherwise seemingly harmless https://github.com/python-babel/babel/commit/e0e7ef168856bb431076cfe45f6c456d12091233 commit.

Are you able to patch your local Babel to revert that change and see if that fixes it? If so, I suppose we'll have to add a switch for this behavior...

ajaeger:
Indeed, reverting that change does not cause the problem.

akx:
Thanks for helping out there, @ajaeger. The fix is in the referenced PR #320.

ajaeger:
Thanks a lot for the fast fix, @akx !

sils1297:
[...]

PR #320, opened 2016-01-08
Make catalog header updating an option

akx:
The change in e0e7ef168856bb had an unexpected and likely undesired effect
when updating catalogs with, e.g. translator names in the header comment.

It's better to make the updating an option and revert back to the pre-2.2
behavior by default.

Fixes https://github.com/python-babel/babel/issues/318

### DIFF
--- a/babel/messages/catalog.py
+++ b/babel/messages/catalog.py
@@ -673,7 +673,7 @@ class Catalog(object):
         if key in self._messages:
             del self._messages[key]
 
-    def update(self, template, no_fuzzy_matching=False):
+    def update(self, template, no_fuzzy_matching=False, update_header_comment=False):
         """Update the catalog based on the given template catalog.
 
         >>> from babel.messages import Catalog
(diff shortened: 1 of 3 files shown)


sils1297:
@akx needs more test coverage.

akx:
@sils1297 The lack of test coverage would be fixed by merging #311. Right now the distutils versions of command code are not tested at all, but the other copies are.

I published the option only in the distutils code anticipating that merge.

ajaeger:
Tested this locally and it works as expected - fixes issue #318 in my setup. Thanks!

akx:
I can also just _remove_ the new option in the distutils code, but that seems like doing it just to make Codecov happy, and we're not writing code for CI robots. :)

sils1297:
@akx then lets merge #311 first?

akx:
Hey @ajaeger, if you have the time, could you check that this branch still works for your UC (after it was rebased)? Thank you!

ajaeger:
@akx tested current head that includes the patch and it works fine. Thanks

Issue #253, opened 2015-09-24
Message extraction is slow for sufficiently large repos (consider parallelizing)


ENuge:
A large repo I am working with currently takes ~30 seconds to build using xgettext (https://www.gnu.org/savannah-checkouts/gnu/gettext/manual/html_node/xgettext-Invocation.html). The same scan with pybabel extract takes ~2 min 49 secs. 

The extraction process is, in both cases, blacklist-based, with the same directories blacklisted. That is, we scan the entire repo for new messages except those in a list of [ignore: foo/**.py] blocks.

ENuge:
Without having looked too deeply into the code, I think it could be possible to split the calls inside extract_from_dir(..) to extract_from_file(..) into multiple processes, one per filename (somewhere near here: https://github.com/python-babel/babel/blob/master/babel/messages/extract.py#L143).

ENuge:
Poked at the extraction code, wanted to get a sense of how much time is spent scanning and extracting strings vs actually writing them to the new file.

Hypothesis (and hope): scanning/extracting takes significantly longer. This part of the process we can parallelize. We can't have concurrent writes to the same output file.

Results:
Total time *extracting*: 159.926753521 secs
Total time *writing*: 0.707601547241 secs
Code I ran: http://pastebin.com/5uSC15MN

etanol:
Thanks for the numbers, looks like an interesting case for optimisation.  Once our workflow is a bit smoother, I will take a look on it.

PR #324, opened 2016-01-12
Support extraction by filename as well as directory

ENuge:
Reason for the change: https://github.com/python-babel/babel/issues/253 . This obviates the need for parallelizing to make things faster (on the consumer's side, we only scan files that are different in a given branch compared to origin's master, which brings scan time down from ~4min 30s to ~5-20s, depending on how old one's branch is). 

Code changes: moved half of the logic from extract_from_dir into its own function. Then added a few "is this a path or directory?" type checks to frontend.py to use the right thing. Those checks are kind of ugly right now but I couldn't think of a more elegant way of doing it. Added/rejiggered a couple of tests.

### DIFF
--- a/babel/messages/extract.py
+++ b/babel/messages/extract.py
@@ -142,28 +142,72 @@ def extract_from_dir(dirname=None, method_map=DEFAULT_MAPPING,
         dirnames.sort()
         filenames.sort()
         for filename in filenames:
-            filename = relpath(
-                os.path.join(root, filename).replace(os.sep, '/'),
-                dirname
-            )
-            for pattern, method in method_map:
-                if pathmatch(pattern, filename):
-                    filepath = os.path.join(absname, filename)
-                    options = {}
-                    for opattern, odict in options_map.items():
-                        if pathmatch(opattern, filename):
-                            options = odict
-                    if callback:
-                        callback(filename, method, options)
-                    for lineno, message, comments, context in \
-                          extract_from_file(method, filepath,
-                                            keywords=keywords,
-                                            comment_tags=comment_tags,
-                                            options=options,
-                                            strip_comment_tags=
-                                                strip_comment_tags):
-                        yield filename, lineno, message, comments, context
-                    break
+            filepath = os.path.join(root, filename).replace(os.sep, '/')
+
+            for message_tuple in check_and_call_extract_file(
+                filepath,
+                method_map,
+                options_map,
+                callback,
+                keywords,
+                comment_tags,
+                strip_comment_tags,
+                dirpath=absname,
+            ):
+                yield message_tuple
+
+
+def check_and_call_extract_file(filepath, method_map, options_map,
+                                callback, keywords, comment_tags,
+                                strip_comment_tags, dirpath=None):
+    """Checks if the given file matches an extraction method mapping, and if so, calls extract_from_file.
+
+    Note that the extraction method mappings are based relative to dirpath.
+    So, given an absolute path to a file `filepath`, we want to check using
+    just the relative path from `dirpath` to `filepath`.
+
+    :param filepath: An absolute path to a file that exists.
+    :param method_map: a list of ``(pattern, method)`` tuples that maps of
+                       extraction method names to extended glob patterns
+    :param options_map: a dictionary of additional options (optional)
+    :param callback: a function that is called for every file that message are
+                     extracted from, just before the extraction itself is
+                     performed; the function is passed the filename, the name
+                     of the extraction method and and the options dictionary as
+                     positional arguments, in that order
+    :param keywords: a dictionary mapping keywords (i.e. names of functions
+                     that should be recognized as translation functions) to
+                     tuples that specify which of their arguments contain
+                     localizable strings
+    :param comment_tags: a list of tags of translator comments to search for
+                         and include in the results
+    :param strip_comment_tags: a flag that if set to `True` causes all comment
+                               tags to be removed from the collected comments.
+    :param dirpath: the path to the directory to extract messages from.
+    """
+    # filename is the relative path from dirpath to the actual file
+    filename = relpath(filepath, dirpath)
+
+    for pattern, method in method_map:
+        if not pathmatch(pattern, filename):
+            continue
+
+        options = {}
+        for opattern, odict in options_map.items():
+            if pathmatch(opattern, filename):
+                options = odict
+        if callback:
+            callback(filename, method, options)
+        for message_tuple in extract_from_file(
+            method, filepath,
+            keywords=keywords,
+            comment_tags=comment_tags,
+            options=options,
+            strip_comment_tags=strip_comment_tags
+        ):
+            yield (filename, ) + message_tuple
+
+        break
 
 
 def extract_from_file(method, filename, keywords=DEFAULT_KEYWORDS,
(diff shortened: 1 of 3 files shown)


sils1297:
Hi, can you fix the issues pointed out by the gitmate bot and refactor your stuff into atomic commits as described in http://coala.readthedocs.org/en/latest/Getting_Involved/Writing_Good_Commits/ ?

sils1297:
Thanks for submitting :)

ENuge:
Appeased the LineLengthBot with Yelp's preferred syntax (which is vertically long, but eh this line is super-awkward regardless).

I'll refactor my commits later today or tomorrow. (I started this change on Yelp's old fork of babel, so this version was more of a copy/paste job. But I'll make the history make some sense.)

sils1297:
@ENuge cool, ping us when it's done.

babel/messages/frontend.py
```diff
@@ -333,29 +337,48 @@ def run(self):
...
+                if os.path.isdir(path):
+                    extracted = extract_from_dir(
+                        path, method_map, options_map,
+                        keywords=self._keywords,
+                        comment_tags=self.add_comments,
+                        callback=callback,
+                        strip_comment_tags=self.strip_comments
+                    )
+                else:
+                    extracted = check_and_call_extract_file(path, method_map,
```
akx: Could you use the same indent style as in the `extract_from_dir` call above? :)

babel/messages/frontend.py
```diff
@@ -333,29 +337,48 @@ def run(self):
...
+                                                            options_map, callback, self._keywords,
+                                                            self.add_comments, self.strip_comments)
                 for filename, lineno, message, comments, context in extracted:
-                    filepath = os.path.normpath(os.path.join(dirname, filename))
+                    if os.path.isdir(path):
+                        filepath = os.path.normpath(os.path.join(path, filename))
+                    else:
+                        filepath = filename
```
akx: Add a comment like `  # already normalized`

babel/messages/extract.py
```diff
@@ -142,28 +142,74 @@ def extract_from_dir(dirname=None, method_map=DEFAULT_MAPPING,
...
+
+    for pattern, method in method_map:
+        if pathmatch(pattern, filename):
+            options = {}
+            for opattern, odict in options_map.items():
+                if pathmatch(opattern, filename):
+                    options = odict
+            if callback:
+                callback(filename, method, options)
+            for lineno, message, comments, context in \
```
akx: Maybe don't extract the tuple here?

```python
for message_tuple in extract...:
   yield (filename,) + message_tuple
```

babel/messages/extract.py
```diff
@@ -142,28 +142,74 @@ def extract_from_dir(dirname=None, method_map=DEFAULT_MAPPING,
...
+    :param dirpath: the path to the directory to extract messages from.
+    """
+    if dirpath is None:
+        dirpath = os.getcwd()
+
+    # filename is the relative path from dirpath to the actual file
+    filename = relpath(filepath, dirpath)
+
+    for pattern, method in method_map:
+        if pathmatch(pattern, filename):
```
akx: Would be better to invert this if (avoids over-indentation):

```python
if not pathmatch(pattern, filename):
  continue
options = ....
```

babel/messages/extract.py
```diff
@@ -142,28 +142,74 @@ def extract_from_dir(dirname=None, method_map=DEFAULT_MAPPING,
...
+                     tuples that specify which of their arguments contain
+                     localizable strings
+    :param comment_tags: a list of tags of translator comments to search for
+                         and include in the results
+    :param strip_comment_tags: a flag that if set to `True` causes all comment
+                               tags to be removed from the collected comments.
+    :param dirpath: the path to the directory to extract messages from.
+    """
+    if dirpath is None:
+        dirpath = os.getcwd()
```
akx: This should not be done in this function; it's fine to just do it in the caller.

babel/messages/extract.py
```diff
@@ -142,28 +142,74 @@ def extract_from_dir(dirname=None, method_map=DEFAULT_MAPPING,
...
+                     positional arguments, in that order
+    :param keywords: a dictionary mapping keywords (i.e. names of functions
+                     that should be recognized as translation functions) to
+                     tuples that specify which of their arguments contain
+                     localizable strings
+    :param comment_tags: a list of tags of translator comments to search for
+                         and include in the results
+    :param strip_comment_tags: a flag that if set to `True` causes all comment
+                               tags to be removed from the collected comments.
+    :param dirpath: the path to the directory to extract messages from.
```
akx: Can you expand a little on this -- namely that this is used to calculate the relative names of files?

babel/messages/frontend.py
```diff
@@ -300,17 +300,21 @@ def finalize_options(self):
...
             ]).keys()
 
-        if not self.input_dirs:
-            raise DistutilsOptionError("no input directories specified")
+        if not self.input_paths:
+            raise DistutilsOptionError("no input files or directories specified")
+
+        for path in self.input_paths:
+            if not os.path.isdir(path) and not os.path.isfile(path):
+                raise DistutilsOptionError("Input path: %s is not a file or directory" % path)
```
akx: Is this even possible? :D

babel/messages/extract.py
```diff
@@ -142,28 +142,74 @@ def extract_from_dir(dirname=None, method_map=DEFAULT_MAPPING,
...
-                                            keywords=keywords,
-                                            comment_tags=comment_tags,
-                                            options=options,
-                                            strip_comment_tags=
-                                                strip_comment_tags):
-                        yield filename, lineno, message, comments, context
-                    break
+            filepath = os.path.join(root, filename).replace(os.sep, '/')
+
+            for (
```
akx: Don't expand the tuple here either.

```python
for message_tuple in check...:
    yield message_tuple
```

is fine and dandy!

akx:
And aside from the handful of comments I had, great work @ENuge ! :)

I'd appreciate a test that passes in a direct filename to the command line command though. :)

babel/messages/frontend.py
ENuge: Heh, this is meant to be checking if the path exists. I didn't really think about the much more direct way of doing that...namely, using os.path.exists(..). Will update.

babel/messages/extract.py
ENuge: Cool, I'll add a comment/docstring line with the return type or something to make up for the loss in explicitness.

babel/messages/frontend.py
ENuge: Yup! Done.

babel/messages/extract.py
ENuge: fwiw extract_from_dir(..) above it does the same check. Though I'm fine with having the caller explicitly pass it in.

akx:
@ENuge: Looking good!

Please squash all of the commits into one though :)

sils1297:
Interesting, travis errored out on mac only for pypy 2.6, maybe a nondeterministic bug?

akx:
@sils1297 Heh, yes... race condition against the clock, it seems. If the wallclock ticks forward a second while those tests are being run, they may fail.

[Should probably freeze time for the duration of those tests.](https://pypi.python.org/pypi/freezegun)

akx:
Super nice work @ENuge!

I'm gonna merge this, but come to think of it, we should add an alias for `input-paths` called `input-dirs` for backward compatibility. I'll make an issue out of that. :)

PR #334, opened 2016-01-22
Extract API now takes 'npgettext' keyword

dan-may:
Initial implementation; needs testing.


akx:
Nice :) I assume this works for your own original use case?

Would you be interested in writing a test for this too?

akx:
Closed via #337.

Issue #330, opened 2016-01-15
Extract: add input-dirs compat alias for new input-paths


akx:
We don't want to break CLI API between minor releases, so add an input-dirs argument alias for the newly renamed input-paths.

Refs #324

PR #335, opened 2016-01-27
Frontend: Add multi-domain support to compile_catalog command

ansiwen:
Some projects have their translations split up into several text
domains within one package.  This change adds the possibility to
specify a space seperated list of domains in the configuration, like
'setup.py compile_catalog --domain="foo bar"', for instance.

### DIFF
--- a/babel/messages/frontend.py
+++ b/babel/messages/frontend.py
@@ -83,7 +83,7 @@ class compile_catalog(Command):
     description = 'compile message catalogs to binary MO files'
     user_options = [
         ('domain=', 'D',
-         "domain of PO file (default 'messages')"),
+         "domains of PO files (space separated list, default 'messages')"),
         ('directory=', 'd',
          'path to base directory containing the catalogs'),
         ('input-file=', 'i',
@@ -118,6 +118,12 @@ class compile_catalog(Command):
                                        'or the base directory')
 
     def run(self):
+        domains = self.domain.split()
+
+        for domain in domains:
+            self._run_domain(domain)
+
+    def _run_domain(self, domain):
         po_files = []
         mo_files = []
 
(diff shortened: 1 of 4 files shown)


ansiwen:
56% patch coverage? That's not fair, just because I had to indent the whole block? 😒

babel/messages/frontend.py
```diff
@@ -118,77 +118,82 @@ def finalize_options(self):
...
-                        mo_files.append(os.path.join(self.directory, locale,
-                                                     'LC_MESSAGES',
-                                                     self.domain + '.mo'))
-        else:
-            po_files.append((self.locale, self.input_file))
-            if self.output_file:
-                mo_files.append(self.output_file)
+        domains = self.domain.split()
+
+        for domain in domains:
```
akx: It would make for a cleaner patch and cleaner code if you just split `def run_domain(self, domain)` into a separate fn, I think?

ansiwen: right, good point, will do that

babel/messages/frontend.py
```diff
@@ -118,6 +118,12 @@ def finalize_options(self):
                                        'or the base directory')
 
     def run(self):
+        domains = self.domain.split()
+
+        for domain in domains:
+            self.__run_domain(domain)
+
+    def __run_domain(self, domain):
```
akx: Ah, please avoid double-underscore-prefixed names. Just `_run_domain` or `run_domain` would be fine!

akx:
@ansiwen Heh, don't worry about codecov's patch measurements. It's a cranky ol' thing sometimes.

babel/messages/frontend.py
ansiwen: What's the reason for that? I learned, double-underscore-prefix automatically adds the classname as prefix, so you have a kind of "private" member.

akx: Sure, but that's also very nonintuitive behavior.

[Even PEP 8 advises against using the mangly names unless you have a good reason.](https://www.python.org/dev/peps/pep-0008/#id42)

As for the privateness viewpoint, Python's attitude generally is "we're all consenting adults here", i.e. no need to go to lengths to hide the member. The single underscore is already a signal that this is some sort of private API, and you should know what you're doing if you're calling or overriding this.

akx:
Thank you @ansiwen :heart_eyes_cat:

ansiwen:
Cool, thanks for the merge. When can I expect a release with this feature? (Since we need it for openstack.)

sils1297:
@akx as there is demand for a release I'm in favour of doing that soon.

@ansiwen if @akx disagrees we can still do a development release that you can get via pypi if you specify the version exactly in your requirements.txt or install with `--pre`

akx:
@ansiwen @sils1297 https://github.com/python-babel/babel/issues/330 will need to be implemented before release (it's a seriously short fix).

ansiwen:
@sils1297 we will wait for a regular release.

Issue #281, opened 2015-11-12
Add keyword specification for npgettext


tsouvarev:
Something like:
`npgettext: ((1, 'c'), 2, 3)`
(django-babel already [supports](https://github.com/python-babel/django-babel/commit/5442d21f57c78944b59651534c9664a2c8ff61b7) it)

PR #337, opened 2016-01-28
Fix #328 #281 Adds npgettext to default keywords

### LINKED ISSUE #328: npgettext strings not extracting
I've been creating a comprehensive POC for translating a Flask app, which implements Flask-Babel. My issue is with the pybabel extraction command however, so I am posting the issue here.

I have tested the variations of gettext and they all work except for npgettext, where strings marked up with this method are not appearing in my POT file when extracted. The strings appear when rendered, though, so the issue seems to be limited to the extraction API.

My POC can be found here: https://github.com/dan-may/translation

sudheesh001:


akx:
This is a duplicate of #334, which is fine in this instance.

However I'll have to ask you to add a test case that proves the features of the `npgettext` ([these Ruby docs seem to document it](http://api.rubyonrails.org/v2.3/classes/I18n/Gettext/Helpers.html#M002786))  are extracted correctly :)

sudheesh001:
> #334

@akx Oh I didn't see that issue  before making this pull request, this would be my first patch to babel and i've looked through the `test_extract.py` which is where I'd have to ideally be making the test case. Looking at the ruby docs you've provided, `npgettext` should be written as
```py
buf = BytesIO(b"""\
    msg = npgettext('Strings', [ 'pylon',  # TRANSLATORS: shouldn't be
                               'pylons' ], # TRANSLATORS: seeing this
                                count)
  """)
```
followed by
```py
messages = list(extract.extract_python(buf, ('npgettext',),
                                               ['TRANSLATORS:'], {}))
```
but I get a `[]` , Am I in the right direction ?

akx:
No worries about the other issue, just wanted to link it up here! :)

I think the invocation (in the `BytesIO`) should _probably_ be
```python
npgettext("Strings", "pylon", "pylons", count)
```

I.e. no list; that's the same signature as in the [Django docs for `npgettext`](https://docs.djangoproject.com/en/1.9/ref/utils/#django.utils.translation.npgettext) (sorry for not mentioning it earlier; I honestly didn't find that one before).

Would that make the extraction pass?

sudheesh001:
Looks like there's a small problem when I push any changes, redoing the PR

PR #338, opened 2016-02-01
Fix #325 Support more format for timezone

### LINKED ISSUE #325: Support more format for timezone
# Overview Description

I cannot use `XXXXX` or `ZZZZZ`

# Steps to Reproduce

See this `pycon` log

```pycon
>>> from babel.dates import format_datetime, get_timezone
>>> dt = datetime(2016, 1, 13, 7, 8, 35)
>>> tz = get_timezone('Asia/Seoul')
>>> dt = tz.localize(dt)
>>> format_datetime(dt, 'ZZZZZ')
  File "/Users/item4/.virtualenvs/lektor/lib/python2.7/site-packages/babel/dates.py", line 618, in format_datetime
    return parse_pattern(format).apply(datetime, locale)
  File "/Users/item4/.virtualenvs/lektor/lib/python2.7/site-packages/babel/dates.py", line 1192, in parse_pattern
    append_field()
  File "/Users/item4/.virtualenvs/lektor/lib/python2.7/site-packages/babel/dates.py", line 1156, in append_field
    % (fieldchar[0] * fieldnum[0]))
ValueError: Invalid length for field: 'ZZZZZ'
>>> format_datetime(dt, 'ZZZ')
u'+0900'
>>> format_datetime(dt, 'XXXXX')
u'XXXXX'
>>> format_datetime(dt, 'XXX')
u'XXX'
```

# Actual Results
`ValueError: Invalid length for field: 'ZZZZZ'` ... etc.

# Expected Results

`ZZZZZ` → `+09:00`
`XXXXX` → `+09:00`

# Reproducibility

100%, Repeat my code.

# Additional Information:

Python 2.7
Mac OS X

See [Date Field Symbol Table](http://unicode.org/reports/tr35/tr35-dates.html#Date_Field_Symbol_Table) for spec.

sachinpali146:
Added the Support for timezone format 'ZZZZZ' and 'X'


babel/dates.py
```diff
@@ -395,7 +395,9 @@ def get_timezone_gmt(datetime=None, width='long', locale=LC_TIME):
 
     :param datetime: the ``datetime`` object; if `None`, the current date and
                      time in UTC is used
-    :param width: either "long" or "short"
+    Added new parameter for width to resolve bug #325
```
akx: This doesn't need to be in the docstring.

babel/dates.py
```diff
@@ -1267,9 +1271,9 @@ def format_milliseconds_in_day(self, num):
         msecs = self.value.microsecond // 1000 + self.value.second * 1000 + \
                 self.value.minute * 60000 + self.value.hour * 3600000
         return self.format(msecs, num)
-
+#Added 5: 'extra' in width to fix issue #https://github.com/python-babel/babel/issues/325
```
akx: This comment is also unnecessary.

babel/dates.py
```diff
@@ -406,6 +408,8 @@ def get_timezone_gmt(datetime=None, width='long', locale=LC_TIME):
     hours, seconds = divmod(seconds, 3600)
     if width == 'short':
         pattern = u'%+03d%02d'
+    elif width== 'extra':                   #Condition added to take 'ZZZZZ' datetime format
```
akx: The comment here is also unnecessary.

babel/dates.py
```diff
@@ -395,7 +395,9 @@ def get_timezone_gmt(datetime=None, width='long', locale=LC_TIME):
 
     :param datetime: the ``datetime`` object; if `None`, the current date and
                      time in UTC is used
-    :param width: either "long" or "short"
+    Added new parameter for width to resolve bug #325
+    :param width: either "long" or "short" or "extra"
```
akx: Perhaps `iso8601` would be a better name for this width, for the TR35 document seems to state the goal is to have ZZZZZ be an ISO8601 compliant timezone formatting.

babel/dates.py
```diff
@@ -406,6 +408,8 @@ def get_timezone_gmt(datetime=None, width='long', locale=LC_TIME):
     hours, seconds = divmod(seconds, 3600)
     if width == 'short':
         pattern = u'%+03d%02d'
+    elif width== 'extra':                   #Condition added to take 'ZZZZZ' datetime format
+        pattern = u'%+03d:%02d'
```
akx: [The TR35 document on ZZZZZ](http://www.unicode.org/reports/tr35/tr35-dates.html#Time_Zone_Goals) seems to say that a zero offset should always be represented as a single "Z", not "+00:00".

Could you fix this to do that? It would probably be worth it to add a check for this just after the total `seconds` is calculated above; simply something like

```python
if width == "iso8601" and seconds == 0:
    return "Z"
```
would probably do fine.)

babel/dates.py
```diff
@@ -1282,6 +1286,14 @@ def format_timezone(self, char, num):
                 return get_timezone_name(self.value.tzinfo, width,
                                          uncommon=True, locale=self.locale)
             return get_timezone_location(self.value.tzinfo, locale=self.locale)
+        # Included additional elif condition to add support for 'X' in timezone format
+        elif char in ['x','X']:
```
akx: Use `elif char in "xX"` or `elif char == "x" or char == "X"` instead of constructing a list here.

babel/dates.py
```diff
@@ -1282,6 +1286,14 @@ def format_timezone(self, char, num):
...
+            if num==1:
+                return get_timezone_gmt(self.value, width='short', locale=self.locale)[:-2]
+            elif num in [2,4]:
```
akx: Use a tuple `num in (2, 4)` or a simple `or` statement here (as in my comment above).

babel/dates.py
```diff
@@ -1323,7 +1335,9 @@ def get_week_number(self, day_of_period, day_of_week=None):
             week_number += 1
         return week_number
 
-
+#Added 'X' and 'x' in PATTERN_CHARS for to include datetime format support for them.
```
akx: These comments are unnecessary too. This information is stored in the Git history; it needn't be in a comment in the file.

babel/dates.py
```diff
@@ -1336,7 +1350,7 @@ def get_week_number(self, day_of_period, day_of_week=None):
     'h': [1, 2], 'H': [1, 2], 'K': [1, 2], 'k': [1, 2],             # hour
     'm': [1, 2],                                                    # minute
     's': [1, 2], 'S': None, 'A': None,                              # second
-    'z': [1, 2, 3, 4], 'Z': [1, 2, 3, 4], 'v': [1, 4], 'V': [1, 4]  # zone
+    'z': [1, 2, 3, 4, 5], 'Z': [1, 2, 3, 4, 5], 'v': [1, 4], 'V': [1, 4],  'x': [1, 2, 3, 4, 5], 'X': [1, 2, 3, 4, 5]# zone
```
akx: This line becomes awkwardly long; move `v`, `V`, `x` and `X` onto a new line.

akx: The new characters (`x`, `X`) should also be included in `PATTERN_CHAR_ORDER`, below.

akx:
@sachinpali146 Thank you for the patch!

However please review the code formatting comments made by @gitmate-bot, above, and my inline comments about the code itself.

In addition, I don't see test cases for the new formatting; could you please add some? Extra credit if some tests match the examples in the TR35 (such as `ZZZZZ` formatting the `America/Los Angeles` timezone as `-08:00`).

sachinpali146:
@akx Thanks for the suggestions, I will work on them. I am working on open source project for first time so I will be needing some guidance.

babel/dates.py
```diff
@@ -1282,6 +1284,16 @@ def format_timezone(self, char, num):
                 return get_timezone_name(self.value.tzinfo, width,
                                          uncommon=True, locale=self.locale)
             return get_timezone_location(self.value.tzinfo, locale=self.locale)
+        # Included additional elif condition to add support for 'X' in timezone format
+        elif char in "xX":
+            if get_timezone_gmt(self.value, width='short', locale=self.locale)=='+0000' and char == 'X':
```
akx: Hmm... It still feels kind of wasteful to have `get_timezone_gmt` go through actually formatting the date just so its return value could be turned into a zero :)

I think it might be better to actually add an argument to `get_timezone_gmt`, something like `return_z` (optional, false by default), which, when set to True, would cause its logic to short-circuit to returning `"Z"` directly after it has calculated that the offset would in fact be zero.

What do you think?

babel/dates.py
```diff
@@ -1282,6 +1284,16 @@ def format_timezone(self, char, num):
...
+                return 'Z'
+            elif num == 1:
+                return get_timezone_gmt(self.value, width='short', locale=self.locale)[:-2]
```
akx: This will return incorrect values for timezones that are not an integer number of hours away from GMT.

The TR-35 document states that the minutes part is _optional_, i.e. if it were `00`, it will be omitted, not that it's always truncated out.

babel/dates.py
```diff
@@ -1167,7 +1169,7 @@ def __getitem__(self, name):
             return self.format_frac_seconds(num)
         elif char == 'A':
             return self.format_milliseconds_in_day(num)
-        elif char in ('z', 'Z', 'v', 'V'):
+        elif char in ('z', 'Z', 'v', 'V', 'x', 'X', 'O'):
```
akx: `O` is added here, but is there actual support for it in `format_timezone`?

babel/dates.py
sachinpali146: Yes I was working on that, but I have one concern in function get_timezone_gmt. When we use width='short' it is returning some thing like '-0700' where as it should return 'GMT-7'. Is this bug?

babel/dates.py
sachinpali146: Yes, initially I was also thinking to add some argument in get_timezone_gmt function, should I go ahead and add the functionality?

babel/dates.py
sachinpali146: It's not completed yet.

babel/dates.py
akx: [The field table](http://unicode.org/reports/tr35/tr35-dates.html#Date_Field_Symbol_Table) has good examples of all the cases.

If `width="short"` returns a value like `-0700`, it looks to be appropriate for Z/ZZ/ZZZ/X/XX/XXXX/x/xx/xxxx, so it's not wrong per se.

It looks like "GMT-7" or whatever is a localized/localizable GMT format, so that's another thing that would just need support added. I think that's outside the scope of this PR, but it's something to keep in mind.

babel/dates.py
akx: Okay. I think we could add support for `O`/`OOOO` in another PR though?

babel/dates.py
sachinpali146: Actually I had tried to coverup every format for Zone in this PR only and "GMT-7" is output for 'O'

So should I create a new function for localized/localizable GMT format and a new PR?

akx: The reason I'd suggest to leave "O" formats for another PR is that [localizable GMT formats seem to be more complicated than Z/X](http://unicode.org/reports/tr35/tr35-dates.html#Time_Zone_Format_Terminology), since they need new data to be read from the CLDR XML files into our locale .dat files, and interpretation of that data.

Let's finish this first and then consider adding support for O? :)

sachinpali146: Yeah we can do that. Btw I had already added the support for 'OOOO' as it was already available.

sachinpali146:
@akx Can you please help me in understanding how codecov works and why commits are failing.

akx:
@sachinpali146 Sure. 

Codecov measures the [code coverage](https://en.wikipedia.org/wiki/Code_coverage) for the tests that are being run for the code.

In this particular case,

> codecov/patch - 66% of diff hit (target 80%)

means that of the lines you have changed, only 66% are covered by the test suite.

Likewise,

> codecov/project - 89.03% (-0.44%) compared to 89.47%

means that while the complete test coverage of the project had been 89.47% before this PR, with the added code it's dropped down to 89.03%.

What this means, all in all, is that there's new code that is left untested, and the way to fix it is to add tests that cover all of the new behavior.

So -- could you write tests to cover the new behavior you've implemented?

sachinpali146:
@akx Thanks for the guidance about code coverage. I had added the test cases for the patch and it covered 100% patch code and also increased the project code coverage :)

babel/dates.py
```diff
@@ -395,8 +402,12 @@ def get_timezone_gmt(datetime=None, width='long', locale=LC_TIME):
 
     :param datetime: the ``datetime`` object; if `None`, the current date and
                      time in UTC is used
-    :param width: either "long" or "short"
+    :param width: either "long" or "short" or "iso8601"
     :param locale: the `Locale` object, or a locale string
+    :param return_z: True or False; Function returns indicator "Z"
+                     when local time offset is 0
+    :param x_format: True or False; Function returns The ISO8601 basic format
```
akx: The `z` flag is fine, but I think this flag needs to be rethought, since it looks like it only affects one situation anyway.

Maybe make this another `width`? `iso8601_short`, for instance?

So `iso8601` would always return the +HH:MM format, and `iso8601_short` would return either `+HH` or `+HH:MM`?

tests/test_dates.py
```diff
@@ -250,6 +250,129 @@ def test_with_float(self):
         formatted_string = dates.format_datetime(epoch, format='long', locale='en_US')
         self.assertEqual(u'April 1, 2012 at 3:30:29 PM +0000', formatted_string)
 
+    def test_with_pattern(self):
```
akx: This seems to only be testing time zone formats, so rename to, say, `test_timezone_formats`? :)

akx:
_Very_ impressive work with the tests, good job! :tada: 

There's still a couple things I'd like to address; see the inline comments, and also pay attention to Gitmate's stylistic complaints. (You seem to be missing a couple whitespaces after commas and so on. :) )

babel/dates.py
```diff
@@ -437,7 +454,10 @@ def get_timezone_location(dt_or_tzinfo=None, locale=LC_TIME):
                          the timezone; if `None`, the current date and time in
                          UTC is assumed
     :param locale: the `Locale` object, or a locale string
-    :return: the localized timezone name using location format
+    :param return_city: True or False
```
akx: The parameter description should include what it does too. Maybe something like

> Return only the exemplar city for the time zone when true

?

akx:
@sachinpali146 The PR is starting to look _very_ good now!

Could you just squash the changes into a couple commits with messages following the [Writing Good Commits](http://coala.readthedocs.org/en/latest/Getting_Involved/Writing_Good_Commits/) rules? Maybe

* one commit adding the new features to `get_timezone_gmt`, `get_timezone_location` and `get_timezone_name` (with the respective tests)
* one commit adding the new format characters to the formatter class (with the respective tests)

---

If you're unfamiliar with rewriting commits like this, try the [Git Book's Interactive Staging](https://git-scm.com/book/en/v2/Git-Tools-Interactive-Staging) tutorial. I'd personally do something like (assuming we're starting from this branch's HEAD and a remote called `upstream` refers to this repository):

```shell
$ git fetch --all # (get all changes from all remotes)
$ git rebase upstream/master # (rebase your changes on top of the current master; there should not be any conflicts to resolve)
$ git reset upstream/master # (reset your repo to a state where your changes are intact in the working tree but the commits aren't there)
$ git add -p # (include the first set of changes)
$ git commit # (write a fantastic commit message)
$ git add -u # (to add the rest of the updated files)
$ git commit # (write another fantastic message)
$ git push -f # (force-push to your PR branch)
```

babel/dates.py
sachinpali146: I did added that info but at wrong location, Should I move it up?
 ` :param return_city: True or False`
   `:return: the localized timezone name using location format;`
   `          if return_city is True then return
            The exemplar city (location) for the time zone`

akx:
ack 949e4cb, b759e08

akx:
ack 949e4cb, b759308

akx:
Thank you very much for the contributions, @sachinpali146! And thanks for putting up with my nitpicking. This is an exemplary patch. :)

Merging this -- have a great weekend!

sachinpali146:
Thank you @akx  for your patience, I had learned lot of thing by working on this PR and your continuous guidance. It is your nitpicking which has helped me to become open source society contributor :)

PR #341, opened 2016-02-03
Extract: Add npgettext to default keywords map

sudheesh001:
Add the npgettext as a default keyword in the keywords map.
Create test_npgettext in test_extract showing the passing tests.

Fixes https://github.com/python-babel/babel/issues/281
Fixes https://github.com/python-babel/babel/issues/328

### DIFF
--- a/babel/messages/extract.py
+++ b/babel/messages/extract.py
@@ -39,7 +39,7 @@ DEFAULT_KEYWORDS = {
     'dngettext': (2, 3),
     'N_': None,
     'pgettext': ((1, 'c'), 2),
-    'npgettext': ((1, 'c', 2, 3))
+    'npgettext': ((1, 'c'), 2, 3)
 }
 
 DEFAULT_MAPPING = [('**.py', 'python')]
--- a/babel/messages/extract.py
+++ b/babel/messages/extract.py
@@ -38,7 +38,8 @@ DEFAULT_KEYWORDS = {
     'dgettext': (2,),
     'dngettext': (2, 3),
     'N_': None,
-    'pgettext': ((1, 'c'), 2)
+    'pgettext': ((1, 'c'), 2),
+    'npgettext': ((1, 'c', 2, 3))
 }
 
 DEFAULT_MAPPING = [('**.py', 'python')]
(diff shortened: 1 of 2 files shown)


sudheesh001:
@akx I hope this is the fix that's expected.
I closed https://github.com/python-babel/babel/pull/337 and opened this pull request.

akx:
@sudheesh001 Thanks! I'll take a closer look later tonight.

babel/messages/extract.py
```diff
@@ -38,7 +38,8 @@
     'dgettext': (2,),
     'dngettext': (2, 3),
     'N_': None,
-    'pgettext': ((1, 'c'), 2)
+    'pgettext': ((1, 'c'), 2),
+    'npgettext': ((1, 'c', 2, 3))
```
akx: I think the tuple should be `((1, 'c'), 2, 3)`?

I.e. that the first argument is `c`ontext and 2 and 3 are messages.

tests/messages/test_extract.py
```diff
@@ -60,6 +60,15 @@ def test_nested_comments(self):
                                                ['TRANSLATORS:'], {}))
         self.assertEqual([(1, 'ngettext', (u'pylon', u'pylons', None), [])],
                          messages)
+        buf = BytesIO(b"""\
```
akx: These tests are nice enough, but can you move them to another function, `def test_npgettext(self):` for instance? :)

tests/messages/test_extract.py
```diff
@@ -131,6 +140,17 @@ def test_multiline(self):
         self.assertEqual([(1, 'ngettext', (u'pylon', u'pylons', None), []),
                           (3, 'ngettext', (u'elvis', u'elvises', None), [])],
                          messages)
+        buf = BytesIO(b"""\
```
akx: These tests are nice enough, but can you move them to another function, `def test_npgettext(self):` for instance? :)

sils1297:
@sudheesh001 also the same commit message guidelines apply here too (although they're not that strictly enforced)

sudheesh001:
@akx @sils1297 I hope i've fixed it, i've rebased and updated this PR.

akx:
This looks good!

Thank you for the patch, @sudheesh001 !

Issue #250, opened 2015-09-23
Add pypy3 support


sils1297:
IMO this is important and it *should* ideally work without modifications except the version check in setup.py. (Also see https://github.com/python-babel/babel/issues/171 )

PR #343, opened 2016-02-07
Set up official support for PyPy 3

akx:
As suspected in #250, Babel runs fine on PyPy3, it's just that `setup.py` complains. No more!

Fixes https://github.com/python-babel/babel/issues/250

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -26,6 +26,8 @@ matrix:
     - os: linux
       python: pypy
     - os: linux
+      python: pypy3
+    - os: linux
       python: 3.3
     - os: linux
       python: 3.4
--- a/setup.py
+++ b/setup.py
@@ -1,8 +1,8 @@
 # -*- coding: utf-8 -*-
 
-import os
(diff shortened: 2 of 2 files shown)


akx:
Assigned @sils1297 as the original owner of #250 :)

sils1297:
@akx functionally good, few nitpicks

akx:
@sils1297 Took another approach here -- instead of strictly checking for versions, only the bit that might fail with unsupported versions (importing `__version__`) has an exception check :)

In addition, to shorten setup.py a little, I removed the dead documentation building bits.

akx:
@sils1297 Re not doing an explicit version check: I looked through some prominent Python libraries and it does seem it's not very common to do a version check like that. (Which, I know, isn't the strongest of arguments...)

Anyway, if we see some sort of elevated issue rate due to this change, we can always begin checking for versions more rigidly again.

akx:
Merging as per comments in earlier iterations of PR.

PR #347, opened 2016-02-13
Additional Support for Date Time pattern formats

sachinpali146:


akx:
Can you remove the merge commit from this PR? It should be as easy as

```shell
git fetch --all
git rebase upstream/master
```

babel/dates.py
```diff
@@ -1272,8 +1276,11 @@ def format_weekday(self, char, num):
                 return self.format(value % 7 + 1, num)
             num = 3
         weekday = self.value.weekday()
-        width = {3: 'abbreviated', 4: 'wide', 5: 'narrow'}[num]
-        context = {3: 'format', 4: 'format', 5: 'stand-alone'}[num]
+        width = {3: 'abbreviated', 4: 'wide', 5: 'narrow', 6: 'short'}[num]
+        if char == 'c':
```
akx: Can you add some sort of reference about the semantics of lower-case `c` using stand-alone context? It doesn't seem immediately obvious, esp. considering this is a format function...

sachinpali146:
I will work on it on Monday

babel/dates.py
sachinpali146: @akx I screwed up the  commit once again. Please help

akx:
@sachinpali146:

doing

```
git fetch --all
git rebase upstream/master
```

seems to untangle the history enough.

You can then do `git reset upstream/master` to return to a state where all of your changes exist, but have not been committed yet.

sachinpali146:
@akx Made some changes, please check

tests/test_dates.py
```diff
@@ -155,6 +160,28 @@ def test_local_day_of_week_standalone(self):
         fmt = dates.DateTimeFormat(d, locale='bn_BD')
         self.assertEqual('4', fmt['c']) # friday is first day of week
 
+    def test_pattern_day_of_week(self):
+        dt = datetime(2016,2,6)
+        fmt = dates.DateTimeFormat(dt, locale='en_US')
```
akx: I'd love to see the "local day of week" formats (`e` and `c`) being tested with a locale that has another start-of-week-day than `en_US`'s Sunday.

Otherwise, this PR looks very good!

akx:
Nice work! Thanks @sachinpali146 :)

PR #348, opened 2016-02-14
Refactor CLDR importer script

akx:
The CLDR importer script was mostly one function of nearly 500 lines, which was frankly silly.

This PR breaks it into multiple functions that do one bit of importing before passing on the proverbial torch to another import function, et cetera. The script behaves exactly as before wrt the resulting locale data files, though!

As the commit messages say,

> Special care was taken to order the new functions in the same order as the original source was; the resulting whitespace-ignoring diff is very minimal.

So, before you are too daunted to review this, **please use `git diff -w` or the `?w=1` query parameter on GitHub:**

```shell
$ git diff --stat master
 scripts/import_cldr.py | 929 +++++++++++++++++++++++++------------------------
 1 file changed, 474 insertions(+), 455 deletions(-)

$ git diff --stat -w master
 scripts/import_cldr.py | 199 +++++++++++++++++++++++++++----------------------
 1 file changed, 109 insertions(+), 90 deletions(-)
```

### DIFF
--- a/scripts/import_cldr.py
+++ b/scripts/import_cldr.py
@@ -365,346 +365,381 @@ def _process_local_datas(sup, srcdir, destdir, force=False, dump_json=False):
         if locale_id in ordinal_rules:
             data['ordinal_form'] = ordinal_rules[locale_id]
 
-        # <localeDisplayNames>
+        parse_locale_display_names(data, tree)
 
-        territories = data.setdefault('territories', {})
-        for elem in tree.findall('.//territories/territory'):
-            if ('draft' in elem.attrib or 'alt' in elem.attrib) \
-                    and elem.attrib['type'] in territories:
-                continue
-            territories[elem.attrib['type']] = _text(elem)
-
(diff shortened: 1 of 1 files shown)


jtwang:
One thing that would help me immensely is to split argument parsing out of the main method. It's a patch we've been applying in our fork in order to build successfully.

Totally OK if you want to punt on this, I can make the change myself later on. :)

```
def import_cldr(srcdir, destdir, force=False, dump_json=False):
   # blah blah blah

if __name__ == '__main__':
     # blah blah parse arguments
     import_cldr(srcdir, destdir, force, dump_json)
```

jtwang:
Done a round of reviews, love this change!

jtwang:
Looks good, other than the unused arg. Fix and ship!

jtwang:
Oh, yes, mega thanks for splitting argparsing and processing logic! <3 <3

akx:
Merging based on previous greenlight :)

PR #349, opened 2016-02-15
Day period API

akx:
This PR adds full support for day periods other than am/pm.

Requested by @sachinpali146 for additional time format character support. :)

See: http://unicode.org/reports/tr35/tr35-dates.html#Day_Period_Rules


RauliL:
LGTM.

akx:
(@RauliL is a colleague who I recruited to do code review here.)

PR #351, opened 2016-02-17
Unittest locale_identifiers() and exists() bijection.

kdeldycke:
I just had a little doubt about the relationship between `locale_identifiers()` and `exists()`, so I made a tiny test to make sure the two refers to the same locale IDs taxonomy.


akx:
:+1:

kdeldycke:
Thanks @akx ! That was fast ! :smiley_cat:

Issue #350, opened 2016-02-16
format_datetime using 'S...S' format is incorrect


bfordham-ubnt:
When using a format string for format_datetime or format_time using microseconds 'S...', the displayed results are **always** incorrect if the datetime's microseconds has less than 5 digits.

```python
from datetime import datetime
d = datetime(2016, 2, 15, 8, 3, 1, 799)
print format_datetime(d, 's.S')
>>> 1.8       #should be 1.0
print format_datetime(d, 's.SSSS')
>>> 1.7990   # should be 1.0008
```

in babel.dates
```python
def format_frac_seconds(self, num):
        value = str(self.value.microsecond)
        return self.format(round(float('.%s' % value), num) * 10**num, num)
```
should be something like

```python
def format_frac_seconds(self, num):
        value = self.value.microsecond // 1000000
        return self.format(round(value, num) * 10**num, num)
```

akx:
@bfordham-ubnt Thanks for the bug report! Since you already have the relevant patch (I took the liberty of adding the proper markup around it) in the issue here, would you be interested in writing a pull request too? If you don't have the time or inclination, it's fine too; just say so, and someone'll pick this up eventually. :)

iamshubh22:
@bfordham-ubnt I would want to work on this.

akx:
@iamshubh22 Go right ahead! :)

akx:
Marking as in progress for @iamshubh22. (I'm unable to set @iamshubh22 as the assignee for the issue,  but anyway.)

PR #353, opened 2016-02-22
format_datetime: using 'S...S' format is incorrect

iamshubh22:
Fixes https://github.com/python-babel/babel/issues/350

### DIFF
--- a/babel/dates.py
+++ b/babel/dates.py
@@ -1287,8 +1287,20 @@ class DateTimeFormat(object):
         return get_period_names(locale=self.locale)[period]
 
     def format_frac_seconds(self, num):
-        value = str(self.value.microsecond)
-        return self.format(round(float('.%s' % value), num) * 10**num, num)
+        value = self.value.microsecond / 1000000
+        return self.format(round(value, num) * 10**num, num)
+        """ Returns fractional seconds. Converts microseconds received to the \
+            nearest round-off value according to the given precesion value.
+
+            Earlier the function computes wrong result if the microseconds \
+            received are less than 5 digits.
+
+            d = datetime(2016, 2, 15, 8, 3, 1, 799)
+            print format_datetime(d, 's.S')
+            >>> 1.0         # previous function computes 1.8
+            print format_datetime(d, 's.SSSS')
+            >>> 1.0008      # previous function computes 1.7990
+        """
 
     def format_milliseconds_in_day(self, num):
         msecs = self.value.microsecond // 1000 + self.value.second * 1000 + \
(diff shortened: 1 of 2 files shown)


akx:
This is missing test cases.

You would need to prove that the earlier behavior is wrong and that this patch fixes it. :)

iamshubh22:
@akx Oh thanks!! I will see to it. :smile:

akx:
Hey @iamshubh22, okay, so now you have a test case, but I'd still like to see some sort of reasoning (maybe in the form of a reference to the particular piece of specification?) that explains why this new behavior is correct :)

In addition, a couple more test cases wouldn't go amiss, explaining the different cases that the newly implemented behavior covers.

babel/dates.py
```diff
@@ -1287,8 +1287,20 @@ def format_period(self, char):
         return get_period_names(locale=self.locale)[period]
 
     def format_frac_seconds(self, num):
-        value = str(self.value.microsecond)
-        return self.format(round(float('.%s' % value), num) * 10**num, num)
+        value = self.value.microsecond / 1000000
+        return self.format(round(value, num) * 10**num, num)
+        """ Returns fractional seconds. Converts microseconds received to the \
```
akx: The docstring for a function needs to be _before_ the code, not after. :)

babel/dates.py
```diff
@@ -1287,8 +1287,20 @@ def format_period(self, char):
...
+            nearest round-off value according to the given precesion value.
```
akx: Typo: it's `precision` :)

babel/dates.py
```diff
@@ -1287,8 +1287,20 @@ def format_period(self, char):
...
+
+            Earlier the function computes wrong result if the microseconds \
```
akx: This doesn't need to be in the docstring. We don't need to document that a function used to do the wrong thing.

babel/dates.py
```diff
@@ -1287,8 +1287,20 @@ def format_period(self, char):
...
+            received are less than 5 digits.
+
+            d = datetime(2016, 2, 15, 8, 3, 1, 799)
```
akx: If you are writing doctests, they should be

```pydoc
>>> foo()
result-of-foo
```
, not
```pydoc
foo()
>>> result-of-foo
```

However these shouldn't probably be here either.

babel/dates.py
iamshubh22: @akx So should I remove this docstring part??

akx: The function should look something like

```python
    def format_frac_seconds(self, num):
        """ Return fractional seconds.

        Rounds the time's microseconds to the precision
        given by the number of digits passed in.
        """
        value = self.value.microsecond / 1000000
        return self.format(round(value, num) * 10**num, num)
```

or similar.

PR #354, opened 2016-02-23
exposing of cldr units, measurement systems and character order

alexbodn:
please merge these for use in ui and in pint, the python measurement units package.

### DIFF
--- a/babel/core.py
+++ b/babel/core.py
@@ -58,6 +58,8 @@ def get_global(key):
     - ``win_mapping``
     - ``zone_aliases``
     - ``zone_territories``
+    - ``script_metadata``
+    - ``territory_scripts``
 
     .. note:: The internal structure of the data may change between versions.
 
@@ -364,6 +366,16 @@ class Locale(object):
     def _data(self):
         if self.__data is None:
             self.__data = localedata.LocaleDataDict(localedata.load(str(self)))
+            #: the script_metadata
+            script = self.script
+            if self.territory and not self.script:
+                territory_scripts = get_global('territory_scripts').get(self.territory, [])
+                for _script in territory_scripts:
+                    _loc = Locale.parse('und_' + _script)
+                    if _loc and _loc.language == self.language:
+                        script = _script
+                        break
+            self._RTL = get_global('script_metadata').get(script, dict()).get('RTL', 'NO')
         return self.__data
 
     def get_display_name(self, locale=None):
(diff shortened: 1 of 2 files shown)


akx:
Hello @alexbodn and thank you for the patch!

Unfortunately (as you can see above), the patch is not in a mergeable state. First, a couple raw technical points:

* You'll need to rebase this on top of the current `master`; the CLDR importer script has been refactored recently (in PR #348), and your changes seem to have been made on top of an older version.
* The two commits should be squashed into one; there's no point in having code in `master` that first adds code, only to delete it in the directly next commit.
** That being said, it would be nice if the separate features that are being added -- units and directionality -- lived in commits of their own.
* The doctests 
* The commit message(s) should be formatted [nicely](http://coala.readthedocs.org/en/wip-fneu-globlins_of_moria/Getting_Involved/Writing_Good_Commits/).

And then, some philosophical (if one can call them that) points:

One: In Babel the properties on `Locale` are not meant to be usually used for end-user consumption. That is, instead of the proposed usage 
```python
locale.unit_patterns[unit][locale.plural_form(val)].format(val)
```
a much nicer API would be a `def format_unit(val, unit, width="short", locale=None):` (or similar) that lived in `babel.numbers`.

Two: Does the CLDR also expose information about what are the preferred measurement units for a given locale? If it does, that would be very valuable information to add in the same batch!

alexbodn:
thanks @alx for your consideration.
indeed, all your comments make sense, but i haven not the faintest idea on how to proceed.
should i save my diff, pull from main babel to my babel fork, clone my fork again, and apply the good parts of my saved patch?
no that i know *how* to do these ,,,

please advice,
alex

akx:
Okay, hey @alexbodn, so I see you've been working on `master` in your own branch :)
In your working copy, do this:

```shell
# first, let's create an actual branch based on this PR
git branch units-measurements-ltr f9c9d3c85a
# now, let's ensure you can sync with the main babel (if you already have an `upstream` remote, skip this)
# since the first step created a local branch for your PR, we'll now reset your fork's `master`:
# (1) fetch the state of all remotes
git fetch --all
# (2) ensure you're on your master branch
git checkout master
# (3) hard-reset its status to the main babel's master
git reset --hard upstream/master
# now, let's work on your feature branch
git checkout units-measurements-ltr
# and now's the trickyish bit, i.e. rebasing your changes on top of the current master.
# this procedure may fail -- or at least leave you in an rebase-in-progress state.
# that's fine, though; you'll just have to fix the merge conflicts.
# (see https://git-scm.com/book/en/v2/Git-Branching-Rebasing
#    & https://githowto.com/resolving_conflicts
# )
git rebase master
# when you're finished with the rebase, your branch is in a state that would cleanly merge;
# you can now continue to work on it, adding commits to implement the suggested things.
# when you're done, you'll just
git push -u origin units-measurements-ltr
# and create a new PR. :)
```

If you feel like that's well over your head, I won't mind working on this functionality based on the work you've done here in this PR! I'll be sure to retain your original authorship in the reworked commits. Just say if you'd prefer that.

alexbodn:
thanks @akx . i will try :)

alexbodn:
hello @akx, please advice with my rebasing:

the rebasing fails on diffs that i have reverted later.
is it possible to remove these from the log, or should i 
rather take the present diff and recreate my fork?

many thanks,
alex

akx:
You can use `rebase --skip` to skip a patch, if that's what you're asking.

akx:
Closed in favor of #367 / #368.

PR #358, opened 2016-02-29
space missing between the two function

saurabhiiit:
Space Missing

### DIFF
--- a/babel/dates.py
+++ b/babel/dates.py
@@ -1287,8 +1287,13 @@ class DateTimeFormat(object):
         return get_period_names(locale=self.locale)[period]
 
     def format_frac_seconds(self, num):
-        value = str(self.value.microsecond)
-        return self.format(round(float('.%s' % value), num) * 10**num, num)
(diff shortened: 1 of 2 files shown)


saurabhiiit:
Sure, but for the better code practices the newline should be present

babel/localtime/_unix.py
```diff
@@ -26,6 +26,7 @@ def _tz_from_env(tzenv):
         raise pytz.UnknownTimeZoneError(
             "tzlocal() does not support non-zoneinfo timezones like %s. \n"
             "Please use a timezone in the form of Continent/City")
+
```
sils1297: you add trailing whitespace here!

saurabhiiit:
@sils1297 yes because of better code practices,  we need to add whitespace between the function, except this , everywhere in the code was having two whitespace between function

sils1297:
you add trailing whitespace which is not ok, this is not about newlines!

PR #360, opened 2016-02-29
format_datetime: using 'S...S' format is incorrect

iamshubh22:
Fixes https://github.com/python-babel/babel/issues/350

### DIFF
--- a/babel/dates.py
+++ b/babel/dates.py
@@ -1287,8 +1287,13 @@ class DateTimeFormat(object):
         return get_period_names(locale=self.locale)[period]
 
     def format_frac_seconds(self, num):
-        value = str(self.value.microsecond)
-        return self.format(round(float('.%s' % value), num) * 10**num, num)
+        """ Return fractional seconds.
+
+        Rounds the time's microseconds to the precision given by the number \
+        of digits passed in.
+        """
+        value = self.value.microsecond / 1000000
+        return self.format(round(value, num) * 10**num, num)
 
     def format_milliseconds_in_day(self, num):
         msecs = self.value.microsecond // 1000 + self.value.second * 1000 + \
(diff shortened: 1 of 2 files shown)


babel/dates.py
```diff
@@ -1287,8 +1287,13 @@ def format_period(self, char):
         return get_period_names(locale=self.locale)[period]
 
     def format_frac_seconds(self, num):
-        value = str(self.value.microsecond)
-        return self.format(round(float('.%s' % value), num) * 10**num, num)
+        """ Return fractional seconds.
+
+        Rounds the time's microseconds to the precision given by the number \
```
akx: No need to use the backslash for line continuations in docstrings :)

iamshubh22: @akx Ok. Got it :smile:

iamshubh22:
@akx TravisCI is failing for pypy for Linux. Why so?

akx:
@iamshubh22 Ah, nothing to worry about if I'm guessing right. I restarted that particular build now.

To be exact, the problem is that occasionally (very seldom though) time manages to roll over a second during a test, so when a test expects a file to say `the time right now` (whatever it is), it may be `the time right now + 1 second`, and that makes the test fail.

akx:
This looks good! 

Thank you for the patch, @iamshubh22! :trophy:

iamshubh22:
@akx Just one doubt. What does this Code coverage tells??

akx:
@iamshubh22:

* *codecov/patch* - 100.00% of diff hit (target 80.00%)
  * This means that all of the lines of code you've touched in the PR are also being tested.
* *codecov/project* - 89.66% remains the same compared to e1ce012
  * This means that coverage doesn't change -- as is reasonable. The previous version was 2 lines of code, the new version is 2 lines of code too.

PR #361, opened 2016-03-02
Parsing of mixed-cased locale ID.

kdeldycke:
Following #351, it seems that mixed-cased locale IDs are not properly checked as existing depending on the platform.

The issue lies in `babel.localedata.exists()` which rely on filesystem assets. As a result, mixed-cased locale IDs are properly parsed on OSX (case-insensitive filesystem) but not on Linux.

The first commit [demonstrate the issue](https://travis-ci.org/python-babel/babel/builds/113096167) depending on the system.
The second commit ensure that locale IDs recognized by Babel are uniques, even if lower-cased.
The third commit fix to the issue.

### DIFF
--- a/babel/localedata.py
+++ b/babel/localedata.py
@@ -15,6 +15,7 @@
 import os
 import threading
 from collections import MutableMapping
+from itertools import chain
 
 from babel._compat import pickle
 
@@ -24,15 +25,22 @@ _cache_lock = threading.RLock()
 _dirname = os.path.join(os.path.dirname(__file__), 'locale-data')
 
 
+def normalize_locale(name):
+    """Return a normalized locale ID or `None` if the ID is not recognized."""
+    name = name.strip().lower()
+    for locale_id in chain.from_iterable([_cache, locale_identifiers()]):
+        if name == locale_id.lower():
+            return locale_id
+
+
 def exists(name):
-    """Check whether locale data is available for the given locale.  Ther
-    return value is `True` if it exists, `False` otherwise.
+    """Check whether locale data is available for the given locale.
+
+    Returns `True` if it exists, `False` otherwise.
 
     :param name: the locale identifier string
     """
-    if name in _cache:
-        return True
-    return os.path.exists(os.path.join(_dirname, '%s.dat' % name))
+    return True if normalize_locale(name) else False
 
 
 def locale_identifiers():
--- a/babel/localedata.py
+++ b/babel/localedata.py
@@ -26,7 +26,11 @@ _dirname = os.path.join(os.path.dirname(__file__), 'locale-data')
 
 
 def normalize_locale(name):
-    """Return a normalized locale ID or `None` if the ID is not recognized."""
+    """Normalize a locale ID by stripping spaces and apply proper casing.
+
+    Returns the normalized locale ID string or `None` if the ID is not
+    recognized.
+    """
     name = name.strip().lower()
     for locale_id in chain.from_iterable([_cache, locale_identifiers()]):
         if name == locale_id.lower():
--- a/babel/localedata.py
+++ b/babel/localedata.py
@@ -44,7 +44,10 @@ def exists(name):
 
     :param name: the locale identifier string
     """
-    return True if normalize_locale(name) else False
+    if name in _cache:
+        return True
+    file_found = os.path.exists(os.path.join(_dirname, '%s.dat' % name))
+    return True if file_found else bool(normalize_locale(name))
 
 
 def locale_identifiers():
--- a/tests/test_localedata.py
+++ b/tests/test_localedata.py
@@ -13,6 +13,8 @@
 
 import doctest
 import unittest
+import random
+from operator import methodcaller
 
 from babel import localedata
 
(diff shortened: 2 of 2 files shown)


kdeldycke:
Looks like this issue relates to #283.

babel/localedata.py
```diff
@@ -24,15 +25,22 @@
 _dirname = os.path.join(os.path.dirname(__file__), 'locale-data')
 
 
+def normalize_locale(name):
+    """Return a normalized locale ID or `None` if the ID is not recognized."""
```
akx: Could you explain what "normalization" means in the docstring?

babel/localedata.py
```diff
@@ -24,15 +25,22 @@
...
+    """Check whether locale data is available for the given locale.
+
+    Returns `True` if it exists, `False` otherwise.
 
     :param name: the locale identifier string
     """
-    if name in _cache:
-        return True
-    return os.path.exists(os.path.join(_dirname, '%s.dat' % name))
+    return True if normalize_locale(name) else False
```
akx: `return bool(normalize_locale(name))` :)

akx: To be honest, I'd like to see this normalization thing as a "slow path", if the previous `os.path.exists` path fails.

babel/localedata.py
kdeldycke: Sure. Done in https://github.com/kdeldycke/babel/commit/a5cb8317626b71d532f78c8d86c6b0e790ad4b7f .

babel/localedata.py
kdeldycke: Done: https://github.com/kdeldycke/babel/commit/1a6731e2b51747764199da40bb39fcb9354dfe58

akx:
Nice, thanks @kdeldycke!

(CI: I'm not waiting for Appveyor, since most builds completed and some failed due to ephemeral errors.)

kdeldycke:
Thanks @akx for the quick merge! :smiley:

babel/localedata.py
```diff
@@ -24,15 +25,29 @@
...
+    Returns the normalized locale ID string or `None` if the ID is not
+    recognized.
+    """
+    name = name.strip().lower()
+    for locale_id in chain.from_iterable([_cache, locale_identifiers()]):
+        if name == locale_id.lower():
+            return locale_id
+
+
 def exists(name):
```
jtwang: Wow, greedy method.
I'm really looking forward to https://github.com/python-babel/babel/pull/305 !

(just an observation, please ignore :)

babel/localedata.py
```diff
@@ -24,15 +25,29 @@
...
+
+    Returns `True` if it exists, `False` otherwise.
 
     :param name: the locale identifier string
     """
     if name in _cache:
         return True
-    return os.path.exists(os.path.join(_dirname, '%s.dat' % name))
+    file_found = os.path.exists(os.path.join(_dirname, '%s.dat' % name))
+    return True if file_found else bool(normalize_locale(name))
```
jtwang: Would iteration over in-mem objects would be faster than disk access? If so, please consider performing the in-memory check first. :)

(super duper minor optimization)

tests/test_localedata.py
```diff
@@ -73,6 +75,23 @@ def test_merge():
...
     for l in localedata.locale_identifiers():
         assert localedata.exists(l)
+
+
+def test_unique_ids():
+    # Check all locale IDs are uniques.
+    all_ids = localedata.locale_identifiers()
+    assert len(all_ids) == len(set(all_ids))
+    # Check locale IDs don't collide after lower-case normalization.
+    lower_case_ids = list(map(methodcaller('lower'), all_ids))
```
jtwang: IMO

`lower_case_ids = [id.lower() for id in all_ids]` is more readable, but don't care much.

tests/test_localedata.py
```diff
@@ -73,6 +75,23 @@ def test_merge():
...
+    assert len(lower_case_ids) == len(set(lower_case_ids))
+
+
+def test_mixedcased_locale():
+    for l in localedata.locale_identifiers():
```
jtwang: Is it necessary to iterate through all ids? Would it be sufficient to spot check some deliberately selected hardcoded values? Eg. 'Fi', 'FI_fi', 'fi_FI', 'fI_fI', 'ZH_hanT_HK'

babel/localedata.py
akx: That'd complicate the whole deal unnecessarily, imho. It'd become something like

* (1) name in cache?
* (2) normalized name in cache?
* (3) name in file system?
* (4) normalized name in file system?

which would kinda interleave normalization with this "simple" existence checking...

tests/test_localedata.py
akx: Yeah, it would be, and more Pythonic, but that's kinda minor, and can be trivially corrected later if need be.

tests/test_localedata.py
akx: That's another optimization that could be done (via `pytest.parametrize`), but I don't think our test suite takes too long to run as it is :)

jtwang:
Ah, crud, didn't notice this was already merged into master. Feel free to ignore all my comments, none of them are major. :)

tests/test_localedata.py
jtwang: True that. I just get the heebie jeebies when I see `random.choice` in unit tests, ha.

akx:
All comments were duly noted and processed, don't fret :D

PR #365, opened 2016-03-07
docs/Makefile: Correct Line length

saurabhiiit:
Fixes: Line length


akx:
To be honest, I think it's clearer for these to be on a single line, really, so I'm closing the PR.

Thank you for the PR though, @saurabhiiit, and sorry it took me this long to get back to it!

PR #367, opened 2016-03-13
imported and exposed support for character order and measurement units and systems

alexbodn:
imported and exposed support for character order:
	>>> Locale('de', 'DE').character_order
	<<< u'left-to-right'
	>>> Locale('ar', 'SA').direction
	<<< u'rtl'

exposed measurement units and measurement systems translation:
	>>> locale = Locale('fr', 'FR')
	>>> val = 12
	>>> unit_length = 'long'
	>>> unit = 'length-foot' + ':' + unit_length
	>>> locale.unit_patterns[unit][locale.plural_form(val)].format(val)
	<<< u'12 pieds'
	>>> locale = Locale('fr', 'FR')
	>>> print locale.measurement_systems['US']
	<<< américain

PR #368, opened 2016-03-19
Character order and measurement systems

akx:
This is a fixed version of @alexbodn 's PR #367.

The `get_unit_name` API from that PR is intentionally excluded; a `format_unit` API will be included in a subsequent PR.


alexbodn:
thank you very much @akx .
the international web sites developers, 
and the pint measurement units project 
will enjoy these enhancements the moment you will pull them.

babel/core.py
```diff
@@ -885,6 +885,41 @@ def ordinal_form(self):
...
+        >>> Locale('fr', 'FR').measurement_systems['US']
+        u'am\\xe9ricain'
+        >>> Locale('en', 'US').measurement_systems['US']
+        u'US'
+
+        """
+        return self._data['measurement_systems']
+
+    @property
+    def character_order(self):
```
jtwang: This would have been a great use of a Python 3.4 Enum. Just an observation, no need to do anything here. :)

babel/core.py
```diff
@@ -885,6 +885,41 @@ def ordinal_form(self):
...
+        """The text direction for the language.
+
+        >>> Locale('de', 'DE').character_order
+        'left-to-right'
+        >>> Locale('ar', 'SA').character_order
+        'right-to-left'
+        """
+        return self._data['character_order']
+
+    @property
```
jtwang: This feels pretty redundant with the method above. I understand the value of providing the CSS-specific value in this package (commonly used), but I don't think it belongs in the Locale class.

We could create helper modules that provide methods like this that transform CLDR values into ones accepted by other systems. Off hand, I can't think of any other ones that might be relevant for CSS (maybe language code?), but, for instance, converting date format strings to JS compatible ones might be useful to someone.

scripts/import_cldr.py
```diff
@@ -422,6 +422,9 @@ def _process_local_datas(sup, srcdir, destdir, force=False, dump_json=False):
         parse_unit_patterns(data, tree)
         parse_date_fields(data, tree)
 
+        for elem in tree.findall('.//layout/orientation/characterOrder'):
```
jtwang: You should keep using parse wrapper methods! It's much easier to read.

Also, I was going to poke you about calling _should_skip_elem, but none of the elements in cldr 28 are draft, etc, so whatevs. :)

babel/core.py
```diff
@@ -885,6 +885,41 @@ def ordinal_form(self):
...
+    def direction(self):
+        """The text direction for the language in CSS short-hand form.
+
+        >>> Locale('de', 'DE').direction
+        'ltr'
+        >>> Locale('ar', 'SA').direction
+        'rtl'
+        """
+        order = self.character_order
+        return ''.join([word[0] for word in order.split('-')])
```
jtwang: *cough* unit test please ;)

scripts/import_cldr.py
```diff
@@ -480,6 +483,10 @@ def parse_locale_display_names(data, tree):
         for listPattern in listType.findall('listPatternPart'):
             list_patterns[listPattern.attrib['type']] = _text(listPattern)
 
+    measurement_systems = data.setdefault('measurement_systems', {})
+    for measurement_system in tree.findall('.//measurementSystemNames/measurementSystemName'):
+        measurement_systems[measurement_system.attrib['type']] = _text(measurement_system)
```
jtwang: ::squints:: I think you can use _import_type_text(measurement_systems, measurement_system).

babel/core.py
akx: What would you like to see unit tested?

We know empirically (`grep | sort | uniq`) `left-to-right` and `right-to-left` are the only `character_order`s we'll ever see, so `ltr` and `rtl` will also be the only possible `direction`s. (And this is future-proof: if we ever get `top-to-bottom` or `bottom-to-top` or `upside-down-and-dinky-donky`, those will be handled too.)

babel/core.py
akx: I don't think this is that much of an evil. After all, `ltr` and `rtl` are fairly common terms... However, now that you _do_ mention it, this should really be called `text_direction`. I shall rename it post-haste.

akx:
Assigned @jtwang for having done initial review on this.

akx:
Disregard the Codecov fail, btw. I think it's being confused again...

babel/core.py
jtwang: Yikes, good catch. :) What do you think about transforming "left-to-right" to "ltr" (yah, I'm still trying to get rid of one of the two methods :)

jtwang:
@akx - OK to merge! Just the nit about two methods, ignore if you want. ;)

babel/core.py
akx: Well, `character_order` (well, `characterOrder`, but we're snakes, not camels) is the CLDR name, so I'd like to keep that as-is. Likewise, `direction` is the CSS property (though I think it's better spelled as `text_direction` in Babel, just to be explicit).

akx:
Nits acknowledged.

akx:
Gah, this PR shows up in my list of open PRs even if it's merged.

PR #369, opened 2016-03-19
Add unit and compound unit formatting

akx:


scripts/import_cldr.py
```diff
@@ -755,14 +755,24 @@ def parse_currency_names(data, tree):
...
-                unit_patterns.setdefault(box, {})[pattern.attrib['count']] = text_type(pattern.text)
+                pattern_box[pattern.attrib['count']] = text_type(pattern.text)
+
+            per_unit_pat = unit.find('perUnitPattern')
+            if per_unit_pat is not None:
+                pattern_box['per'] = text_type(per_unit_pat.text)
+
+        for unit in elem.findall('compoundUnit'):
+            unit_type = unit.attrib['type']
+            box = "%s:%s" % (unit_type, unit_length_type)
```
jtwang: I wonder why the original implementation uses string concatenation instead of nested dicts (eg date formats)

scripts/import_cldr.py
```diff
@@ -755,14 +755,24 @@ def parse_currency_names(data, tree):
 
 def parse_unit_patterns(data, tree):
     unit_patterns = data.setdefault('unit_patterns', {})
+    compound_patterns = data.setdefault('compound_unit_patterns', {})
     for elem in tree.findall('.//units/unitLength'):
         unit_length_type = elem.attrib['type']
         for unit in elem.findall('unit'):
             unit_type = unit.attrib['type']
+            box = "%s:%s" % (unit_type, unit_length_type)
```
jtwang: For readability, I'd rename:
- box -> unit_and_length and
- pattern_box -> unit_and_length_patterns

Don't care much, but please do at the very least rename
- pattern_box -> box_patterns

babel/numbers.py
```diff
@@ -385,6 +384,205 @@ def format_scientific(number, format=None, locale=LC_NUMERIC):
...
+    Unknown values will return None.
+
+    >>> _find_unit_pattern("horse", locale="en")
+
+    :param unit_id: the code of a measurement unit.
+    :param length: The length for the formatting patterns (long, short or narrow)
+    :return: A key to the `unit_patterns` mapping, or None.
+    """
+    locale = Locale.parse(locale)
+    pattern = "%s:%s" % (unit_id.split(":", 1)[0], length)
```
jtwang: Maybe call this unit_length_key? Don't really care all that much.

babel/numbers.py
```diff
@@ -385,6 +384,205 @@ def format_scientific(number, format=None, locale=LC_NUMERIC):
...
+    if pattern in locale.unit_patterns:
+        return pattern
+    for unit_pattern in sorted(locale.unit_patterns, key=len):
+        if unit_pattern.endswith(pattern):
```
jtwang: Hmm.. What's an example that you would expect to satisfy this condition?

I would have expected a fallback condition along the lines of: `unit_pattern.startswith(unit_id + ':')`

babel/numbers.py
```diff
@@ -385,6 +384,205 @@ def format_scientific(number, format=None, locale=LC_NUMERIC):
     return pattern.apply(number, locale)
 
 
+def _find_unit_pattern(unit_id, length='long', locale=LC_NUMERIC):
```
jtwang: Needs more fallback and length testing please. <3

babel/numbers.py
```diff
@@ -385,6 +384,205 @@ def format_scientific(number, format=None, locale=LC_NUMERIC):
...
+    locale = Locale.parse(locale)
+    pattern = "%s:%s" % (unit_id.split(":", 1)[0], length)
+    if pattern in locale.unit_patterns:
+        return pattern
+    for unit_pattern in sorted(locale.unit_patterns, key=len):
+        if unit_pattern.endswith(pattern):
+            return unit_pattern
+
+
+def format_unit(value, measurement_unit, length='long', format=None, locale=LC_NUMERIC):
```
jtwang: Test all kwargs, blah blah blah. ;)

babel/numbers.py
```diff
@@ -385,6 +384,205 @@ def format_scientific(number, format=None, locale=LC_NUMERIC):
...
+
+    unit_length = _find_unit_pattern(measurement_unit, length, locale)
+    if not unit_length:
+        raise ValueError("%s is not a known unit with length %s in %s" % (measurement_unit, length, locale))
+    unit_patterns = locale.unit_patterns[unit_length]
+    formatted_value = format_decimal(value, format, locale)
+    plural_form = locale.plural_form(value)
+    if plural_form in unit_patterns:
+        return unit_patterns[plural_form].format(formatted_value)
+    # Fall back to a bad representation.
```
jtwang: Would it be better to fall back to *any* plural form? eg. if 'many' doesn't exist, fall back to 'one' or 'zero' or 'few' or whatever is there.

If not, I think `measurement_unit` is pretty user unreadable. I'd prefer to fall back to `displayName` (which babel doesn't support) or show no unit at all.

babel/numbers.py
```diff
@@ -385,6 +384,205 @@ def format_scientific(number, format=None, locale=LC_NUMERIC):
...
+    :param length: The length for the resultant formatting.
+    :param locale: the `Locale` object or locale identifier
+    :return: A key to the `unit_patterns` mapping, or None.
+    """
+    locale = Locale.parse(locale)
+    numerator_unit_length = _find_unit_pattern(numerator_unit, length, locale)
+    denominator_unit_length = _find_unit_pattern(denominator_unit, length, locale)
+    if not (numerator_unit_length and denominator_unit_length):
+        return None
+    # Since compound units are named "speed-kilometer-per-hour:long", we'll have to slice off
```
jtwang: Another case for switching [unit:length][plural] -> [unit][length][plural] :)

akx: Sure. That'll have to wait until 3.0 though, since someone might be using the raw data API.

scripts/import_cldr.py
akx: Commit 5204709 introduced the colons in the first place, but for alternative forms.
Then 9327e08 reused colons/string concat for lengths (misguidedly, perhaps).

I guess "for no particular reason" is the answer :)

babel/numbers.py
akx: As of CLDR 28, there's actually no situation in which we _could_ reach the fallback, it seems!

For all locales, every plural form/unit pattern combination is currently covered.

babel/units.py
```diff
@@ -0,0 +1,241 @@
...
+    """
+    locale = Locale.parse(locale)
+
+    q_unit = _find_unit_pattern(measurement_unit, locale=locale)
+    if not q_unit:
+        raise UnknownUnitError(unit=measurement_unit, locale=locale)
+    unit_patterns = locale._data["unit_patterns"][q_unit].get(length, {})
+    formatted_value = format_decimal(value, format, locale)
+    plural_form = locale.plural_form(value)
+    if plural_form in unit_patterns:
```
jtwang: CLDR here casually mentions length fallbacks (at least that's how I interpreted it here). It'd be awesome to see it here, I don't think it would be that difficult to add an extra iteration through a hardcoded, ordered list of length strings. If it's gross, feel free to punt.

babel/units.py
```diff
@@ -0,0 +1,241 @@
...
+
+    >>> _find_compound_unit("mile", "gallon", locale="en")
+    'consumption-mile-per-gallon'
+
+    :param numerator_unit: The numerator unit's identifier
+    :param denominator_unit: The denominator unit's identifier
+    :param locale: the `Locale` object or locale identifier
+    :return: A key to the `unit_patterns` mapping, or None.
+    """
+    locale = Locale.parse(locale)
```
jtwang: Did you intend to call this "numerator_form" or "numerator_unit_pattern" instead?

babel/units.py
```diff
@@ -0,0 +1,241 @@
...
+    numerator_unit_length = _find_unit_pattern(numerator_unit, locale=locale)
+    denominator_unit_length = _find_unit_pattern(denominator_unit, locale=locale)
+    if not (numerator_unit_length and denominator_unit_length):
+        return None
```
jtwang: Nit: Maybe mention that the numerator and denominator values we'd have in this case are "length-kilometer" and "duration-hour"

babel/units.py
```diff
@@ -0,0 +1,241 @@
...
+def format_compound_unit(
+    numerator_value, numerator_unit=None,
+    denominator_value=1, denominator_unit=None,
+    length='long', format=None, locale=LC_NUMERIC
+):
+    """
+    Format a compound number value, i.e. "kilometers per hour" or similar.
+
+    Both unit specifiers are optional to allow for formatting of arbitrary values still according
+    to the locale's general "per" formatting specifier.
```
jtwang: Please refer to http://www.unicode.org/reports/tr35/tr35-general.html#perUnitPatterns here

babel/units.py
```diff
@@ -0,0 +1,241 @@
...
+            per_pattern = locale._data["unit_patterns"].get(denominator_unit, {}).get(length, {}).get("per")
+            if per_pattern:
+                return per_pattern.format(formatted_numerator)
+        formatted_denominator = format_unit(
+            denominator_value, denominator_unit, length=length, format=format, locale=locale
+        )
+    else:  # Bare denominator
+        formatted_denominator = format_decimal(denominator_value, format=format, locale=locale)
+
+    per_pattern = locale._data["compound_unit_patterns"].get("per", {}).get(length, "{0}/{1}")
```
jtwang: Same length fallback comment as for format_unit() above.

jtwang:
Ow, my head. Yours must hurt more. ;)

jtwang:
Looks good, push!

akx:
Ignoring the single Travis failure; it's one of those split-second unluckinesses.

Issue #372, opened 2016-04-05
Wrong plural forms for Czech and Slovak localizations


ykshatroff:
The plural forms returned by `babel.messages.plurals.get_plural` for Czech `cs` and Slovak `sk` locales are wrong.   
Looking at the list of plural forms at e.g. https://developer.mozilla.org/en-US/docs/Mozilla/Localization/Localization_and_Plurals, we can see that Czech and Slovak languages use Plural rule #*8*, but `get_plural`, and hence `babel.messages.Catalog.plural_forms` property give us the same form as Russian which is Plural rule #*7*. At the same time, the `babel.core.Locale.plural_form` return a correct PluralRule.
 
**Steps to reproduce:** 
Import the `get_plural` function and check its output for Czech locale:

    >>> from babel.core import Locale
    >>> from babel.messages.plurals import get_plural
    >>> cs_locale = Locale.parse('cs')
    >>> cs_locale.plural_form  # this is correct
    <PluralRule u'one: i in 1 and v in 0, few: i in 2..4 and v in 0, many: v not in 0'>

**Actual results:**

    >>> get_plural(cs_locale)
    (3, '(n%10==1 && n%100!=11 ? 0 : n%10>=2 && n%10<=4 && (n%100<10 || n%100>=20) ? 1 : 2)')

**Expected results:**

    >>> get_plural(cs_locale)
    (3, '(n==1) ? 0 : (n>=2 && n<=4) ? 1 : 2')

The above applies to Slovak language.

**Reproducibility:** always

akx:
Thanks for the bug report and the PR!

We should probably ensure that all of those plural rules in the file are up to date, esp. as there's that damning comment

```python
# XXX: remove this file, duplication with babel.plural
```

in there!

ykshatroff:
Actually I was thinking about replacing this PLURALS dict with a 
function that derives plural forms directly from the corresponding 
babel.plural.PluralRule. But this is a little too much for the start, 
not suitable for a hotfix. Still I'd probably take this job if there's 
nobody more eager.

08.04.2016 11:02, Aarni Koskela writes:
> Thanks for the bug report and the PR!
>
> We should probably ensure that all of those plural rules in the file are
> up to date, esp. as there's that damning comment
>
> # XXX: remove this file, duplication with babel.plural
>
> in there!
>
> —
> You are receiving this because you authored the thread.
> Reply to this email directly or view it on GitHub
[...]

PR #373, opened 2016-04-05
Сorrect plural forms for cs and sk locales

ykshatroff:
This PR changes plural forms for Czech (cs) and Slovak (sk) locales from Plural rule #_7_ to Plural rule #_8_ according to https://developer.mozilla.org/en-US/docs/Mozilla/Localization/Localization_and_Plurals .

Fixes https://github.com/python-babel/babel/issues/372


akx:
Thanks for the Mozilla ref! I think we'll have to take a look at that list of rules at some point; there's bound to be something else that's not quite correct still.

Issue #378, opened 2016-04-07
KeyError raised from format_time() with zh_TW on Babel 2.3.1


jun66j5:
(Originally, reported in https://trac.edgewall.org/ticket/12445)
This issue doesn't occur on Babel 2.2.0.
```
>>> import babel
>>> babel.__version__
'2.3.1'
>>> from datetime import datetime
>>> from babel.dates import format_time
>>> format_time(datetime(2016, 4, 8, 12, 34, 56), locale='zh_TW')
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/dev/shm/babel231/local/lib/python2.7/site-packages/babel/dates.py", line 787, in format_time
    return parse_pattern(format).apply(time, locale)
  File "/dev/shm/babel231/local/lib/python2.7/site-packages/babel/dates.py", line 1208, in apply
    return self % DateTimeFormat(datetime, locale)
  File "/dev/shm/babel231/local/lib/python2.7/site-packages/babel/dates.py", line 1205, in __mod__
    return self.format % other
  File "/dev/shm/babel231/local/lib/python2.7/site-packages/babel/dates.py", line 1242, in __getitem__
    return self.format_period(char)
  File "/dev/shm/babel231/local/lib/python2.7/site-packages/babel/dates.py", line 1384, in format_period
    return get_period_names(locale=self.locale)[period]
  File "/dev/shm/babel231/local/lib/python2.7/site-packages/babel/localedata.py", line 207, in __getitem__
    orig = val = self._data[key]
KeyError: 'pm'
```
I guess that `format_period()` should pass `context='format'` to `get_period_names()`.
```diff
diff --git a/babel/dates.py b/babel/dates.py
index 3c75b38..64e7b01 100644
--- a/babel/dates.py
+++ b/babel/dates.py
@@ -1381,7 +1381,7 @@ class DateTimeFormat(object):

     def format_period(self, char):
         period = {0: 'am', 1: 'pm'}[int(self.value.hour >= 12)]
-        return get_period_names(locale=self.locale)[period]
+        return get_period_names(context='format', locale=self.locale)[period]

     def format_frac_seconds(self, num):
         """ Return fractional seconds.
```

jtwang:
Thanks for the awesome details! It really helped a lot. :)

Anyhow, this was probably introduced with the day period API change.

[...]

PR #379, opened 2016-04-08
Fix day period formatting (#378)

akx:
See issue #378 for discussion.

We should release 2.3.2 when this fix is in, if no other bugs/regressions are found.

### DIFF
--- a/babel/dates.py
+++ b/babel/dates.py
@@ -1239,6 +1239,7 @@ class DateTimeFormat(object):
         elif char in ('E', 'e', 'c'):
             return self.format_weekday(char, num)
         elif char == 'a':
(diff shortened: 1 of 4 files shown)


akx:
I kinda want the same stuff @codecov-io is smoking...

akx:
Self-merging as this is a regression fix.

The only failing CI is Codecov, which is on some sort of robot drugs... as usual.

jtwang:
Thanks for tackling this! I expected there to be waaaay more changes, it's awesome that there aren't! I am kind of bummed that zh_Hant is still missing am/pm from it's abbreviated day period definitions. :P

PR #383, opened 2016-04-08
fix typos in messages frontend

sebdiem:
The 'no-wrap', 'ignore-obsolete', 'no-fuzzy-matching', 'previous' and
'update-header-comment' were not correctly parsed in the update_catalog
command.

PR #386, opened 2016-04-11
Fix CLI command usage regressions

### LINKED ISSUE #384: Babel 2.3.2 ignores ungettext_lazy
With OpenStack Horizon, I noticed that the extraction of strings suddenly skips a lot of strings.

Just compare file sizes:
-rw-r--r-- 1 aj users 707 Apr 11 08:01 django232.pot
-rw-r--r-- 1 aj users 911 Apr 11 08:00 django2.pot

to reproduce:

1) Unzip the attached zip file
2) Run "pybabel extract  -F babel-django.cfg --add-comments Translators: -o  django232.pot  -k gettext_noop -k gettext_lazy -k ngettext_lazy:1,2 -k ugettext_noop -k ugettext_lazy -k ungettext_lazy:1,2 -k pgettext_lazy:1c,2 -k npgettext_lazy:1c,2,3 test-babel"
3) Compare the above between Babel 2.2.0 and 2.3.2

the zip file contains my locally generated pot files using Babel 2.2.0 (django2.pot) and 2.3.2 (django232.pot).

[babel232-bug.zip](https://github.com/python-babel/babel/files/212360/babel232-bug.zip)

### LINKED ISSUE #385: Strange new syntax for update --no-fuzzy-matching
Hi,

    pybabel update --help

    -N NO_FUZZY_MATCHING, --no-fuzzy-matching=NO_FUZZY_MATCHING
                        do not use fuzzy matching

Is this change of syntax intended? Shouldn't it be just `--no-fuzzy-matching`, without the equals?

This was using version 2.3.2
Tks

akx:
This fixes #384 and includes @sebdiem 's patch from #383 (which fixes #385 too).

### DIFF
--- a/babel/messages/frontend.py
+++ b/babel/messages/frontend.py
@@ -218,7 +218,7 @@ class extract_messages(Command):
          'charset to use in the output file (default "utf-8")'),
         ('keywords=', 'k',
          'space-separated list of keywords to look for in addition to the '
-         'defaults'),
+         'defaults (may be repeated multiple times)'),
         ('no-default-keywords', None,
          'do not include the default keywords'),
         ('mapping-file=', 'F',
@@ -248,12 +248,12 @@ class extract_messages(Command):
          'set project version in output'),
         ('add-comments=', 'c',
          'place comment block with TAG (or those preceding keyword lines) in '
-         'output file. Separate multiple TAGs with commas(,)'),
+         'output file. Separate multiple TAGs with commas(,)'),  # TODO: Support repetition of this argument
         ('strip-comments', None,
          'strip the comment TAGs from the comments.'),
         ('input-paths=', None,
          'files or directories that should be scanned for messages. Separate multiple '
-         'files or directories with commas(,)'),
+         'files or directories with commas(,)'),  # TODO: Support repetition of this argument
         ('input-dirs=', None,  # TODO (3.x): Remove me.
          'alias for input-paths (does allow files as well as directories).'),
     ]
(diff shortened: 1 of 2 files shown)


babel/messages/frontend.py
```diff
@@ -248,12 +248,12 @@ class extract_messages(Command):
          'set project version in output'),
         ('add-comments=', 'c',
          'place comment block with TAG (or those preceding keyword lines) in '
-         'output file. Separate multiple TAGs with commas(,)'),
+         'output file. Separate multiple TAGs with commas(,)'),  # TODO: Support repetition of this argument
```
ENuge: Not strictly necessary, but I'm a fan of linking these to issue numbers, like TODO(#386): or something.

babel/messages/frontend.py
```diff
@@ -295,13 +294,18 @@ def finalize_options(self):
...
+        else:
+            keywords = DEFAULT_KEYWORDS.copy()
+
+        for kwarg in (self.keywords or ()):
+            keywords.update(parse_keywords(kwarg.split()))
+
+        self.keywords = keywords
+
+        if not self.keywords:
+            raise DistutilsOptionError('no keywords specified')
```
ENuge: I think the old error message ('you must specify new keywords if you disable the default ones') was more descriptive, given that's the only time this will happen.

babel/messages/frontend.py
akx: Refs are visible in the Git commit messages too, so not going to bother adding issue id this time :)

tests/messages/test_frontend.py
```diff
@@ -1227,3 +1230,73 @@ def test_parse_keywords():
...
+        "ugettext_lazy",
+        "ungettext_lazy:1,2",
+        "pgettext_lazy:1c,2",
+        "npgettext_lazy:1c,2,3",
+    ]
+
+    if split:  # Generate a command line with multiple -ks
+        kwarg_text = " ".join("-k %s" % kwarg_spec for kwarg_spec in kwarg_specs)
+    else:  # Generate a single space-separated -k
+        kwarg_text = "-k \"%s\"" % " ".join(kwarg_specs)
```
ENuge: I think it's cleaner to use single quotes (') for the outer string to avoid needing to escape the inner double quote ("), but up to you.

akx: Haha, honestly using double quotes is a habit I've learned both from writing C and from the fact that the double quote is easier to write on a Finnish keyboard than the single quote: Shift+2 vs. a separate key right next to Enter :)

But yeah, could've just as well used single quotes there!

PR #389, opened 2016-04-14
Re-harmonize argument parsing between distutils and optparse CLIs

### LINKED ISSUE #388: 2.3.3 distutils extract_messages misparses --keywords
We run in OpenStack to extract only the "_LW" keyword the following:

 python setup.py extract_messages --no-default-keywords \
            --keyword _LW \
            --add-comments Translators: \
            --input-dirs ${modulename} \
            --output-file ${POT}

With babel 2.3.3 the --no-default-keywords has no effect, the files contain far too much content.

### LINKED ISSUE #390: pybabel extract forgot the -s CLI flag in version 2.3
# Overview Description
The -s flag for pybable extract was a shorthand form of --strip-comments. As part of #311 the shorthand form was removed.

# Steps to Reproduce

1. Try to run `pybabel extract -s`

# Actual Results

    $ pybabel extract -s
    Usage: pybabel extract [options] <input-paths>
    
    pybabel: error: no such option: -s

# Expected Results

    $ pybabel extract -s
    Usage: pybabel extract [options] dir1 <dir2> ...
    
    pybabel: error: incorrect number of arguments

# Reproducibility

Introduced with v2.3 with the consolidation of the CLI interfaces.

akx:
I so hope this will be the last we see of this series of bugs...

Fixes #388
Fixes #390
Refs #384
Refs #311

### DIFF
--- a/babel/messages/frontend.py
+++ b/babel/messages/frontend.py
@@ -98,6 +98,12 @@ class Command(_Command):
     # declared in the base class.)
     boolean_options = ()
 
+    #: Option aliases, to retain standalone command compatibility.
+    #: Distutils does not support option aliases, but optparse does.
+    #: This maps the distutils argument name to an iterable of aliases
+    #: that are usable with optparse.
+    option_aliases = {}
+
     #: Log object. To allow replacement in the script command line runner.
     log = distutils_log
 
@@ -311,6 +317,12 @@ class extract_messages(Command):
     ]
     as_args = 'input-paths'
     multiple_value_options = ('add-comments', 'keywords')
+    option_aliases = {
+        'keywords': ('--keyword',),
+        'mapping-file': ('--mapping',),
+        'output-file': ('--output',),
+        'strip-comments': ('--strip-comment-tags',),
+    }
 
     def initialize_options(self):
         self.charset = 'utf-8'
@@ -870,6 +882,7 @@ class CommandLineInterface(object):
             strs = ["--%s" % name]
             if short:
                 strs.append("-%s" % short)
+            strs.extend(cmdclass.option_aliases.get(name, ()))
             if name == as_args:
                 parser.usage += "<%s>" % name
             elif name in cmdclass.boolean_options:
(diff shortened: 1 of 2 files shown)


akx:
Assigning @jtwang just to have some eyes on this.

ENuge:
(Pretend I'm jtwang), shipit! I'm not super familiar with the different ways users can do option parsing, but the logic looks fine and you've got it tested. Hopefully this doesn't introduce any more regressions. :)

akx:
Well, now that #390 appeared, I'm going to add a take a super duper close look at every single parameter that may have been mutated in #311. So re-review after that's done . . .

akx:
Ping @ENuge / @jtwang again.

babel/messages/frontend.py
```diff
@@ -39,6 +39,48 @@
     from configparser import RawConfigParser
 
 
+def listify_value(arg, split=None):
```
jtwang: Will a single arg ever contain quotes/whitespace? Eg "foo \"i am a.zip\" bar"

akx: That sort of usage will never have been (how's that for a temporal form?) supported by Babel... At least not the optparse frontend.

akx:
@ENuge: Mind taking one more look?

babel/messages/frontend.py
```diff
@@ -825,6 +882,7 @@ def _configure_command(self, cmdname, argv):
             strs = ["--%s" % name]
             if short:
                 strs.append("-%s" % short)
+            strs.extend(cmdclass.option_aliases.get(name, ()))
```
ENuge: Checking if this, combined with `strs = ["--%s" % name]`, doesn't cause any problems? If `name` is "output" you'll end up with strs being ['--output', --output']. I don't have a super solid understanding of all the different options uh, options, so for all I know this intentional. If not an issue, shipit! Everything else looks good to me.

akx: Oop, good catch.

The alias mapping should be `'output_file': ('--output',)`...

akx: (and as of ee8abd6, it is.)

akx:
So yeah, with that last comment fixed, merging. Will roll another bugfix release tomorrow.

babel/messages/frontend.py
ENuge: Err, oh. I don't know why I was looking at the old version. Shipit!

PR #392, opened 2016-04-18
jslexer.py: Change jsx_tag regex

karloskar:
A less greedy jsx_tag regex. Can handle both translatable strings as
children or as props to a component.

Also made a test case with the kind of JSX syntax variants we're using.


akx:
Ah, so the new tag regexp only matches tag starts. Yeah, that sounds about fine for our purposes. Thanks!

PR #405, opened 2016-05-17
Use CLDR 29

### LINKED ISSUE #404: Upgrade to CLDR 29
Here we go again...

http://blog.unicode.org/2016/03/cldr-version-29-released.html
http://www.unicode.org/reports/tr35/tr35-43/tr35.html

akx:
Everything... seems to work, at least!

Fixes #404

### DIFF
--- a/babel/dates.py
+++ b/babel/dates.py
@@ -1080,12 +1080,16 @@ def get_period_id(time, tzinfo=None, type=None, locale=LC_TIME):
     locale = Locale.parse(locale)
 
     # The LDML rules state that the rules may not overlap, so iterating in arbitrary
-    # order should be alright.
-    for rule_id, rules in locale.day_period_rules.get(type, {}).items():
+    # order should be alright, though `at` periods should be preferred.
+    rulesets = locale.day_period_rules.get(type, {}).items()
+
+    for rule_id, rules in rulesets:
         for rule in rules:
             if "at" in rule and rule["at"] == seconds_past_midnight:
                 return rule_id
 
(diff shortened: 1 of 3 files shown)


tests/test_day_periods.py
```diff
@@ -9,7 +9,7 @@
     ("de", time(7, 42), "morning1"),  # (from, before)
     ("de", time(3, 11), "night1"),  # (after, before)
     ("fi", time(0), "midnight"),  # (at)
-    ("en_US", time(12), "noon"),  # (at)
+    ("en", time(12), "noon"),  # (at)
```
jtwang: Why did this change? Purely to make the tests consistent?

jtwang:
Thanks for tackling this! I expected there to be waaaay more changes, it's awesome that there aren't! I am kind of bummed that zh_Hant is still missing am/pm from it's abbreviated day period definitions. :P

jtwang:
ack dc449a1c522bda4d84ab3a01ac1dc9ca8e8bf06b, 09acaaa2e6737f2d03f4c8371e81760357d89df5

(although more test coverage would be awesome)

tests/test_day_periods.py
akx: Good point. I'm actually not sure why I changed that...

akx:
@jtwang Rebased and fixed that en_US thing.

jtwang:
ack 34c92cb220a6eeac40f1cb5b60241ddffca6c3ff

Looks good, merge! (What's the worst that can happen ;)

Issue #35, opened 2013-07-27
python-format flag incorrectly forced on all extractors


wichert:
This is a copy of [old ticket 318](http://babel.edgewall.org/ticket/318).

During extraction the results from an extractor plugin are used to create a message which is then added to the catalog. When the Message instance is created it can set the `python-format` flag if the text matches a specific regular expression. That means that a text like this:

```
This is a text about 50% of all users.
```

will be flagged as a python-format string even it does not originate from python code, or from a piece of python code that will never use formatting. This is breaking our translations in a pretty bad way currently.

As far as I can see the extractor is the only thing that should decide the flags for a message. The Message class itself should never try to guess flags or force them upon plugins, especially since extractors have no way to override this behaviour.

Changaco:
> As far as I can see the extractor is the only thing that should decide the flags for a message.

I disagree. For example, how would the python extractor know whether the flag should be `python-format` or `python-brace-format` (#333)? It can only guess, and not any better than the `Message` class can.

`xgettext` has a `--flag` argument that's supposed to handle this ([doc](https://www.gnu.org/software/gettext/manual/html_node/xgettext-Invocation.html)). Calling `xgettext -o - --flag=_:1:python-format file.py`, with `file.py` containing calls like `_("Some message")`, results in all those messages being flagged with `python-format`.

lelit:
Is there a particular reason why the [PYTHON_FORMAT](https://github.com/python-babel/babel/blob/master/babel/messages/catalog.py#L33) regexp accepts whitespaces _after_ the initial `%` sign?
That's the main reason why wichert's example at the start (that is, `This is a text about 50% of all users.`) is being marked with `python-format`...
For now I replaced that with a `FULLWIDTH PERCENT SIGN`, to not trigger an error at compile time, but I wonder if the optional spaces should be removed from the regexp instead.

wichert:
Because Python allows it:

```
Python 2.7.15 (default, Jun 17 2018, 12:46:58) 
[GCC 4.2.1 Compatible Apple LLVM 9.1.0 (clang-902.0.39.2)] on darwin
Type "help", "copyright", "credits" or "license" for more information.
>>> '% d' % 1
' 1'
```

Note the leading whitespace in the result, which is significant here. This is inherited from C, which also allows it:

```C
#include <stdio.h>
#include <stdlib.h>

int main() {
	printf("% d\n", 1);
	return 0;
}
```

produces:

```
wichert@notos /tmp $ make x
[...]

PR #410, opened 2016-05-29
Allow full control on Decimal behaviour

### LINKED ISSUE #90: Bypass number rounding

etanol:
It turns out that the solution to allow the user to change the rounding mode was way simpler than expected at the beginning.

According to [UTS #35][1] sections [3.3][2] and [3.7][3], the user is allowed to change rounding mode.  But the default should be _half even_ mode.  This is already the default Python rounding mode in all decimal implementations, so the solution was to stop indicating the rounding mode on `Decimal` operations.

Some documentation has been added to show how to properly change the rounding mode and other `Context` parameters.

Therefore, this should fix #90 

[1]: http://www.unicode.org/reports/tr35/tr35-numbers.html
[2]: http://www.unicode.org/reports/tr35/tr35-numbers.html#Formatting
[3]: http://www.unicode.org/reports/tr35/tr35-numbers.html#Rounding

### DIFF
--- a/babel/_compat.py
+++ b/babel/_compat.py
@@ -61,16 +61,12 @@ number_types = integer_types + (float,)
 # Use cdecimal when available
 #
 from decimal import (Decimal as _dec,
-                     InvalidOperation as _invop,
-                     ROUND_HALF_EVEN as _RHE)
+                     InvalidOperation as _invop)
 try:
     from cdecimal import (Decimal as _cdec,
-                          InvalidOperation as _cinvop,
-                          ROUND_HALF_EVEN as _CRHE)
+                          InvalidOperation as _cinvop)
     Decimal = _cdec
     InvalidOperation = (_invop, _cinvop)
-    ROUND_HALF_EVEN = _CRHE
 except ImportError:
     Decimal = _dec
     InvalidOperation = _invop
-    ROUND_HALF_EVEN = _RHE
--- a/babel/_compat.py
+++ b/babel/_compat.py
@@ -58,15 +58,13 @@ number_types = integer_types + (float,)
 
 
 #
-# Use cdecimal when available
+# Since Python 3.3, a fast decimal implementation is already included in the
+# standard library.  Otherwise use cdecimal when available
 #
-from decimal import (Decimal as _dec,
-                     InvalidOperation as _invop)
-try:
-    from cdecimal import (Decimal as _cdec,
-                          InvalidOperation as _cinvop)
-    Decimal = _cdec
-    InvalidOperation = (_invop, _cinvop)
-except ImportError:
-    Decimal = _dec
-    InvalidOperation = _invop
+if sys.version_info[:2] >= (3, 3):
+    import decimal
+else:
+    try:
+        import cdecimal as decimal
+    except ImportError:
+        import decimal
--- a/docs/numbers.rst
+++ b/docs/numbers.rst
@@ -86,6 +86,58 @@ the specification. The following table is just a relatively brief overview.
   +----------+-----------------------------------------------------------------+
 
 
+Rounding Modes
+==============
+
+Since Babel makes full use of Python's `Decimal`_ type to perform number
+rounding before formatting, users have the chance to control the rounding mode
+and other configurable parameters through the active `Context`_ instance.
+
+By default, Python rounding mode is ``ROUND_HALF_EVEN`` which complies with
+`UTS #35 section 3.3`_.  Yet, the caller has the opportunity to tweak the
+current context before formatting a number or currency:
+
+.. code-block:: pycon
+
+    >>> from babel.numbers import decimal, format_decimal
+    >>> with decimal.localcontext(decimal.Context(rounding=decimal.ROUND_DOWN)):
+    >>>    txt = format_decimal(123.99, format='#', locale='en_US')
+    >>> txt
+    u'123'
+
+It is also possible to use ``decimal.setcontext`` or directly modifying the
+instance returned by ``decimal.getcontext``.  However, using a context manager
+is always more convenient due to the automatic restoration and the ability to
+nest them.
+
+Whatever mechanism is chosen, always make use of the ``decimal`` module imported
+from ``babel.numbers``.  For efficiency reasons, Babel uses the fastest decimal
+implementation available, such as `cdecimal`_.  These various implementation
+offer an identical API, but their types and instances do **not** interoperate
+with each other.
+
+For example, the previous example can be slightly modified to generate
+unexpected results on Python 2.7, with the `cdecimal`_ module installed:
+
+.. code-block:: pycon
+
+    >>> from decimal import localcontext, Context, ROUND_DOWN
+    >>> from babel.numbers import format_decimal
+    >>> with localcontext(Context(rounding=ROUND_DOWN)):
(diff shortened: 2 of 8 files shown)


babel/_compat.py
```diff
@@ -58,19 +58,13 @@
...
-    Decimal = _cdec
-    InvalidOperation = (_invop, _cinvop)
-    ROUND_HALF_EVEN = _CRHE
-except ImportError:
-    Decimal = _dec
-    InvalidOperation = _invop
-    ROUND_HALF_EVEN = _RHE
+if sys.version_info[:2] >= (3, 3):
+    import decimal
+else:
```
etanol: This seems odd, according to Travis, [`babel/_compat.py`][1] is 100% covered.

[1]: https://codecov.io/gh/python-babel/babel/src/f12e975b5af270b8846e8e01df37e3b3ad2e3b57/babel/_compat.py

akx:
ack 6238f66 8341064 f12e975

akx:
LGTM, but if someone else (@jtwang?) wants to give this an once-over, that'd be grand. :)

docs/numbers.rst
```diff
@@ -86,6 +86,58 @@ the specification. The following table is just a relatively brief overview.
   +----------+-----------------------------------------------------------------+
 
 
+Rounding Modes
```
jtwang: +++++++

Thanks for the great docs!

Could you add a doctest or a note to the docstring for `babel.numbers.format_decimal`? Same for `format_currency` - especially around the interaction with custom currency formats and `currency_digits`. (omg so complicated)

docs/numbers.rst
```diff
@@ -86,6 +86,58 @@ the specification. The following table is just a relatively brief overview.
...
+
+It is also possible to use ``decimal.setcontext`` or directly modifying the
+instance returned by ``decimal.getcontext``.  However, using a context manager
+is always more convenient due to the automatic restoration and the ability to
+nest them.
+
+Whatever mechanism is chosen, always make use of the ``decimal`` module imported
+from ``babel.numbers``.  For efficiency reasons, Babel uses the fastest decimal
+implementation available, such as `cdecimal`_.  These various implementation
+offer an identical API, but their types and instances do **not** interoperate
```
jtwang: Hmm... is this considered a non-backwards compatible change?

jtwang: Ohhh nevermind, we were already using `cdecimal`...

jtwang:
Thank you so much for taking this on!

sublee:
@jtwang No it doesn't.  What I need is the precision setting for `format_number` and the similar functions.  Not rounding option.

Precision to show usually depend on business logic, not specific locale.  The code for determining precision should be static against dynamic locales.  For example:

```python
>>> print(format_number(123456789.123456789, prec=5, locale=random.choice(locales)))
123.456.789,12346
```

For now, we can't separate precision from locales.  Because `format_number` detects precision from a `format` option but the format string comes from locales.

docs/numbers.rst
etanol: Well, in terms of Babel API is fully backwards compatible.  If your code manipulates the current decimal context while also using Babel, you may see slightly different results.

But if you are playing with decimal precision and exponents, aside from the rounding mode, you should have seen this for some time already

etanol:
The rounding performed by babel is just for the purpose of digit truncation before formatting a numeric or monetary amount.  That said, you can always use the `format` argument of the [`babel.numbers.format_decimal`][1] function, to control how many decimals you want to display, independently of the locale used.  For example:

```
>>> from babel.numbers import format_decimal
>>> format_decimal(1234.1, format='#,##0.000', locale='en_US')
u'1,234.100'
>>> format_decimal(1234.1, format='#,##0.00##', locale='en_US')
u'1,234.10'
>>> format_decimal(1234.123456, format='#,##0.00##', locale='en_US')
u'1,234.1235'
```

Also note that you can pass instances of `Decimal` (properly imported, once this PR is merged) to the number formatting functions.  And, if you are serious about decimal number precision, such as in finances, you **should** be using `Decimal` everywhere in your business logic.

In summary, this PR allows users to control the rounding behavior when digit truncation is necessary to comply with the format in use, either locale's default or custom.

[1]: http://babel.pocoo.org/en/latest/api/numbers.html#babel.numbers.format_decimal

sublee:
@etanol Custom `format` options work well with several locales but not whole of the locales.

The `#,##0.00##` format, you presented, is proper for `en_US` or `ru_RU` because they have the same length of comma groups.  But numbers in `hi_IN` has a different comma rule.  Commas should be inserted every 2 digits except the first group.  Only the first group has 3 digits especially.  In other words, the default number format of `hi_IN` is `#,##,##0.###`.

So, if I want to set a precision by the `format` option, I should know the comma rule of the locale I will use:

```python
>>> format_number(123456789.123456789, locale='hi_IN')
'12,34,56,789.123'
>>> # The format is yours.  The precision specified exactly but commas are wrong.
>>> format_number(123456789.123456789, format='#,##0.00##', locale='hi_IN')
'123,456,789.1235'
>>> # All is okay but the format depends on the hi_IN locale.
>>> format_number(123456789.123456789, format='#,##,##0.00##', locale='hi_IN')
'12,34,56,789.1235'
```

As I said, it depends on locales instead of business logic.

etanol:
@sublee Right, I see what you mean.  I have a couple of ideas to refactor the `babel.numbers` module and maybe expose some tuning over the number formatting by skipping over string patterns.  But due to lack of spare time, don't expect any change (at least from my part) until the end of this year.

sublee:
@etanol Sounds great!  I'll wait your work.

akx:
So yeah, even if this doesn't exactly fix #90, I think it's a step in a nice direction.

@jtwang, @sublee, what do you think, shall we merge?

jtwang:
@akx - sure, I didn't spot anything worrisome. :)

Issue #407, opened 2016-05-20
NumberPattern.pattern does not preserve original pattern


jtwang:
It basically keeps the positive format only.

```
>>> babel.numbers.parse_pattern(u'¤#,##0.00;(¤#,##0.00)').pattern
u'\\xa4#,##0.00'
```

Ideally, `pattern` would be `'¤#,##0.00;(¤#,##0.00)'`

This makes it difficult to copy/recreate an existing NumberPattern object.

Also, `repr()` uses `self.pattern`, so `repr(parse_pattern(u'¤#,##0.00;(¤#,##0.00)'))` returns the same thing as `repr(parse_pattern(u'¤#,##0.00'))`, which is misleading.

akx:
Fixed in #414 :)

PR #414, opened 2016-06-02
parse_pattern now preserves original pattern #407

jtwang:
More precisely, `babel.numbers.parse_pattern()` only passes the positive string format to `NumberPattern()`. Fixed the method so that it now passes the entire original pattern string.

### DIFF
--- a/babel/numbers.py
+++ b/babel/numbers.py
@@ -487,13 +487,15 @@ def parse_pattern(pattern):
             raise ValueError('Invalid number pattern %r' % pattern)
         return rv.groups()
 
+    pos_pattern = pattern
+
     # Do we have a negative subpattern?
     if ';' in pattern:
-        pattern, neg_pattern = pattern.split(';', 1)
-        pos_prefix, number, pos_suffix = _match_number(pattern)
+        pos_pattern, neg_pattern = pattern.split(';', 1)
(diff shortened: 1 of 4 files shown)


jtwang:
Whoops, forgot to recreate .dat files prior to running `make test`, hence tests passing locally but failing on  AppVeyor.

babel/numbers.py
```diff
@@ -561,6 +563,18 @@ def __init__(self, pattern, prefix, suffix, grouping,
         else:
             self.scale = 0
 
+    def __copy__(self):
```
akx: Hmm, so is a concrete implementation of `__copy__` here actually even necessary?

I mean, wouldn't the builtin implementation of copy just basically do the same -- create a new instance and shallow-copy the instance properties?

jtwang: Uhhhh... you know, you're right. I can't remember why I was ~sure~ that copying didn't work. :|

jtwang:
Argh, I got bit by the changes in string handling between Python 2 and 3.

akx:
Nice. :ok_hand: 

Don't mind the coverage fluke, it's just codecov not being able to merge results from different platforms again.

jtwang:
Many thanks, @akx !

PR #418, opened 2016-06-16
Prevent update_catalog from stomping over comments in existing po files

Mabusto:
`pybabel update` will ignore existing comments in a po file. This issue is outlined in greater detail here (https://babel.edgewall.org/ticket/170) and this is the suggested fix. Auto comments are not copied over here as they will work when extractor runs, but user comments are stored in the .po file, therefore will be overwritten unless they are merged in.

### DIFF
--- a/babel/messages/catalog.py
+++ b/babel/messages/catalog.py
@@ -753,6 +753,7 @@ class Catalog(object):
             else:
                 oldmsg = remaining.pop(oldkey, None)
             message.string = oldmsg.string
+            message.user_comments = list(distinct(oldmsg.user_comments))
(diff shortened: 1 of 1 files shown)


akx:
Thanks for the patch, @Mabusto! Can you add a test that verifies the behavior?

jsuter:
Can we get this merged in? We're running into the same issue. I'd commit a test if I could, but I have no idea how.

Mabusto:
Hi there,

This is a relatively simply change. I see that it is marked for review. Is there anything I can do to help move this along and get it merged in?

PR #420, opened 2016-06-18
Read locale from .po if it has Language header

### LINKED ISSUE #422: Language tag is not read from .po header
When you read in a .po file that already has the "Language" header specified, writing it out again will erase the Language header. I sent pull request #420 to fix this.

kruton:
When a .po file was written out it would write the "Language"
header, but it wouldn't subsequently read that header back in. This
change makes that symmetric so you can write out a file and get back the
same Catalog by reading it in again.

Additionally other values like charset get overridden when the header is
read, so codify the behavior that the read "Language" header gets set as
the locale of the Catalog object.

Fixes #422

### DIFF
--- a/babel/messages/catalog.py
+++ b/babel/messages/catalog.py
@@ -402,6 +402,8 @@ class Catalog(object):
                 self.msgid_bugs_address = value
             elif name == 'last-translator':
                 self.last_translator = value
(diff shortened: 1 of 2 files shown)


akx:
Seems sensible. Thank you for the patch, @kruton!

PR #424, opened 2016-06-24
Fix extraction order

ngrilly:
Before Babel 2.3, the input directories were always extracted 
in the order specified by the command line arguments.

For example, `pybabel extract dir_a dir_b dir_c` would process 
`dir_a` first, then `dir_b`, and finally `dir_c`.

Since Babel 2.3, the inputs are stored in a Python dictionary, 
and we loop over items in the dictionary using items().
This makes the order unspecified, not matching the order 
specified on the command line anymore.

### DIFF
--- a/babel/messages/frontend.py
+++ b/babel/messages/frontend.py
@@ -416,7 +416,7 @@ class extract_messages(Command):
                               copyright_holder=self.copyright_holder,
                               charset=self.charset)
 
-            for path, (method_map, options_map) in mappings.items():
+            for path, method_map, options_map in mappings:
                 def callback(filename, method, options):
                     if method == 'ignore':
                         return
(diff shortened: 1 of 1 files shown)


buckbaskin:
@ngrilly Have you considered using a [collections.OrderedDict](https://docs.python.org/2/library/collections.html#collections.OrderedDict)? I believe it does what you are looking for (maintain insertion order). See the [first recipe](https://docs.python.org/2/library/collections.html#ordereddict-examples-and-recipes) for an example of what I'm thinking about.

ngrilly:
@buckbaskin Yes, I've considered it. It even was my first thought :-) But I think using an OrderectDict would be overkill in this case, because the only operation we do over the data structure returned by `_get_mappings` is to iterate over it, nothing else.

In the gettext philosophy, the default behavior is to store messages in .po files in their extraction order. This preserves message "locality", and keep related messages together, which facilitates the translator's work. This is why I think fixing this is important.

akx:
Yeah, this looks fine, thanks! (`_get_mappings()` is an underscore-private API anyway; if someone was depending on it returning a dict, they should be prepared to deal with it changing.)

Thank you for the patch, @ngrilly!

PR #425, opened 2016-06-25
Increase coverage of JSX extraction tests

georgschoelly:
b5e0ad8 fixed a few bugs with jsx parsing but only added
tests for the lexer. This commit adds a test for the extraction
which did not work before:

```
<tag>{ _"Text" }</tag>
```

This test case is not present yet in the suite, only tags with
attributes (`<tag attr="value">`) are, but they get extracted correctly
already.


akx:
Hi @georgschoelly -- I'd love to merge this PR, but tests currently don't pass due to a CI configuration issue (not your fault!). Can you either rebase this on top of the current `master` or alternatively check the "Allow maintainers to edit" checkbox on the right? 👉

georgschoelly:
I checked the box, I currently don't have an active project with python-babel. Thanks for your work!

PR #427, opened 2016-07-01
Python message extractor respects unicode_literals in __future__

### LINKED ISSUE #426: Python message extractor doesn't respect unicode_literals in __future__
`extract_python` of Babel doesn't respect `from __future__ import unicode_literals` in Python 2.x.  With the below example code:

```python
# coding: utf-8
from __future__ import unicode_literals
_('\xa0')
```

We get this encoding error:

```
UnicodeDecodeError: 'utf8' codec can't decode byte 0xa0 in position 10: invalid start byte
```

A workround is to use `u''` literals explicitly:

```python
# coding: utf-8
from __future__ import unicode_literals
_(u'\xa0')
```

sublee:
This patch fixes #426.

### DIFF
--- a/babel/messages/extract.py
+++ b/babel/messages/extract.py
@@ -22,7 +22,7 @@ from os.path import relpath
 import sys
 from tokenize import generate_tokens, COMMENT, NAME, OP, STRING
 
-from babel.util import parse_encoding, pathmatch
+from babel.util import parse_encoding, parse_future_flags, pathmatch
 from babel._compat import PY2, text_type
 from textwrap import dedent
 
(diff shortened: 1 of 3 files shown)


akx:
The two commits should be squashed into one, since the second seems to just add Py3.5 compatibility which was lacking in the first commit.

Looks fine to me otherwise. :)

sublee:
@akx Thank you for reviewing my patch.  I squashed my commits as you advised.  Please review again.

akx:
Thanks for the patch, @sublee! :)

sublee:
@akx Thank you for accepting it! :)

Issue #27, opened 2013-07-22
Pofile parser can't handle multi-line obsolete IDs or Strings


nickretallack:
Any multi-line obsolete message just comes out as "".

It should be obvious why, when you look at [this line](https://github.com/mitsuhiko/babel/blob/master/babel/messages/pofile.py#L222).  Every line of an obsolete string begins with a #, but every time the script sees a # it throws away all its state variables assuming it is probably starting a new message now, or accumulating occurrences and comments about one.

Here's a test case:
```
import babel.messages.pofile as pofile
from StringIO import StringIO

buffer = StringIO("""
#~ msgid ""
#~ "ID"
#~ msgstr ""
#~ "String"
""")

catalog = pofile.read_po(buffer)
assert catalog.obsolete.keys() == ['ID']
assert catalog.obsolete.values() == 'String'
```

akx:
Closed via #429 :)

PR #429, opened 2016-07-08
Support reading multi line obsolete units in po files

mbirtwell:
This fixes issue #27. I also found with multi-line obsolete units it would mark the following unit as obsolete. This fixes that also. 

The fixes are entirely confined to the second commit. The first commit refactors read_po so that all the state is stored on a class. This is just because I found it more readable than having to use lists to store a single boolean which we wanted to update from inside the closures.

### DIFF
--- a/babel/messages/pofile.py
+++ b/babel/messages/pofile.py
@@ -73,11 +73,16 @@ def denormalize(string):
         return unescape(string)
 
 
-class _PoFileParser(object):
+class PoFileParser(object):
+    """Support class to  read messages from a ``gettext`` PO (portable object) file
+    and add them to a `Catalog`
 
-    def __init__(self, locale=None, domain=None, ignore_obsolete=False, charset=None):
+    See `read_po` for simple cases.
+    """
+
+    def __init__(self, catalog, ignore_obsolete=False):
+        self.catalog = catalog
         self.ignore_obsolete = ignore_obsolete
-        self.catalog = Catalog(locale=locale, domain=domain, charset=charset)
         self.counter = 0
         self.offset = 0
         self.messages = []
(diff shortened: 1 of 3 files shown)


sils1297:
ack https://github.com/python-babel/babel/pull/429/commits/651f9bd6ecb46d27b381ecbfdae1a99d14fa48f0

sils1297:
@akx can you review the other two commits?

babel/messages/pofile.py
```diff
@@ -73,6 +73,151 @@ def denormalize(string):
...
+        self.flags = []
+        self.user_comments = []
+        self.auto_comments = []
+        self.obsolete = False
+        self.context = []
+        self.in_msgid = False
+        self.in_msgstr = False
+        self.in_msgctxt = False
+
+    def _add_message(self):
```
akx: I'd appreciate a docstring for this function. Even if it is underscore-private, for future developers it'd be nice to understand what this does -- as I understand, it saves (for lack of a better word) the current parser state as a message in the catalog, then resets the parser for a following message? Or am I mistaken?

babel/messages/pofile.py
```diff
@@ -73,6 +73,151 @@ def denormalize(string):
         return unescape(string)
 
 
+class _PoFileParser(object):
+
+    def __init__(self, locale=None, domain=None, ignore_obsolete=False, charset=None):
+        self.ignore_obsolete = ignore_obsolete
+        self.catalog = Catalog(locale=locale, domain=domain, charset=charset)
```
akx: As this is actually just a parser (that adds messages from a po-file style stream into a catalog), it feels like it should be the caller's responsibility to pass in a `catalog` object.

That way this parser could be used as a lower-level API to "append" messages into a premade catalog object.

akx:
As for the refactoring into a class in general, applaudable work! :)

akx:
Thanks @mbirtwell 👌

Issue #430, opened 2016-07-08
Task: revise hard-coded plurals mapping


akx:
The plurals mapping `babel/messages/plurals.py` could use an once-over, just to see we're up to date and up to spec.

https://developer.mozilla.org/en-US/docs/Mozilla/Localization/Localization_and_Plurals could be a decent reference.

Refs #373

PR #431, opened 2016-07-11
Updated plurals.py to match Mozilla and CLDR references

mshenfield:
This commit updates the plural rules, and in some cases the plural
number, for languages based on review of
https://developer.mozilla.org/en-US/docs/Mozilla/Localization/Localization_and_Plurals,
and
http://www.unicode.org/cldr/charts/29/supplemental/language_plural_rules.html.

## Changed
I changed the interface (either the number of plurals, or the meaning
of each index), for the following languages:
• Arabic (ar): numplurals same, semantics differ per Mozilla's "Plural rule #12". 3 has been made more accurate, 4 has been switched with 5, and the meaning of both has been refined to match the rule.
• Belarusian (ar): numplurals updated to 3, using Mozilla's "Plural
rule #7". Previously the default, where anything except one is plural.
• Breton (br): numplurals updated to 6, using Mozilla's "Plural rule
#16". Previosly the default, where anything except one is plural.
• Bosnian (bs): numplurals updated to 3, using Mozilla's "Plural rule
#7". Previously the default, where anything except one is plural.
• Irish Gaelic (ga): numplurals update from 3 to 5, using Mozilla's
"Plural rule #11". Cases have been added for n between 3 and 6, and n
between 7 and 10.
• Icelandic (is): same numplurals (2), semantics updated using
Mozilla's "Plural rule #15". 0 now means that n ends in 1, and is not
11, instead of just the number 1.
• Maltese (mt): same numplurals, same rule. Fixed rule, index 1 now
includes numbers ending in 1 (except the number 1)

## Ignoring Discrepancy
I'm ignoring these two rules listed on the Mozilla resource
• Hungarian - listed as `(n != 1)`. It is not required to pluralize a
noun if a qualitative or quantitative amount precedes it
http://www.hungarianreference.com/plurals.aspx.
• Vietnamese - listed as `(n != 1)`. It only seems to have one plural
according to this lesson:
http://yourvietnamese.com/vietnamese-grammar/nouns-in-vietnamese/.

## Not Included
There are some additional languages listed in the Mozilla reference
that are not in the plurals document, and could be added:
• Lower Sorbian (dsb) (4, '(n%100==1 ? 0 : n%100==2 ? 1 : n%100==3 ||
n%100==4 ? 2 : 3)')
• Upper Sorbian (hsb) (4, '(n%100==1 ? 0 : n%100==2 ? 1 : n%100==3 ||
n%100==4 ? 2 : 3)')
• Scottish Gaelic (gd)  (4, '(n==1 || n==11 ? 0 : n==2 || n==12 ? 1 :
(n>=3 && n<=10) || (n>=13 && n<=19) ? 2 : 3)')
• Faroese (fo) (2, '(n != 1)')
• Frisian (fy) (2, '(n != 1)')
• Macedonian (mk) (3, '(n%10==1 ? 0 : n%10==2 ? 1 : 2)')

Closes https://github.com/python-babel/babel/issues/430

### DIFF
--- a/babel/messages/catalog.py
+++ b/babel/messages/catalog.py
@@ -479,7 +479,7 @@ class Catalog(object):
         >>> Catalog(locale='en').num_plurals
         2
         >>> Catalog(locale='ga').num_plurals
-        3
+        5
 
         :type: `int`"""
         if self._num_plurals is None:
@@ -496,7 +496,7 @@ class Catalog(object):
         >>> Catalog(locale='en').plural_expr
         '(n != 1)'
         >>> Catalog(locale='ga').plural_expr
-        '(n==1 ? 0 : n==2 ? 1 : 2)'
+        '(n==1 ? 0 : n==2 ? 1 : n>=3 && n<=6 ? 2 : n>=7 && n<=10 ? 3 : 4)'
 
         :type: `string_types`"""
         if self._plural_expr is None:
--- a/babel/messages/plurals.py
+++ b/babel/messages/plurals.py
@@ -35,7 +35,7 @@ PLURALS = {
     # Aragonese
     # 'an': (),
     # Arabic - From Pootle's PO's
-    'ar': (6, '(n==0 ? 0 : n==1 ? 1 : n==2 ? 2 : n>=3 && n<=10 ? 3 : n>=11 && n<=99 ? 4 : 5)'),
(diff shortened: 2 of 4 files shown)


akx:
Hey, brilliant, thanks @mshenfield! 👍 

Never mind the @gitmate-bot comments for this sort of stuff -- it's "autogenerated" code, so it's fine to be over the usual length limit. (@sils1297 , how does one disable that warning for a file?)

akx:
Thanks @mshenfield 😸

PR #437, opened 2016-07-27
Miscellaneous small improvements

scop:


akx:
Disregard Codecov's complaints -- it's trying its best, but can be a bit silly at times.

Anyway, these are nice little changes indeed! Thanks. Merging! :)

Oh, and if you have the time and inclination, there are open PRs that could use more pairs of eyes for code review :)

PR #438, opened 2016-08-09
Add option 'add_location' for location line formatting

rrader:
Support of add_location option added, can be either 'full', 'file'
or 'never'. 'full' includes both file and line number; 'file' includes
only the file name without line number; 'never' doesn't include the
location comment at all (same as --no-location).

Python babel implementation lacks of this option. We would like to use it to avoid merge conflicts while merging po files.

Option `add_location` is implemented in the GNU xgettext, quote from the manual:

> ‘--add-location=type’
> 
> ```
> Generate ‘#: filename:line’ lines (default).
> 
> The optional type can be either ‘full’, ‘file’, or ‘never’. If it is not given or ‘full’, it generates the lines with both file name and line number. If it is ‘file’, the line number part is omitted. If it is ‘never’, it completely suppresses the lines (same as --no-location).
> ```

https://www.gnu.org/software/gettext/manual/html_node/xgettext-Invocation.html

### DIFF
--- a/babel/messages/frontend.py
+++ b/babel/messages/frontend.py
@@ -104,6 +104,10 @@ class Command(_Command):
     #: that are usable with optparse.
     option_aliases = {}
 
+    #: Choices for options that needed to be restricted to specific
+    #: list of choices.
+    option_choices = {}
+
     #: Log object. To allow replacement in the script command line runner.
     log = distutils_log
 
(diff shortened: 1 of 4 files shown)


babel/messages/frontend.py
```diff
@@ -317,13 +326,17 @@ class extract_messages(Command):
...
+        'add-location': ('full', 'file', 'never',),
+    }
 
     def initialize_options(self):
         self.charset = 'utf-8'
         self.keywords = None
         self.no_default_keywords = False
         self.mapping_file = None
         self.no_location = False
+        self.add_location = None
```
akx: This basically defaults to `'full'` though, doesn't it?

babel/messages/frontend.py
```diff
@@ -401,6 +415,11 @@ def finalize_options(self):
             if not self.version:
                 self.version = self.distribution.get_version()
 
+        if self.add_location == 'never':
+            self.no_location = True
+        elif self.add_location == 'file':
+            self.include_lineno = False
+
```
akx: It might be worth it to `del self.add_location` here, as the variable has already served its purpose and further modifications to it will not have any effect.

akx:
Thanks for the patch, @rrader! :) Much appreciated.

alxpy:
@akx, hello! When will be the new release with this patch?)

Issue #402, opened 2016-05-05
Add `--ignore-dirs` or similar to `extract`


akx:
See #53, #253, #401, #124.

Implementation idea:
- Add `directory_filter` callable arg to `extract_from_dir()`; this would be used instead of the hard-coded `if subdir.startswith('.') or subdir.startswith('_'):`
- Generate a directory filter callable from an `--ignore-dirs` argument, if passed in the CLI

rlam3:
Looking forward to this patch to ignore certain directories/subdirectories.

PR #447, opened 2016-09-22
adding directory_filter arg to customized ignored dirs

jmqd:
Per #402, adding directory_filter arg and replacing hard-coded .startswith() calls. I am unsure where to add --ignore-dirs argument.

This is my first time ever doing a pull request, so please forgive any rookie mistakes. :)

### DIFF
--- a/babel/messages/extract.py
+++ b/babel/messages/extract.py
@@ -63,7 +63,8 @@ def _strip_comment_tags(comments, tags):
 
 def extract_from_dir(dirname=None, method_map=DEFAULT_MAPPING,
                      options_map=None, keywords=DEFAULT_KEYWORDS,
-                     comment_tags=(), callback=None, strip_comment_tags=False):
+                     comment_tags=(), callback=None, strip_comment_tags=False,
+                     directory_filter=('.','_')):
     """Extract messages from any source files found in the given directory.
 
     This function generates tuples of the form ``(filename, lineno, message,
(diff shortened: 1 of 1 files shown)


babel/messages/extract.py
```diff
@@ -128,6 +129,9 @@ def extract_from_dir(dirname=None, method_map=DEFAULT_MAPPING,
                      positional arguments, in that order
     :param strip_comment_tags: a flag that if set to `True` causes all comment
                                tags to be removed from the collected comments.
+    :param directory_filter: a list of strings that identify the starting
```
akx: To be honest, this might be more useful if it declared a list of [`fnmatch`](https://docs.python.org/3.5/library/fnmatch.html#module-fnmatch) compatible patterns.

Also, now that I think of it, it could maybe just accept a callable? After all there's no command-line support for this argument yet, so it needn't be a "command line compatible". The default could be `default_directory_filter`,  defined as 

```python
def default_directory_filter(path):
  subdir = os.path.basename(path)
  return not (subdir.startswith('.') or subdir.startswith('_'))
```

and the implementation in the walk loop simply
```python
dirnames[:] = [dirname for dirname in dirnames if directory_filter(os.path.join(root, dirname)]
```

akx:
Also: no rookie mistakes made! :)

babel/messages/extract.py
```diff
@@ -138,7 +142,7 @@ def extract_from_dir(dirname=None, method_map=DEFAULT_MAPPING,
     absname = os.path.abspath(dirname)
     for root, dirnames, filenames in os.walk(absname):
         for subdir in dirnames:
-            if subdir.startswith('.') or subdir.startswith('_'):
+            if any(subdir.startswith(chars) for chars in directory_filter):
```
mvnm: starstwith can also accept a tuple of strings to try .. so `if subdir.startswith(directory_filter)` should suffice

Also while at it, `extract_messages` command should be modified to accept the `directory_filter` argument in order to be passed to `extract_from_dir`

lelit:
Is this going to be accepted? If so, what needs to be done?

Inside one of my apps there is a huge NPM's `node_modules` hierarchy, and the extract command is needlessly traversing it, so this feature would be nice to have...

Issue #442, opened 2016-08-25
Parsing time and date must be enhanced


renzon:
# Overview Description

parse_time can not handle missing day times or 12 hour clock as well. 

Both parse_time and parse_date raise IndexError when date/time is not well format, making error handling hard
# Steps to Reproduce
1. All of above time should be parsed:

``` python
    assert dates.parse_time('15:30', locale='en_US') == time(15, 30)
    assert dates.parse_time('3:30', locale='en_US') == time(3, 30)
    assert dates.parse_time('00:30', locale='en_US') == time(0, 30)
    assert dates.parse_time('03:30 PM', locale='en_US') == time(15, 30)
    assert dates.parse_time('03:30 pM', locale='en_US') == time(15, 30)
    assert dates.parse_time('03:30 pm', locale='en_US') == time(15, 30)
    assert dates.parse_time('03:30 Pm', locale='en_US') == time(15, 30)
    assert dates.parse_time('03:30:21 AM', locale='en_US') == time(3, 30, 21)
    assert dates.parse_time('03:30:00 AM', locale='en_US') == time(3, 30)
    assert dates.parse_time('03:30:00 AM', locale='en_US') == time(3, 30)
```
1. Try parsing invalid time: ('', 'a', 'aaa', '15', '15:')
2. Try parsing invalid date: ('', 'a', 'aaa', '15', '15/15/')
# Actual Results
1. 

``` python
>>> from babel import dates
>>> dates.parse_time('15:30', locale='en_US')
Traceback (most recent call last):
  File "<input>", line 1, in <module>
  File "/Users/renzo/PycharmProjects/babel/babel/dates.py", line 1186, in parse_time
    second = int(numbers[indexes['S']])
IndexError: list index out of range
>>> dates.parse_time('03:30:21 AM', locale='en_US')
datetime.time(3, 30, 21)
>>> dates.parse_time('03:30 Pm', locale='en_US')
Traceback (most recent call last):
  File "<input>", line 1, in <module>
  File "/Users/renzo/PycharmProjects/babel/babel/dates.py", line 1186, in parse_time
    second = int(numbers[indexes['S']])
IndexError: list index out of range
>>> dates.parse_time('')
Traceback (most recent call last):
  File "<input>", line 1, in <module>
  File "/Users/renzo/PycharmProjects/babel/babel/dates.py", line 1184, in parse_time
    hour = int(numbers[indexes['H']])
IndexError: list index out of range
>>> dates.parse_time('a')
[...]

PR #448, opened 2016-09-28
Fix for Issue #442

DBauer15:
Parsing time and date must be enhanced #442 

rezon wrote, that he wanted a way to handle errors more precisely, so I implemented two new Exceptions dedicated to these issues.
I further implemented a more tolerant way of parsing time and date strings, allowing the user to leave certain fields blank (like TIME without the 'seconds' part).

To test all these new things, there is a new unit test in the tests folder.

### DIFF
--- a/babel/Exceptions/ParseDateException.py
+++ b/babel/Exceptions/ParseDateException.py
@@ -1,3 +1,0 @@
-class ParseDateException(ValueError):
-    def __init__(self, message):
-        self.message = message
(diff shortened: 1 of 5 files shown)


afalco103:
while this is unmerged, I copied the solution into my project. There's an issue with it the 12hr offset. It specifically looks for 'pm' to determine it, but thats not locale friendly. Other languages have different ways of writing 'pm'

PR #449, opened 2016-10-05
Fix spelling word into Raise TypeError

roramirez:


akx:
Thanks! Every minor patch counts. :)

PR #452, opened 2016-10-12
Fix parsing pofile with obsoletes

mbirtwell:
Fixes the handling of the unit before the obsolete unit. Previously it would mark the unit before an obsolete unit as obsolete also.

Fixes multi-line msgctxt statements. Previously it would just take the first line.

Some refactoring:
- the transition between finishing one unit and starting the nextis clearer
- separate the processing of keywords and continuation lines
- combine the reset and initialisation code
- Make the handling of strings consistent.
- Add some nascent error handling, removed some errors in test inputs

### DIFF
--- a/tests/messages/test_pofile.py
+++ b/tests/messages/test_pofile.py
@@ -79,7 +79,7 @@ msgstr ""''')
                          message.id)
 
     def test_fuzzy_header(self):
-        buf = StringIO(r'''\
+        buf = StringIO(r'''
 # Translations template for AReallyReallyLongNameForAProject.
 # Copyright (C) 2007 ORGANIZATION
 # This file is distributed under the same license as the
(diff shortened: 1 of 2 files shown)


tests/messages/test_pofile.py
```diff
@@ -79,7 +79,7 @@ def test_read_multiline(self):
                          message.id)
 
     def test_fuzzy_header(self):
-        buf = StringIO(r'''\
+        buf = StringIO(r'''
```
akx: Out of curiosity -- is the parser still able to read files that start with a newline, i.e. is this just cleanup? :)

akx:
Could you fold the Python 2.6 changes into the relevant earlier commits?

Otherwise this looks nice :)

tests/messages/test_pofile.py
akx: Fair enough! Thank you for the explanation.

PR #459, opened 2016-12-15
Fix setup.py option 'add_location'

alxpy:
Fix bug in setup.py options.

### DIFF
--- a/babel/messages/frontend.py
+++ b/babel/messages/frontend.py
@@ -277,7 +277,7 @@ class extract_messages(Command):
          'path to the mapping configuration file'),
         ('no-location', None,
          'do not include location comments with filename and line number'),
-        ('add-location', None,
+        ('add-location=', None,
          'location lines format. If it is not given or "full", it generates '
          'the lines with both file name and line number. If it is "file", '
          'the line number part is omitted. If it is "never", it completely '


babel/messages/frontend.py
```diff
@@ -277,7 +277,7 @@ class extract_messages(Command):
          'path to the mapping configuration file'),
         ('no-location', None,
```
sils: why not add it for this and the others?

alxpy: because there are different types of options: `no-location` is [boolean option](https://github.com/python-babel/babel/blob/master/babel/messages/frontend.py#L885), `add-location` is [multiple value option](https://github.com/python-babel/babel/blob/master/babel/messages/frontend.py#L887)

example which does't work without this fix:
```
python setup.py extract_messages \
--mapping-file international/babel_mapping.ini \
--output-file proj/i18n/file.pot \
--charset utf-8 --add-location file \
--omit-header --sort-output
```
`omit-header` and `sort-output` -- boolean options, `output-file` and `add-location` -- multiple value options

alxpy: @sils so is it OK for you?

sils: sorry, yes. Too much going on :/

alxpy: @sils can you accept PR?

sils:
ack https://github.com/python-babel/babel/pull/459/commits/9be702b334453d0a22c18e7443a172cc5ee9f7d3

PR #460, opened 2016-12-17
Add information about precision on format_percent

viniciusfeitosa:
I've written more information about precision  on format_percent.


akx:
This is conflicted by #494, so closing.

PR #462, opened 2016-12-29
Pull flags out of regular expressions

singingwolfboy:
In Python, you can pass flags to regular expressions. There are two ways to do so: by passing the flag as a `flag` argument to whatever regex function you're using, or by embedding the flag in the regex itself. If the flag is embedded in the regex, it is ambiguous unless it is at the start of the expression, and in Python 3.6 this is now deprecated.

This pull request pulls all flags out of regexes and passes them via the `flag` argument, instead. This is clearer and more consistent.

### DIFF
--- a/babel/localtime/_unix.py
+++ b/babel/localtime/_unix.py
@@ -5,7 +5,7 @@ import sys
 import pytz
 import subprocess
 
-_systemconfig_tz = re.compile(r'^Time Zone: (.*)$(?m)')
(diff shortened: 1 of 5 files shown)


akx:
Looks good, thanks @singingwolfboy!

Can you `git commit --amend --reset-author` and force-push this branch to force appveyor to re-check the PR? It seems to have had an intermittent failure of some sort I can't work around.

akx:
(Also, @sils, can you make the Appveyor build non-required for merge? It seems to be flaky.)

sils:
(Everything for you @akx :))

PR #465, opened 2017-01-10
Make `windows_zone_mapping`lossless. Fixes #464

### LINKED ISSUE #464: get_global('windows_zone_mapping') is lossy
According to http://unicode.org/repos/cldr/trunk/common/supplemental/windowsZones.xml, there is a one-to-many relationship from Windows timezone to 'standard' timezone name. But the importer (see https://github.com/python-babel/babel/blob/5c02138363a0c0e1fbbfeccf2c6f5f731b6dee92/scripts/import_cldr.py#L234) blindly selects only the first timezone name for the map value:

```python
>>> get_global('windows_zone_mapping')['Romance Standard Time']
'Europe/Paris'
```

I think `get_global('windows_zone_mapping')` should either return tuples with all the related timezone names:

```python
>>> get_global('windows_zone_mapping')['Romance Standard Time']
('Europe/Paris', 'Europe/Brussels', 'Europe/Copenhagen', 'Europe/Madrid Africa/Ceuta')
```

or be turned around so it maps timezone name -> Windows timezone:

```python
>>> get_global('windows_zone_mapping')['Europe/Copenhagen']
'Romance Standard Time'
```

ecederstrand:
Changes  `get_global('windows_zone_mapping')` to return tuple values. This preserves all input data and makes it possible to easily make a reverse map without losing keys.

### DIFF
--- a/scripts/import_cldr.py
+++ b/scripts/import_cldr.py
@@ -231,7 +231,7 @@ def parse_global(srcdir, sup):
     _zone_territory_map = {}
     for map_zone in sup_windows_zones.findall('.//windowsZones/mapTimezones/mapZone'):
         if map_zone.attrib.get('territory') == '001':
-            win_mapping[map_zone.attrib['other']] = map_zone.attrib['type'].split()[0]
+            win_mapping[map_zone.attrib['other']] = map_zone.attrib['type'].split()
         for tzid in text_type(map_zone.attrib['type']).split():
             _zone_territory_map[tzid] = text_type(map_zone.attrib['territory'])
     for key_elem in bcp47_timezone.findall('.//keyword/key'):


akx:
Hey and sorry it's taken us ages to get back to this PR. 😞 

Anyway I'm thinking we could do better here – what if we import the lossless, detailed mapping as, say, `detailed_windows_zone_mapping` or similar? The shape could be `{(zone, territory): windows_zone, ...}`.

Then it's trivial to create `windows_zone_mapping` during CLDR import time from that data. That'd leave us backward compatible API-wise (even if I don't exactly truly believe too many people are currently using `windows_zone_mapping` – it's hard to tell since Github's code search just turns up thousands of vendored copies of Babel) so this could be shipped in Babel 2.x.

ecederstrand:
This solution is fine with me.

My use case is a Microsoft Exchange client that needs to translate from `pytz` timezones to Windows timezone names, and I ended up storing a static map with the translations, generated from the CLDR: https://github.com/ecederstrand/exchangelib/blob/master/exchangelib/winzone.py So this is no longer urgent for me, but I promise to use `Babel` instead when this is solved and hits a release :-)

Issue #471, opened 2017-02-20
extract_messges seems can not ignore hidden files


dum3ng:
Say I have a folder contains some template file:

templates/
  .some.html
  .#other.html
  layout.html

what I want is to ignore the hidden files and only parse the `layout.html`. But what Babel does seems to be:
1, get all the directory and files under some directory
2, match the filename with the pattern defined in mapping_file

The second step does not take the hidden file into consideration.
And the mapping_file seems does not support the glob pattern like : `[!#]*.html` or `[a-z]*.html`.
I think use the built-in `glob` module of python will be better.

PR #476, opened 2017-03-08
Fixed in-place editing of dir list while iterating

MarcDufresne:
I was having an issue where iterating the following list of directories would still have directories with `.` at the beginning (seems to relate to #471)

All that was causing babel to find messages in the `.tox` directory which should be ignored.

Input:
```py
[
    '.git', 
    '.idea', 
    '.tox', 
    '.zanata-cache', 
    'supervisor', 
    'tests', 
    'userdoc',
    'userdoc.egg-info'
]
```

Output:
```py
[
    '.idea', 
    '.zanata-cache', 
    'supervisor', 
    'tests', 
    'userdoc', 
    'userdoc.egg-info'
]
```

So as you can see only two directories have been removed. (Happens because changing the list while iterating over it messes up the iterator, thus causing skips in the list)

My solution creates a copy of the list and changes the copy while iterating over the original. Once iteration is over we assign the value on the copy to the original.

Output with this patch:
```py
[
    'supervisor', 
    'tests', 
    'userdoc', 
    'userdoc.egg-info'
]
```
We can clearly see that all directories starting with `.` have been removed and the function now works as intended.

### DIFF
--- a/babel/messages/catalog.py
+++ b/babel/messages/catalog.py
@@ -395,6 +395,7 @@ class Catalog(object):
             elif name == 'last-translator':
                 self.last_translator = value
             elif name == 'language':
+                value = value.replace('-', '_')
                 self.locale = Locale.parse(value)
             elif name == 'language-team':
                 self.language_team = value
--- a/tests/messages/test_pofile.py
+++ b/tests/messages/test_pofile.py
@@ -36,6 +36,12 @@ msgstr ""
 "Language: en_US\n"''')
         catalog = pofile.read_po(buf, locale='de')
         self.assertEqual(Locale('en', 'US'), catalog.locale)
+        buf = StringIO(r'''
+msgid ""
+msgstr ""
+"Language: ko-KR\n"''')
+        catalog = pofile.read_po(buf, locale='de')
+        self.assertEqual(Locale('ko', 'KR'), catalog.locale)
 
     def test_preserve_domain(self):
         buf = StringIO(r'''msgid "foo"


akx:
I don't believe this works correctly.

[The documentation for os.walk](https://docs.python.org/3/library/os.html#os.walk) explicitly says that one should modify `dirnames` in-place (emphasis mine):

> When `topdown` is True, *the caller can modify the dirnames list in-place (perhaps using del or slice assignment)*, and `walk()` will only recurse into the subdirectories whose names remain in dirnames; [...]

The line 

```python
dirnames = filtered_dirnames
```

does _not_ modify the original object the name `dirnames` points to; it reassigns another object to that name, and the `os.walk` generator does not have knowledge of that new object.

A simpler patch that should achieve the same effect is replacing

```python
for subdir in dirnames:
```

with

```python
for subdir in dirnames[:]:
```

i.e. grabbing a copy of `dirnames` just for the filter iteration.

Though, honestly, a more Pythonic way, I guess, would be to replace the for loop with a list comprehension (note the `[:]` slice assignment, which edits the original `dirnames` object):

```python
dirnames[:] = [
  subdir
  for subdir
  in dirnames
  if not (subdir.startswith('.') or subdir.startswith('_'))
]
```

MarcDufresne:
Any updates on this guys? We're using a workaround at the moment but it would be great if it was fixed in the upstream :)

akx:
Ahhh, sorry @MarcDufresne, this got limboed due to Travis being silly. I re-restarted the build now, let's hope it can manage to post the result correctly so I can merge without having to invoke admin privileges...

akx:
... Well, that didn't work like it should have. Let me recreate the PR for you, @MarcDufresne. Sorry!

PR #482, opened 2017-03-21
Sort translations according to msgid and msgctxt

### LINKED ISSUE #481: Option --sort-output only uses msgid to sort translations
When you have several identical `msgid` with different `msgctxt`, using `--sort-ouput` will only sort the translations according to `msgid` leading to a random order in the output.

This leads to useless diff in the VCS:

```
-msgctxt "payment mode"
+msgctxt "subscription"
 msgid "Expired"
 msgstr "Expired"

-msgctxt "subscription"
+msgctxt "payment mode"
 msgid "Expired"
 msgstr "Expired"
```

I'd like Babel to also use `msgctxt` to sort the output.

xavfernandez:
Fixes #481

### DIFF
--- a/babel/messages/catalog.py
+++ b/babel/messages/catalog.py
@@ -120,8 +120,8 @@ class Message(object):
         """Compare Messages, taking into account plural ids"""
         def values_to_compare(obj):
             if isinstance(obj, Message) and obj.pluralizable:
-                return obj.id[0]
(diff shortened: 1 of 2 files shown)


akx:
Looks great! Thank you! 😻

xavfernandez:
Thanks for your work 👍

PR #491, opened 2017-04-07
Add currency utilities and helpers.

kdeldycke:
Add a couple of new utilities for currencies:
* `list_currency`: to list the whole set of currency codes supported by Babel.
* `validate_currency `.
* `is_currency `: to quickly validate if a currency is supported by Babel or not, not unlike what is currently available for locale (see: `babel.localedata.exists()`).
* `normalize_currency`: to normalize a currency string to the format expected by Babel.
* `get_currency_precision`: to get the natural precision of a currency.

### DIFF
--- a/babel/numbers.py
+++ b/babel/numbers.py
@@ -20,12 +20,87 @@
 #  - http://www.unicode.org/reports/tr35/ (Appendix G.6)
 import re
 from datetime import date as date_, datetime as datetime_
+from itertools import chain
 
 from babel.core import default_locale, Locale, get_global
-from babel._compat import decimal
+from babel._compat import decimal, string_types
+from babel.localedata import locale_identifiers
 
 
 LC_NUMERIC = default_locale('LC_NUMERIC')
+
+
+class UnknownCurrencyError(Exception):
+    """Exception thrown when a currency is requested for which no data is available.
+    """
+
+    def __init__(self, identifier):
+        """Create the exception.
+        :param identifier: the identifier string of the unsupported currency
+        """
+        Exception.__init__(self, 'Unknown currency %r.' % identifier)
+
+        #: The identifier of the locale that could not be found.
+        self.identifier = identifier
+
+
+def list_currencies(locale=None):
+    """ Return a `set` of normalized currency codes.
+
+    .. versionadded:: 2.5.0
+
+    :param locale: filters returned currency codes by the provided locale.
+                   Expected to be a locale instance or code. If no locale is
+                   provided, returns the list of all currencies from all
+                   locales.
+    """
+    # Get locale-scoped currencies.
+    if locale:
+        currencies = Locale.parse(locale).currencies.keys()
+    else:
+        currencies = get_global('all_currencies')
+    return set(currencies)
+
+
+def validate_currency(currency, locale=None):
+    """ Check the currency code is recognized by Babel.
+
+    Accepts a ``locale`` parameter for fined-grained validation, working as
+    the one defined above in ``list_currencies()`` method.
+
+    Raises a `ValueError` exception if the currency is unknown to Babel.
+    """
+    if currency not in list_currencies(locale):
+        raise UnknownCurrencyError(currency)
+
+
+def is_currency(currency, locale=None):
+    """ Returns `True` only if a currency is recognized by Babel.
+
+    This method always return a Boolean and never raise.
+    """
+    if not currency or not isinstance(currency, string_types):
+        return False
+    try:
+        validate_currency(currency, locale)
+    except UnknownCurrencyError:
+        return False
+    return True
+
+
+def normalize_currency(currency, locale=None):
+    """Returns the normalized sting of any currency code.
+
+    Accepts a ``locale`` parameter for fined-grained validation, working as
+    the one defined above in ``list_currencies()`` method.
+
+    Returns None if the currency is unknown to Babel.
+    """
+    if isinstance(currency, string_types):
+        currency = currency.upper()
+    if not is_currency(currency, locale):
+        return
+    return currency
 
 
 def get_currency_name(currency, count=None, locale=LC_NUMERIC):
@@ -36,10 +111,10 @@ def get_currency_name(currency, count=None, locale=LC_NUMERIC):
 
     .. versionadded:: 0.9.4
 
-    :param currency: the currency code
+    :param currency: the currency code.
     :param count: the optional count.  If provided the currency name
                   will be pluralized to that number if possible.
-    :param locale: the `Locale` object or locale identifier
+    :param locale: the `Locale` object or locale identifier.
     """
     loc = Locale.parse(locale)
     if count is not None:
@@ -56,10 +131,24 @@ def get_currency_symbol(currency, locale=LC_NUMERIC):
     >>> get_currency_symbol('USD', locale='en_US')
     u'$'
 
-    :param currency: the currency code
-    :param locale: the `Locale` object or locale identifier
+    :param currency: the currency code.
+    :param locale: the `Locale` object or locale identifier.
     """
     return Locale.parse(locale).currency_symbols.get(currency, currency)
+
+
+def get_currency_precision(currency):
+    """Return currency's precision.
+
+    Precision is the number of decimals found after the decimal point in the
+    currency's format pattern.
+
+    .. versionadded:: 2.5.0
+
+    :param currency: the currency code.
+    """
+    precisions = get_global('currency_fractions')
+    return precisions.get(currency, precisions['DEFAULT'])[0]
 
 
 def get_territory_currencies(territory, start_date=None, end_date=None,
@@ -102,7 +191,7 @@ def get_territory_currencies(territory, start_date=None, end_date=None,
 
     .. versionadded:: 2.0
 
-    :param territory: the name of the territory to find the currency fo
+    :param territory: the name of the territory to find the currency for.
     :param start_date: the start date.  If not given today is assumed.
     :param end_date: the end date.  If not given the start date is assumed.
     :param tender: controls whether tender currencies should be included.
@@ -326,12 +415,8 @@ def format_currency(number, currency, format=None, locale=LC_NUMERIC,
             raise UnknownCurrencyFormatError("%r is not a known currency format"
                                              " type" % format_type)
     if currency_digits:
-        fractions = get_global('currency_fractions')
-        try:
-            digits = fractions[currency][0]
-        except KeyError:
-            digits = fractions['DEFAULT'][0]
-        frac = (digits, digits)
+        precision = get_currency_precision(currency)
+        frac = (precision, precision)
     else:
         frac = None
     return pattern.apply(number, locale, currency=currency, force_frac=frac)
--- a/scripts/import_cldr.py
+++ b/scripts/import_cldr.py
@@ -12,6 +12,7 @@
 # individuals. For the exact contribution history, see the revision
 # history and logs, available at http://babel.edgewall.org/log/.
 
+import collections
 from optparse import OptionParser
 import os
 import re
@@ -218,6 +219,7 @@ def parse_global(srcdir, sup):
     likely_subtags = global_data.setdefault('likely_subtags', {})
     territory_currencies = global_data.setdefault('territory_currencies', {})
     parent_exceptions = global_data.setdefault('parent_exceptions', {})
+    all_currencies = collections.defaultdict(set)
     currency_fractions = global_data.setdefault('currency_fractions', {})
     territory_languages = global_data.setdefault('territory_languages', {})
     bcp47_timezone = parse(os.path.join(srcdir, 'bcp47', 'timezone.xml'))
@@ -286,14 +288,18 @@ def parse_global(srcdir, sup):
         region_code = region.attrib['iso3166']
         region_currencies = []
         for currency in region.findall('./currency'):
+            cur_code = currency.attrib['iso4217']
             cur_start = _parse_currency_date(currency.attrib.get('from'))
             cur_end = _parse_currency_date(currency.attrib.get('to'))
-            region_currencies.append((currency.attrib['iso4217'],
-                                      cur_start, cur_end,
-                                      currency.attrib.get(
-                                          'tender', 'true') == 'true'))
+            cur_tender = currency.attrib.get('tender', 'true') == 'true'
+            # Tie region to currency.
+            region_currencies.append((cur_code, cur_start, cur_end, cur_tender))
+            # Keep a reverse index of currencies to territorie.
+            all_currencies[cur_code].add(region_code)
         region_currencies.sort(key=_currency_sort_key)
         territory_currencies[region_code] = region_currencies
+    global_data['all_currencies'] = dict([
+        (currency, tuple(sorted(regions))) for currency, regions in all_currencies.items()])
 
     # Explicit parent locales
     for paternity in sup.findall('.//parentLocales/parentLocale'):
--- a/babel/core.py
+++ b/babel/core.py
@@ -45,6 +45,7 @@ def get_global(key):
 
     The keys available are:
 
+    - ``all_currencies``
     - ``currency_fractions``
     - ``language_aliases``
     - ``likely_subtags``
--- a/tests/test_numbers.py
+++ b/tests/test_numbers.py
@@ -17,6 +17,10 @@ import pytest
 from datetime import date
 
 from babel import numbers
+from babel.numbers import (
+    list_currencies, validate_currency, UnknownCurrencyError, is_currency, normalize_currency, get_currency_precision)
+from babel.core import Locale
+from babel.localedata import locale_identifiers
 from babel._compat import decimal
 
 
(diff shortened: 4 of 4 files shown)


babel/numbers.py
```diff
@@ -20,46 +20,104 @@
...
+from babel.core import default_locale, Locale, get_global, UnknownLocaleError
 from babel._compat import decimal
+from babel.localedata import locale_identifiers
 
 
 LC_NUMERIC = default_locale('LC_NUMERIC')
 
 
-def get_currency_name(currency, count=None, locale=LC_NUMERIC):
+def list_currency(locale=None):
```
akx: This should be `list_currencies`, I think.

babel/numbers.py
```diff
@@ -20,46 +20,104 @@
...
+    """
+    # Get locale-scopped currencies.
+    if locale:
+        return Locale.parse(locale).currencies.keys()
+    # Aggregate currency codes from all locales.
+    return set(chain.from_iterable([
+        Locale.parse(l).currencies.keys() for l in locale_identifiers()]))
+
+
+def is_currency(currency_code, locale=None):
```
akx: Maybe `is_currency_supported`? Or `validate_currency`?

babel/numbers.py
```diff
@@ -20,46 +20,104 @@
...
+    """ Return a list of normalized currency codes.
+
+    .. versionadded:: 2.5.0
+
+    :param locale: filters returned currency codes by the provided locale.
+                   Expected to be a locale instance or code. If no locale is
+                   provided, returns the list of all currencies from all
+                   locales.
+    """
+    # Get locale-scopped currencies.
```
akx: *scoped :)

babel/numbers.py
```diff
@@ -20,46 +20,104 @@
...
 from babel._compat import decimal
+from babel.localedata import locale_identifiers
 
 
 LC_NUMERIC = default_locale('LC_NUMERIC')
 
 
-def get_currency_name(currency, count=None, locale=LC_NUMERIC):
+def list_currency(locale=None):
+    """ Return a list of normalized currency codes.
```
akx: The docstring is lying in that this may return either a `dict_keys` object (when `locale` is passed in) or a `set`, but never a list.

Maybe always return a `set`?

babel/numbers.py
```diff
@@ -20,46 +20,104 @@
...
+                   Expected to be a locale instance or code. If no locale is
+                   provided, returns the list of all currencies from all
+                   locales.
+    """
+    # Get locale-scopped currencies.
+    if locale:
+        return Locale.parse(locale).currencies.keys()
+    # Aggregate currency codes from all locales.
+    return set(chain.from_iterable([
+        Locale.parse(l).currencies.keys() for l in locale_identifiers()]))
```
akx: Yigh, parsing and instantiating every known locale sounds a little expensive, especially since most of this will probably end up being a no-op (in that there's likely no specific language that provides a currency code the other languages don't know about).

Could this be done in some other way? Maybe at CLDR import time? An `all_currencies` key in the `global` data? Could `currency_fractions` be used?

babel/numbers.py
```diff
@@ -20,46 +20,104 @@
...
+def is_currency(currency_code, locale=None):
+    """ Check the currency code is supported by Babel.
+
+    Accepts a ``locale`` parameter for fined-grained validation, working as
+    the one defined above in ``list_currency()`` method.
+    """
+    try:
+        if currency_code not in list_currency(locale):
+            return False
+    except (ValueError, UnknownLocaleError):
```
akx: Hmmm... Maybe `UnknownLocaleError` shouldn't be swallowed? Also, when might one expect a `ValueError`?

babel/numbers.py
```diff
@@ -20,46 +20,104 @@
...
+        return False
+    return True
+
+
+def normalize_currency(currency_code, locale=None):
+    """ Normalize any currency code string.
```
akx: Hmm, so "normalization" here means just upper-casing the code, then validating it?

The docstring should at least let the user know that `ValueError` will be raised if the code is not known to Babel.

babel/numbers.py
```diff
@@ -20,46 +20,104 @@
...
+def get_currency_name(currency_code, count=None, locale=LC_NUMERIC):
     """Return the name used by the locale for the specified currency.
 
     >>> get_currency_name('USD', locale='en_US')
     u'US Dollar'
 
     .. versionadded:: 0.9.4
 
-    :param currency: the currency code
+    :param currency: the currency ISO code.
```
akx: If you're changing the names of the parameters, also change the docstrings to match. That said, I'm not sure `currency` needs to be changed to `currency_code` throughout -- everyone should know it's a code. (Not to mention changing these parameter names would also be a semver-major change.)

babel/numbers.py
```diff
@@ -20,46 +20,104 @@
...
-    :param locale: the `Locale` object or locale identifier
+    :param currency: the currency ISO code.
+    :param locale: the `Locale` object or locale identifier.
+    """
+    return Locale.parse(locale).currency_symbols.get(
+        currency_code, currency_code)
+
+
+def currency_precision(currency_code):
+    """Return currency's precision.
```
akx: A description of what "precision" means would be nice.

babel/numbers.py
```diff
@@ -20,46 +20,104 @@
...
-    :param currency: the currency code
-    :param locale: the `Locale` object or locale identifier
+    :param currency: the currency ISO code.
+    :param locale: the `Locale` object or locale identifier.
+    """
+    return Locale.parse(locale).currency_symbols.get(
+        currency_code, currency_code)
+
+
+def currency_precision(currency_code):
```
akx: This oughta be `get_currency_precision` (to have a verb in all method names).

kdeldycke:
Thanks @akx for the thorough and quick review! I'm going to answer and fix each of your comment monday.

akx:
@kdeldycke Sure thing! Enjoy your weekend!

babel/numbers.py
kdeldycke: Done in 9b231741e884c5167e276eeb7f0e13fcd5f13538. Thanks for the feedback! :)

babel/numbers.py
kdeldycke: Oh yes. Makes total sense. Done in 7ce48eec93344dba50b183b9c4b1385e885b18a6.

babel/numbers.py
kdeldycke: Good catch! Done in e130ed4ed161cc9c57c5d8ed424b39c872b1fc2f.

babel/numbers.py
kdeldycke: I'm not too happy with that brute force approach either. At the time I rote that wasn't really certain about the completeness of the CLDR data, hence the way I took.

I think I'll first ensure with a unit-test that, as you said, that there's no specific language providing a currency code any other languages don't know about. If we can verify this assumption then I'll try to dig into the CLDR import code and see what I can do.

babel/numbers.py
kdeldycke: Oh yes, sorry for that. It was too ambitious and `currency` as it is now is good enough.

I reverted in commit 5ba7c7b0debb650a0f0e80e3ab8a196ac9549193.

babel/numbers.py
kdeldycke: Done in feaa8935d6b4d1cb36ecf2183050c2f0ebc3a35c.

babel/numbers.py
kdeldycke: Done: 1af1d8a09bc0716c000dcec71accbac221586321.

babel/numbers.py
kdeldycke: The validation of currency is really what I'm trying to do here.

Now I'm used to have an `is_currency` and `validate_currency` couple helpers when it comes to data validation. I modified the code in f89d4d21cb0c6233cdf4d3abaf84d890c857d72b to reflect that. Do you like it?

babel/numbers.py
kdeldycke: Good point. Done in 6fd2539e7f1e069b6b8aeadd965881315c24abc4.

babel/numbers.py
kdeldycke: Currently yes: normalization is only upper-casing. But that method allows us to enhance normalization in more clever way in the future, if the need arises.

As for the `ValueError`, I just did the change in 5d9f416a4cca5eaeedd3dea0124c970d26e7ad9d.

babel/numbers.py
```diff
@@ -20,14 +20,73 @@
...
+def validate_currency(currency, locale=None):
+    """ Check the currency code is recognized by Babel.
+
+    Accepts a ``locale`` parameter for fined-grained validation, working as
+    the one defined above in ``list_currencies()`` method.
+
+    Raises a `ValueError` exception if the currency is unknown to Babel.
+    """
+    if currency not in list_currencies(locale):
+        raise ValueError("Unknown currency {!r}.".format(currency))
```
akx: Maybe a custom `UnknownCurrencyError` exception (that derives from `ValueError`) could be useful here? Easier to catch in caller functions, too :)

babel/numbers.py
```diff
@@ -56,10 +115,25 @@ def get_currency_symbol(currency, locale=LC_NUMERIC):
...
     u'$'
 
-    :param currency: the currency code
-    :param locale: the `Locale` object or locale identifier
+    :param currency: the currency code.
+    :param locale: the `Locale` object or locale identifier.
     """
-    return Locale.parse(locale).currency_symbols.get(currency, currency)
+    return Locale.parse(locale).currency_symbols.get(
+        currency, currency)
```
akx: Super minor nitpick, but no need to re-wrap this line :)

We have max-line-length 120 in the project.

kdeldycke:
Thanks @akx for that second round of reviews. Going to fix all remaining issues and add tests. I'll notify you once everything is clean and tidy.

babel/numbers.py
kdeldycke: So that was what I suspected all along. The reason I'm iterating on all locales is that not all locales have the same knowledge of currencies:
```python
>>> from babel.core import Locale
>>> from babel.localedata import locale_identifiers
>>> import babel
>>> babel.__version__
'2.4.0'
>>> currency_sets = set()
>>> for locale in locale_identifiers():
...     currency_sets.add(frozenset(Locale.parse(locale).currencies.keys()))
... 
>>> currency_sets
set([frozenset([]),
     frozenset(['CVE', 'NGN']),
     frozenset(['XAF']),
     frozenset(['AZN']),
(...)
     frozenset(['AED',
                'AFA',
                'AFN',
                'ALL',
                'AMD',
                'ANG',
                'AOA',
                'ARS',
                'AUD',
                'AWG',
                'AZN',
                'BAM',
                'BBD',
                'BDT',
                'BGN',
                'BHD',
                'BIF',
                'BMD',
                'BND',
                'BOB',
                'BRL',
                'BSD',
                'BTN',
                'BWP',
                'BYR',
                'BZD',
                'CAD',
                'CDF',
                'CHF',
                'CLP',
                'CNY',
                'COP',
                'CRC',
                'CUC',
                'CUP',
                'CVE',
                'CZK',
                'DJF',
                'DKK',
                'DOP',
                'DZD',
                'EGP',
                'ERN',
                'ETB',
                'EUR',
                'FJD',
                'FKP',
                'GBP',
                'GEL',
                'GHS',
                'GIP',
                'GMD',
                'GNF',
                'GTQ',
                'GYD',
                'HKD',
                'HNL',
                'HRK',
                'HTG',
                'HUF',
                'IDR',
                'ILS',
                'INR',
                'IQD',
                'IRR',
                'ISK',
                'JMD',
                'JOD',
                'JPY',
                'KES',
                'KGS',
                'KHR',
                'KMF',
                'KPW',
                'KRW',
                'KWD',
                'KYD',
                'KZT',
                'LAK',
                'LBP',
                'LKR',
                'LRD',
                'LTL',
                'LVL',
                'LYD',
                'MAD',
                'MDL',
                'MGA',
                'MKD',
                'MMK',
                'MNT',
                'MOP',
                'MRO',
                'MUR',
                'MVR',
                'MWK',
                'MXN',
                'MYR',
                'MZN',
                'NAD',
                'NGN',
                'NIO',
                'NOK',
                'NPR',
                'NZD',
                'OMR',
                'PAB',
                'PEN',
                'PGK',
                'PHP',
                'PKR',
                'PLN',
                'PYG',
                'QAR',
                'RON',
                'RSD',
                'RUB',
                'RWF',
                'SAR',
                'SBD',
                'SCR',
                'SDG',
                'SEK',
                'SGD',
                'SHP',
                'SLL',
                'SOS',
                'SRD',
                'SSP',
                'STD',
                'SYP',
                'SZL',
                'THB',
                'TJS',
                'TMT',
                'TND',
                'TOP',
                'TRY',
                'TTD',
                'TWD',
                'TZS',
                'UAH',
                'UGX',
                'USD',
                'UYU',
                'UZS',
                'VEF',
                'VND',
                'VUV',
                'WST',
                'XAF',
                'XCD',
                'XOF',
                'XPF',
                'XXX',
                'YER',
                'ZAR',
                'ZMK',
                'ZWR'])])
>>> len(currency_sets)
105
```

babel/numbers.py
kdeldycke: Done in 25f7839608ca0a372f1149c69d7e34d830d43a3a.

babel/numbers.py
kdeldycke: Done in 188caf3d7435afcd0a3d9282429cca13406acdcc.

kdeldycke:
Just added unit-tests on all methods and squashed commits.

The only thing left to do is to get rid of the brute force approach in `list_currencies`. I'll work on that now.

kdeldycke:
I'm finished working on this PR. All issues have been addressed.

@akx : unless you have further comments to add, this PR is ready to be merged.

kdeldycke:
The tests seems to fail for an unrelated reason... 😢

kdeldycke:
Everything's fine now. Just waiting for @akx final review before merging. 😃

benselme:
So, should we merge this ?

kdeldycke:
Oh yes please! Let's merge this! 😃 

AFAIK this PR is just waiting for @akx to lift the reviewing embargo.

akx:
Sorry for the delay! Let me just re-read the final version :)

scripts/import_cldr.py
```diff
@@ -286,15 +287,20 @@ def parse_global(srcdir, sup):
...
+            cur_tender = currency.attrib.get('tender', 'true') == 'true'
+            # Tie region to currency.
+            region_currencies.append((cur_code, cur_start, cur_end, cur_tender))
+            # Keep a reverse index of currencies to territorie.
+            all_currencies.setdefault(cur_code, []).append(region_code)
         region_currencies.sort(key=_currency_sort_key)
         territory_currencies[region_code] = region_currencies
 
+    # Deduplicate regions and makes them pickable.
+    all_currencies = dict([(code, tuple(set(regions))) for code, regions in all_currencies.items()])
```
akx: Um, this doesn't actually affect the `all_currencies` object that is already in `global_data`.
What do you mean with "pickable" (i.e. what is the requirement for the value to be a `tuple`)?

I mean, if the value was a `set` all the time, the deduplication code would be unnecessary:

```python
for currency in region.findall('./currency'):
    all_currencies.setdefault(cur_code, set()).add(region_code)
```

kdeldycke: I had to resort to that unsexy code to make the metadata serialization happy. Changing the code as you proposed ends up with that failure:

```python
____ test_compatible_classes_in_global_and_localedata[babel/global.dat] ____

filename = 'babel/global.dat'

    @pytest.mark.parametrize('filename', [
        'babel/global.dat',
        'babel/locale-data/root.dat',
        'babel/locale-data/en.dat',
        'babel/locale-data/en_US.dat',
        'babel/locale-data/en_US_POSIX.dat',
        'babel/locale-data/zh_Hans_CN.dat',
        'babel/locale-data/zh_Hant_TW.dat',
        'babel/locale-data/es_419.dat',
    ])
    def test_compatible_classes_in_global_and_localedata(filename):
        # Use pickle module rather than cPickle since cPickle.Unpickler is a method
        # on Python 2
        import pickle
    
        class Unpickler(pickle.Unpickler):
    
            def find_class(self, module, name):
                # *.dat files must have compatible classes between Python 2 and 3
                if module.split('.')[0] == 'babel':
                    return pickle.Unpickler.find_class(self, module, name)
                raise pickle.UnpicklingError("global '%s.%s' is forbidden" %
                                             (module, name))
    
        with open(filename, 'rb') as f:
>           return Unpickler(f).load()

tests/test_core.py:318: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
/System/Library/Frameworks/Python.framework/Versions/2.6/lib/python2.6/pickle.py:858: in load
    dispatch[key](self)
/System/Library/Frameworks/Python.framework/Versions/2.6/lib/python2.6/pickle.py:1090: in load_global
    klass = self.find_class(module, name)
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

self = <tests.test_core.Unpickler instance at 0x10e2e9128>, module = '__builtin__', name = 'set'

    def find_class(self, module, name):
        # *.dat files must have compatible classes between Python 2 and 3
        if module.split('.')[0] == 'babel':
            return pickle.Unpickler.find_class(self, module, name)
        raise pickle.UnpicklingError("global '%s.%s' is forbidden" %
>                                    (module, name))
E       UnpicklingError: global '__builtin__.set' is forbidden

tests/test_core.py:315: UnpicklingError
```

akx: To be honest, it looks like that test is broken, then.
`__builtin__.set` should be available alright on both Python 2 and Python 3.
(I think what's happening is that there is no builtin operation in the pickle format for `set`s, so it resorts to "let's construct a class instance of `__builtin__.set`" instead.)

Anyway, fair! How about

```python
all_currencies = collections.defaultdict(set)
for currency in region.findall('./currency'):
    # ...
    all_currencies[cur_code].add(region_code)
    # ...
global_data['all_currencies'] = {
    currency: tuple(sorted(regions))
    for (currency, regions)
    in all_currencies.items()
}
```
?

kdeldycke: Done in 4309726d56302e1c472dece78f5b7e6647b6c6ad.

kdeldycke:
Just addressed the last comment and squashed all commits.

This PR is ready to be merged.

akx:
Good to go! Thanks for the code and for putting up with me ;), @kdeldycke!

kdeldycke:
Thanks @akx for the merge! No worries, I'm happy to share the burden of open-source project maintenance! :)

PR #492, opened 2017-04-10
Fixed in-place editing of dir list while iterating

akx:
This is a reissue of #476, since Git and Github and Travis were being silly.


akx:
Thanks @MarcDufresne :)

PR #498, opened 2017-05-27
Fix an invalid string to bytes comparison in Python 3

rowillia:
In Python 3 this branch would never evaluate to `False` due to comparing a string to bytes.

This was flagged by running some unit tests with the `-bb` flag, so I've added
`-bb` to run with the babel unit tests in Python3.5

Here's some docs on the `-bb` flag:
https://docs.python.org/3.5/library/warnings.html#default-warning-filters
https://docs.python.org/3/howto/pyporting.html#use-continuous-integration-to-stay-compatible

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -33,6 +33,8 @@ matrix:
       python: 3.4
     - os: linux
       python: 3.5
+      env:
+      - PYTHON_TEST_FLAGS=-bb
 
 install:
   - bash .ci/deps.${TRAVIS_OS_NAME}.sh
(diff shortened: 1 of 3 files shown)


rowillia:
Ping @akx !  Anything else I can do here?  It's a pretty straight forward fix and I'd love to get this in (this issue prevents us from, for example, running our unit tests with the `-bb` flag on with an unpatched version of babel)

rowillia:
@akx Would it be possible to cut a release with this fix in it?

PR #504, opened 2017-06-19
introduce PyInstaller support. Fixes #500

### LINKED ISSUE #500: Using __file__ in a PyInstaller environment is not possible
Babel is not working in my projects which are processed with PyInstaller.
The reason for this: the direct access to "`__file__`".

wodo:


benselme:
Thanks, that's much better. 

I still think we shouldn't be creating a new module (`pi_support.py`) for just one function.

Sorry for the CI mess. @akx, @etanol, any idea what's happening ?

etanol:
I agree with @benselme:

> I still think we shouldn't be creating a new module (`pi_support.py`) for just one function.

@wodo, can you move the function to `babel/core.py` or `babel/__init__.py`?

wodo:
Yes, I will try to do it.

wodo:
Implementing `get_base_dir` in core.py results in an import cycle

benselme:
What about `babel/__init__.py` then ?

wodo:
Implementing in `babel/__init__.py` results in an import cycle.
The same problem when implementing it in `babel/util.py`.

benselme:
It actually works in `util.py`... I just pushed that change to your PR, and better tests. 

See if it's ok with you, then you can squash yet again, and then we can merge.

Thanks !

wodo:
Hello @benselme,
Thanks, but the appveyor is reporting the same troubles...
Its not working.

benselme:
Ha I see ! The tests pass locally, but it's setup.py that crashes. You're right, that's not good. Let me look into that a bit more. `localedata` is not a great place for `get_base_dir` either.

wodo:
I close this PR and will open a follow up.
Using `babel/core.py` or `babel/util.py` seems not possible.
I'm not able to squashig all this stuff...

PR #506, opened 2017-06-24
Fix Travis CI configuration

etanol:
The python versions defined in the Travis CI configuration may be parsed as decimals.  This could be the reason behind the latest Travis *failures*.

First attempt to fix the situation.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -12,15 +12,15 @@ cache:
 matrix:
   include:
     - os: linux
-      python: 2.6
+      python: "2.6"
     - os: linux
-      python: 2.6
+      python: "2.6"
       env:
       - CDECIMAL=m3-cdecimal
     - os: linux
-      python: 2.7
+      python: "2.7"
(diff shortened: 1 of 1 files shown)


benselme:
It looks like it's commit be718f3a7b2b71e936dcc3b498b7013537aa9036 that broke things

benselme:
The appveyor failure is because of a test that uses `datetime.now`.

benselme:
So I reverted all changes made in latest merged PR, and it still fails.
There's also this very weird codecov behaviour with reports of decreased coverage when there shouldn't be any.

akx:
@benselme Regarding @codecov-io, I've found it occasionally gets confused about multi-environment coverage (and especially now that *NIX coverage isn't being measured as Travis is MIA). Don't worry about it (yet).

etanol:
Dropping this in favor of the fix for travis-ci/travis-ci#7974.

PR #508, opened 2017-06-25
Revert changes to makefile to fix travis builds

benselme:
Some changes introduced in #498 apparently made travis
fail on newer pull requests.


akx:
This probably shouldn't revert the Python change though?

benselme:
Sure, but I was getting desperate :)
I will just close this. The problem lies elsewhere obviously.

PR #510, opened 2017-07-11
don't crash with a multiline import

pablogamboa:
the real fix would modify the regex to deal properly with multiline imports


akx:
@pablogamboa Can you explain the problem, or preferably add a test case that's fixed by this?

PR #512, opened 2017-07-16
Doc: always redirect to the python gettext module

Linkid:
The link to the python gettext module was dead in the doc. So, redirect the user
to the official doc.


akx:
Thank you! :)

Issue #514, opened 2017-08-03
Class DateTimePattern breaks API under Python 3.


sfermigier:
`DateTimePattern` has a `__unicode__` magic method, but no `__str__`.

PR #515, opened 2017-08-03
Fixes Python 3 compatibility issue.

sfermigier:
Cf. issue #514

### DIFF
--- a/babel/dates.py
+++ b/babel/dates.py
@@ -1214,16 +1214,14 @@ class DateTimePattern(object):
     def __repr__(self):
         return '<%s %r>' % (type(self).__name__, self.pattern)
 
-    if PY2:
-        def __unicode__(self):
-            return self.pattern
+    def __unicode__(self):
+        return self.pattern
 
-        def __str__(self):
-            self.__unicode__().encode('utf-8')
-
-    else:
-        def __str__(self):
-            return self.pattern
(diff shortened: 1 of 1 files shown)


babel/dates.py
```diff
@@ -1215,8 +1214,16 @@ def __init__(self, pattern, format):
...
         return '<%s %r>' % (type(self).__name__, self.pattern)
 
-    def __unicode__(self):
-        return self.pattern
+    if PY2:
+        def __unicode__(self):
+            return self.pattern
+
+        def __str__(self):
+            self.__unicode__().encode('utf-8')
```
akx: This statement is missing its `return` statement.

babel/dates.py
```diff
@@ -1215,8 +1214,16 @@ def __init__(self, pattern, format):
     def __repr__(self):
         return '<%s %r>' % (type(self).__name__, self.pattern)
 
-    def __unicode__(self):
-        return self.pattern
+    if PY2:
```
akx: To be honest, I'd be more comfortable if the PY2 code lived within the functions instead of the class definition changing depending on interpreter version:

```python
def __unicode__(self):
    return self.pattern

def __str__(self):
    pat = self.pattern
    if PY2:
        pat = pat.encode('utf-8')
    return pat
```

PR #519, opened 2017-08-18
Parse multi-line __future__ imports better

akx:
Augments #426
Augments #427
Augments #510
Closes #510


akx:
@sublee, care to take a look? :)

sublee:
@akx Looks better.

I suggest you to add more multi-line import styles for unit testing:

```python
from __future__ import (
    print_function,
    division)
```

```python
from __future__ import \
    print_function, \
    division
```

akx:
@sublee Thanks! Excellent points -- neither of those cases actually worked. They do now! :)

Issue #521, opened 2017-08-28
Validation / normalization of locales and currencies differ in behavior


svisser:
Babel 2.5.0

For currencies the functions `babel.numbers.is_currency` and `babel.numbers.normalize_currency` accept quite a few values (None, strings, numbers, lists, sets) and return the expected result (a boolean or a string).

The functions `babel.localedata.normalize_locale` and `babel.localedata.exists` raise an exception in most cases except when a string is provided.

It would be nice if these functions behaved the same way (at least for `None`?) so that the following approach can be used:

```python
my_locale = babel.localedata.normalize_locale(my_locale)
my_currency = babel.numbers.normalize_currency(my_currency)

# at this point, the values are correct strings or None
```

PR #523, opened 2017-09-09
modified normalize_locale and exists to handle various unexpected input

suhojm:
made normalize_locale and exists to handle various unexpected input such as list or None just like currency, and added test cases in test_localedata.py
resolves: #521

### DIFF
--- a/babel/localedata.py
+++ b/babel/localedata.py
@@ -47,8 +47,8 @@ def normalize_locale(name):
     for locale_id in chain.from_iterable([_cache, locale_identifiers()]):
         if name == locale_id.lower():
             return locale_id
-    
-    
+
+
 def exists(name):
     """Check whether locale data is available for the given locale.
 
--- a/babel/localedata.py
+++ b/babel/localedata.py
@@ -48,6 +48,7 @@ def normalize_locale(name):
         if name == locale_id.lower():
             return locale_id
     
+    
 def exists(name):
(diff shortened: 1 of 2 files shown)


babel/localedata.py
```diff
@@ -41,6 +41,8 @@ def normalize_locale(name):
     Returns the normalized locale ID string or `None` if the ID is not
     recognized.
     """
+    if not name or not isinstance(name, string_types):
```
akx: It would probably be better and/or more Pythonic to just cast to the text type here, instead of strictly requiring a string type.

babel/localedata.py
```diff
@@ -54,6 +56,8 @@ def exists(name):
 
     :param name: the locale identifier string
     """
+    if not name or not isinstance(name, string_types):
```
akx: Same here: It would probably be better and/or more Pythonic to just cast to the text type here, instead of strictly requiring a string type.

svisser: Perhaps that's functionality that could be implemented at a later date?

Using `isinstance()` would make it match the currency functions, e.g., https://github.com/python-babel/babel/blob/master/babel/numbers.py#L82.

I'm also not sure whether accepting an object for which `str(obj)` returns a locale should be supported long-term?

akx: I suppose you're correct. Didn't remember the currency functions already acted like this anyway.

PR #528, opened 2017-09-15
Invalid string escape sequence related fixes

scop:


scop:
Rebased against current master.

akx:
Thank you! (Kiitos! :) )

Issue #531, opened 2017-09-25
Allow option to abort when po file is invalid


Bedrock02:
# Overview Description
If given the following snippet of po file. Reading the po file will log a warning that the po file is invalid. If not aborted it will attempt to log the invalid line(s). If using py27, it will try to format a string with unicode characters causing the app to break.

```
"naqern.enlabeuragfpury@qsb-zcb.tp.pn."
msgstr "Merci de prendre le temps de remplir le sondage.
Pour toute question, veuillez communiquer avec Gvz Crgref à gvz.crgref@qsb-zcb.tp.pn ou le 613-698-5385, ou avec Naqern Enlabe Uragfpury à naqern.enlabeuragfpury@qsb-zcb.tp.pn
"
```

# Steps to reproduce
1. Create a PO file with an invalid line (see snippet above)
2. read_po

# Expected Results
1. To be able to log the invalid line or just abort reading the po file

PR #532, opened 2017-09-26
pofile.py: Added new exception called PoFileError and thrown if flagged

Bedrock02:
This new exception is thrown when the po parser finds an invalid pofile.
This helps handle invalid po files that are parsed. Invalid po files may cause
other possible errors such as a UnicodeEncodeError.

Closes https://github.com/python-babel/babel/issues/531

### DIFF
--- a/babel/messages/pofile.py
+++ b/babel/messages/pofile.py
@@ -19,6 +19,7 @@ from babel.util import wraptext
 from babel._compat import text_type
 
 
+
 def unescape(string):
     r"""Reverse `escape` the given string.
 
@@ -73,6 +74,15 @@ def denormalize(string):
         return unescape(string)
 
 
+class PoFileError(Exception):
+    """Exception thrown by PoParser when an invalid po file is encountered."""
+    def __init__(self, message, catalog, line, lineno):
+        super(PoFileError, self).__init__('{message} on {lineno}'.format(message=message, lineno=lineno))
+        self.catalog = catalog
+        self.line = line
+        self.lineno = lineno
+
+
 class _NormalizedString(object):
 
     def __init__(self, *args):
@@ -104,11 +114,12 @@ class PoFileParser(object):
         'msgid_plural',
     ]
 
-    def __init__(self, catalog, ignore_obsolete=False):
+    def __init__(self, catalog, ignore_obsolete=False, abort_invalid=False):
         self.catalog = catalog
         self.ignore_obsolete = ignore_obsolete
         self.counter = 0
         self.offset = 0
+        self.abort_invalid = abort_invalid
         self._reset_message_state()
 
     def _reset_message_state(self):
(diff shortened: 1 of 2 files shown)


babel/messages/pofile.py
```diff
@@ -18,6 +18,8 @@
 from babel.util import wraptext
 from babel._compat import text_type
 
+__all__ = ['PoFileError']
```
akx: Remove this, please. We haven't had an `__all__` stanza in this file, so adding one that just exports a single name will break `from babel.messages.pofile import *` imports in client code.

babel/messages/pofile.py
```diff
@@ -73,6 +75,10 @@ def denormalize(string):
         return unescape(string)
 
 
+class PoFileError(Exception):
+    """Exception thrown by PoParser when an invalid po file is encountered."""
```
akx: It would be nice if this was a more full-fledged exception, to make it easier to use programmatically by client code:

```python
class PoFileError(Exception):
  def __init__(self, message, catalog, line, lineno):
    super(PoFileError, self).__init__('{message} on {lineno}'.format(message=message, lineno=lineno))
    self.catalog = catalog
    self.line = line
    self.lineno = lineno
```
and naturally the `raise` below would be
```python
raise PoFileError(msg, self.catalog, line, lineno)
```
This way client code can introspect the error without having to try to parse the message.

What do you think?

babel/messages/pofile.py
```diff
@@ -276,11 +283,13 @@ def parse(self, fileobj):
             self._add_message()
 
     def _invalid_pofile(self, line, lineno, msg):
+        if self.abort_invalid:
+            raise PoFileError("Invalid POFile encountered line_number={}, msg={}".format(lineno, msg))
```
akx: See the discussion above about a better exception :)

tests/messages/test_pofile.py
```diff
@@ -429,6 +429,62 @@ def test_missing_plural_in_the_middle(self):
...
 
+    def test_abort_invalid_po_file(self):
+        invalid_po = '''
+            msgctxt ""
+            "{\"checksum\": 2148532640, \"cxt\": \"collector_thankyou\", \"id\": "
+            "270005359}"
+            msgid ""
+            "Thank you very much for your time.\n"
+            "If you have any questions regarding this survey, please contact Tim "
+            "at andrea.raynorhentschel@dfo.ca."
```
akx: Please make absolutely sure there aren't any emails or names belonging to real people in the test data. (Consider this duplicated for all the names and emails here.)

babel/messages/pofile.py
Bedrock02: 👍

babel/messages/pofile.py
Bedrock02: Sounds like a good idea, will implement soon

babel/messages/pofile.py
Bedrock02: 👍

tests/messages/test_pofile.py
Bedrock02: 👍

Bedrock02:
@akx thanks for the quick feedback. I applied the suggested changes in a separate commit. After changes come to a final state I could squash to one commit.

akx:
@Bedrock02 This looks good now :) Mind rebasing on top of current master and squashing the two commits while you're at it?

Bedrock02:
@akx branch has been rebased on top of the current master and I squashed the two commits. Let me know if there is anything else that I should do.

akx:
@Bedrock02 Can you rebase once more on top of the current master so we get rid of the CI failure (as since #546 was just merged we don't support Python 2.6 anymore)? Thank you!

Bedrock02:
@akx 👍

akx:
Thank you :)

PR #533, opened 2017-09-28
Fix PyInstaller support

### LINKED ISSUE #529: "Support PyInstaller" broke Pyinstaller support
PyInstaller has supported Babel for ages (it has had a hook for Babel since 2009). #505 actually broke PyInstaller support AFAICS (tested on Python 3.6 with PyInstaller dev version and Python 3.5 with the latest PyInstaller release). PyInstaller will automatically bundle the data files (`babel/locale-data`) using the hook and when the executable is run, these files are extracted to a temporary folder, say `$TMP/_MEI123456/babel/locale-data`. However the new code will look for it in `{sys._MEIPASS}/locale-data`, i.e., `$TMP/_MEI123456/locale-data` and fail.

PyInstaller actually sets `__file__` for "frozen" modules to a path relative to the temp folder that the data files get extracted to (see https://github.com/pyinstaller/pyinstaller/issues/1598#issuecomment-147753507). So the old was already doing the right thing - if it didn't work in some setup, that was really a PyInstaller bug.

yagebu:
PyInstaller+Babel was supported for quite a while before 2.5.0 - which broke support for this configuration by default. PyInstaller actually synthesizes a `__file__` for frozen aka bundled code that makes this "just work".

Closes #529.

### DIFF
--- a/babel/core.py
+++ b/babel/core.py
@@ -68,7 +68,7 @@ def get_global(key):
     """
     global _global_data
     if _global_data is None:
-        dirname = localedata.get_base_dir()
+        dirname = os.path.join(os.path.dirname(__file__))
         filename = os.path.join(dirname, 'global.dat')
         if not os.path.isfile(filename):
             _raise_no_data_error()
--- a/babel/localedata.py
+++ b/babel/localedata.py
@@ -22,7 +22,7 @@ from babel._compat import pickle, string_types
 
 
 def get_base_dir():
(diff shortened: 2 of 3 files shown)


yagebu:
Pinging @akx: Is anything blocking this from getting merged?

akx:
Hi @yagebu – I'm so sorry I dropped the ball on this one! Merging now and releasing 2.5.2 thereafter.

Issue #451, opened 2016-10-11
Locale.parse cannot parse the locale 'xh' (xhosa)


hmpf:
# Steps to reproduce:

```
from babel.core import Locale

Locale.parse('xh')
```
# Actual Results

```
    File "<censored>/.virtualenvs/test/lib/python3.5/site-packages/babel/core.py", line 333, in parse
        raise UnknownLocaleError(input_id)
babel.core.UnknownLocaleError: unknown locale 'xh'
```
# Expected Results

A locale object for `xh` that can be understood by `babel.messages.catalog.Catalog`.
# Reproducibility

Always
# Addition information:

This is for babel 2.3.4 as installed with pip and run on python 3.5.

It falls off at the final line (333): `raise UnknownLocaleError(input_id)`

Debug-by-assert shows that the first  `get_global('likely_subtags')` spits out `'xh_Latn_ZA'` which is then not handled by `_try_load_reducing(parse_locale(likely_subtag))`. The next attempt at `parse_locale` spits out `('xh', 'ZA', 'Latn', None)` which `_try_load_reducing` again interprets as None.

I'm assuming this is because there are no files prefixed with `xh` in `locale-data`?

I'm trying to load and write po-files, but it looks like I can't use babel to set the locale-header on them.

Would it at least be possible to raise another error than a generic UnknownLocaleError?

http://www.unicode.org/cldr/charts/23/summary/xh_ZA.html

jtwang:
Yep, you're correct in that the version of the CLDR we use ([v28](http://www.unicode.org/repos/cldr/tags/release-28/common/main/)) doesn't contain definitions for xh or its dialects. Same for trunk. What's strange is that your link above shows that it existed at some point (looks like the last version that contained it was [v25](http://www.unicode.org/repos/cldr/tags/release-25/common/main/)). It's mentioned in the supplementary data for v28 though. Weird!

I don't have time right now to dig through the [release notes for v26](http://cldr.unicode.org/index/downloads/cldr-26), but there might be something in there that explains why this was dropped. There could be something in the CLDR that "redirects" the locale that I don't know about (and neither does Babel). (There's lots in the CLDR that I haven't looked at, so this is totally possible :)

In the meantime, the exception is correct IMO - the CLDR and thus babel doesn't have locale data for xh so it is, technically, unknown.

[...]

PR #534, opened 2017-09-30
Feature/parsexh

valentin-p:
Improve exception handling when the language exists but is not in the locale data
#451

### DIFF
--- a/babel/core.py
+++ b/babel/core.py
@@ -105,6 +105,21 @@ class UnknownLocaleError(Exception):
         self.identifier = identifier
 
 
+class NotFoundLocaleDataError(Exception):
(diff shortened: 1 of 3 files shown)


akx:
Can you explain how this determines whether a language exists vs. not being in the locale data?

valentin-p:
Basically I assume that if we can do core.parse_locale on an identifier, this identifier is determined, then we can raise an error explaining that the data locale file does is not found. 
If this identifier cannot be parse then we should not check if the data locale file if ready or not. 

Would you prefer to check if we can download the data locale from http://www.unicode.org/repos/cldr/trunk/seed/main/xh.xml if we can parse the identifier?

PR #520, opened 2017-08-18
Using a fixed datetime to avoid test failures

narendravardi:
freezegun library added to Travis build and Appveyor

Fixes #507

More details: https://github.com/python-babel/babel/pull/513

PR #537, opened 2017-10-16
Keep tox dependencies in sync

kdeldycke:
Travis config was updated in 1da04fd0af20162fe4e2b503a5fcae8f25267010 to address #520, and adds a dependency on `freezegun`. That dependency was  forgotten in `tox.ini`.


akx:
Thanks! I wonder why this wasn't picked up in CI...

Either way, it would probably be a good idea to migrate testing requirements to `extras_require` stanzas in `setup.py` at some point, so we can just `pip install -e .[dev]` or something.

kdeldycke:
Thanks for the merge! :)

PR #540, opened 2017-10-21
Test and document all supported Python versions

jdufresne:
Adds Python 3.6 and PyPy3 to Travis CI and tox.

Add trove classifiers for all supported versions. Helps package users know if the library is compatible.


.travis.yml
```diff
@@ -35,6 +35,10 @@ matrix:
       python: 3.5
       env:
       - PYTHON_TEST_FLAGS=-bb
+    - os: linux
+      python: 3.6
+    - os: linux
+      python: pypy3
```
akx: This would be a duplicate (see lines 28/29).

jdufresne:
Thanks for catching that. Fixed in latest version.

Issue #545, opened 2017-12-26
bug in format_time bzw in _get_time


Estartu:
I have a datetime.time object with a timezone 
when i call format_time with a different timezone the original timezone is lost in _get_time because in line 172 of dates.py  time = time.replace(tzinfo=tzinfo) is called. So the timezone in the time object is replace with the timezone given in the format_time call without adjusting the time. 

shouldn't this line not be time = time.astimezone(tzinfo) 

Regards
   Estartu

Estartu:
astimezone isn't available for datetime.time so it must be done the hard way.
the problem is the dayligthsavingtime needs a date to do the calculation. So a date is needed to do the correct calculation. Done some coding and come up with the following 

    tempdt =datetime.datetime.now(tz=time.tzinfo)).replace(hour=time.hour, minute=time.minute, second=time.second, microsecond=time.microsecond)
    time = tempdt.astimezone(tz=tzinfo).time().replace(tzinfo=tzinfo)

The last replace is needed because .time() of datetime.datetime dosen't copy the tzinfo from the object

PR #547, opened 2017-12-29
bugfix #545: gettime replaces tzinfo without calculation the new time

Estartu:


akx:
Hi and thanks for the contribution!

Can you explain what this should be doing and add a test case that proves it's doing the right thing (whatever it is)? :)

akx:
Closing due to lack of information.

PR #554, opened 2018-01-26
Add explicit signatures to some date autofunctions

### LINKED ISSUE #542: Misleading doc of locale argument defaults
Looking at the API doc for e.g. [format_datetime](http://babel.pocoo.org/en/latest/api/dates.html#babel.dates.format_datetime), I just noticed that it states the default value for the `locale` parameter is `'en_US_POSIX'`. That is *very* misleading as it actually uses the result of `babel.core.default_locale('LC_TIME')` which may not be anything like `en_US_POSIX` (and just happens to be so on the machine which built the documentation).

This can be fixed by providing an explicit signature as first line of the docstring (which will replace the extracted one) but I don't know that this is the fix the project would expect.

Please advise.

xmo-odoo:
Because `default_locale` is eager, it generates misleading documentation when built (including on the site) as the doc will show the LC_TIME for the machine which built the doc.

An explicit autofunction signature fixes that and provides a more correct view of the signature. 

Alternative fix: have dates.LC_TIME be a custom object with a relevant repr e.g.

```python
class _LC(str):
    def __new__(cls, category):
        category = 'LC_' + category
        v = str.__new__(cls, default_locale(category))
        v._category = category
        return v

    def __repr__(self):
        return "default_locale('%s')" % self._category
LC_TIME = _LC('TIME')
```

### DIFF
--- a/docs/api/dates.rst
+++ b/docs/api/dates.rst
@@ -9,17 +9,17 @@ Python `datetime`, `date` and `time` objects and work with timezones.
 Date and Time Formatting
 ------------------------
 
-.. autofunction:: format_datetime
+.. autofunction:: format_datetime(datetime=None, format='medium', tzinfo=None, locale=default_locale('LC_TIME'))
 
-.. autofunction:: format_date
+.. autofunction:: format_date(date=None, format='medium', locale=default_locale('LC_TIME'))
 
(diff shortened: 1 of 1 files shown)


akx:
Thank you! 🙇

PR #563, opened 2018-03-02
Fixes #402 and fixes #53 add ignore_dir_prefix option original credit @nandoflorestan

### LINKED ISSUE #53: messages not extracted from folders starting with _ or .

### LINKED ISSUE #402: Add `--ignore-dirs` or similar to `extract`

andy-pi:
Babel currently ignores directories prefixed . and _ 
Variously listed as feature / bug since this behavior is not explicit
To exclude directories now use --ignore_dir_prefix option and specify a prefix, e.g. _

based on the following by @nandoflorestan
https://github.com/nandoflorestan/babel/commit/9d2e0ba5cc19ef871a72d4d8a60459e179df1b6b#diff-870e23eed5eae4756d0fe90bb4799c81L896

### DIFF
--- a/babel/messages/extract.py
+++ b/babel/messages/extract.py
@@ -63,7 +63,8 @@ def _strip_comment_tags(comments, tags):
 
 def extract_from_dir(dirname=None, method_map=DEFAULT_MAPPING,
                      options_map=None, keywords=DEFAULT_KEYWORDS,
-                     comment_tags=(), callback=None, strip_comment_tags=False):
+                     comment_tags=(), callback=None, strip_comment_tags=False,
+                     ignore_dir_prefixes=None):
     """Extract messages from any source files found in the given directory.
 
     This function generates tuples of the form ``(filename, lineno, message,
@@ -137,10 +138,11 @@ def extract_from_dir(dirname=None, method_map=DEFAULT_MAPPING,
 
     absname = os.path.abspath(dirname)
     for root, dirnames, filenames in os.walk(absname):
-        dirnames[:] = [
-            subdir for subdir in dirnames
-            if not (subdir.startswith('.') or subdir.startswith('_'))
-        ]
+        for subdir in dirnames:
+            if ignore_dir_prefixes:
+                for prefix in ignore_dir_prefixes:
+                    if subdir.startswith(prefix):
+                        dirnames.remove(subdir)
         dirnames.sort()
         filenames.sort()
         for filename in filenames:
--- a/babel/messages/frontend.py
+++ b/babel/messages/frontend.py
@@ -313,7 +313,7 @@ class extract_messages(Command):
          'files or directories with commas(,)'),  # TODO: Support repetition of this argument
         ('input-dirs=', None,  # TODO (3.x): Remove me.
          'alias for input-paths (does allow files as well as directories).'),
-        ('ignore_dir_prefixes=', None,
+        ('ignore-dir-prefixes=', None,
           'Bypass directories whose name start with PREFIX. You can specify this more than once.'),
     ]
     boolean_options = [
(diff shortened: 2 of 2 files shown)


andy-pi:
@akx I'm a first time contributor, looking to improve my python / open source skills doing a bit of bug fixing. Can you please help with the codecov check issue - I cannot see it showing any error apart from the fact that the test coverage percentage has reduced - obviously, since I added a few lines. I don't know how to add a test for this bugfix or even if it is needed, or if it would increase the percentage enough to pass anyway. Any help appreciated.

babel/messages/extract.py
```diff
@@ -137,10 +138,11 @@ def extract_from_dir(dirname=None, method_map=DEFAULT_MAPPING,
 
     absname = os.path.abspath(dirname)
     for root, dirnames, filenames in os.walk(absname):
-        dirnames[:] = [
-            subdir for subdir in dirnames
-            if not (subdir.startswith('.') or subdir.startswith('_'))
-        ]
+        for subdir in dirnames:
+            if ignore_dir_prefixes:
```
akx: If `ignore_dir_prefixes` defaults to an empty tuple, this check is unnecessary. Also, it should be done outside the `for subdir...` loop anyway, as there's no way the prefix list will change during that iteration.

babel/messages/extract.py
```diff
@@ -137,10 +138,11 @@ def extract_from_dir(dirname=None, method_map=DEFAULT_MAPPING,
...
+                for prefix in ignore_dir_prefixes:
+                    if subdir.startswith(prefix):
+                        dirnames.remove(subdir)
```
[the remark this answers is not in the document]
akx: Also, once you've removed `subdir`, you should break out of the loop here; if multiple `prefix`es match the same `subdir`, `remove` will raise `ValueError: list.remove(x): x not in list`.

babel/messages/frontend.py
```diff
@@ -313,6 +313,8 @@ class extract_messages(Command):
          'files or directories with commas(,)'),  # TODO: Support repetition of this argument
         ('input-dirs=', None,  # TODO (3.x): Remove me.
          'alias for input-paths (does allow files as well as directories).'),
+        ('ignore-dir-prefixes=', None,
+          'Bypass directories whose name start with PREFIX. You can specify this more than once.'),
```
akx: How is specifying this more than once handled? All I see is attempting to split a string by `,`.

andy-pi:
@akx Thanks for reviewing. Yes I would like to develop this, and I'll take on board your suggestions. I'm pretty inexperienced as a python dev - so not really sure how to write tests so I'd need some help with that - would you be able to suggest something?

akx:
@andy-pi You can look at the tests that Babel currently ships :) With py.test, the test framework we use, the gist of it is that tests are functions named `test_<something>` within `test_<something>.py` files and use the `assert` statement to, well, assert things about what the code does. [This](https://github.com/python-babel/babel/blob/efea2dcd8d1aa7b776e9cec171a3b56edb08e3b4/tests/messages/test_frontend.py#L1378-L1384) is a simple test for the message extraction frontend (CLI), for instance.

For this feature I'd like to see, at the very least, tests that

* ensure that `.` and `_` paths get ignored by default (as things work now)
* ensure that an user of the API can pass in a custom filter function which gets correctly used
* ensure that an user of the CLI can pass in multiple exclusion paths (or prefixes, or globs, however you end up implementing that) and they work

You may want to refactor the current code a little for testability; for instance, the `for ... os.walk` loop that this code concerns could be wholesale extracted into a new small method that can then be unit tested.

HTH!

andy-pi:
@akx ok thanks. I will look at it and get to work!

PR #564, opened 2018-03-07
Skip empty message when writing mo file

cedk:
Instead of using the msgid, this allows to have gettext fallback to
retrieve the translation. Otherwise gettext will anyway return msgid.


akx:
Is this the behavior of the GNU gettext .mo utilities too?

I'd like to see a test added, anyway. :)

cedk:
I added a test.
Here is the place where msgfmt from GNU Gettext does the same:  https://github.com/autotools-mirror/gettext/blob/master/gettext-tools/src/msgfmt.c#L1296

Issue #569, opened 2018-04-03
Discussion: Backward compatibility apocalypse


akx:
Hey all,

So as you may have noticed, development on Babel has been a teensy weensy slow as of late, with just minor bugfixes and so on happening. Partially it's due to personal stuff on my side, but there's also something I dread (and if you read the title, you may already be guessing what it'll be).

As you may know, Babel is actually a pretty popular package – according to the PyPI download stats that are available on Google BigQuery, [various versions of Babel have been downloaded over 800,000 times](https://gist.github.com/akx/cf069e214f8654931d3fda2d3ac4cbc4) in March 2018; that means, uh, 26 900 times a day, so 1 100 times per hour, 19 times per minute... so, uh, someone downloads Babel once every 3 seconds. 😅 That means we have quite a bit of responsibility in trying to make sure Babel is stable and usable – or at least that's how it feels to me.

As you may know, the Python community at large isn't that good at pinning their dependency versions, not even loosely. Granted, many Python packages don't follow any specific versioning structure like [Semantic Versioning](https://semver.org/) – not even Babel does at present – so that's  probably part of the reason; secondly, I suppose it's just not been traditional to do so in the Python ecosystem.

There's another [dataset on BigQuery](https://medium.com/google-cloud/github-on-bigquery-analyze-all-the-code-b3576fd2b150) that I [used to dig deeper](https://github.com/akx/babel-usage) into Babel's unpinning situation, by analyzing `requirements.txt` and `setup.py` files that seem to be referring to Babel. I'm sure I missed a lot of repos and my methodology is flawed, but the bottom line is this: there are [tons](https://gist.github.com/akx/69597e59bdd61b3d74924bc480b2975f) of packages that depend on "practically any version of Babel since 2013, I don't care" (`Babel>=1.3`) or "just give me that Babel juice" (`Babel`), or whatever. There are lots of fairly prominent projects in that list (unless they're false positives), such as [Odoo](https://github.com/odoo/odoo/blob/f1a884e046dda1593fc4baeeeb9b27bd668936db/setup.py#L137), [Invenio](https://github.com/inveniosoftware/invenio-assets/blob/e0f48743982d16cffa312fa9fe392c62db906d89/setup.py#L41), [WTForms](https://github.com/wtforms/wtforms/blob/master/setup.py#L43), [Sentry](https://github.com/getsentry/sentry/blob/master/requirements-dev.txt#L2), some Openshift components...

By this point, you're probably seeing where I'm going with this.

[...]

Issue #550, opened 2018-01-15
Release 2.5.2 is not backward compatible


chessbr:
The last release includes this commit https://github.com/python-babel/babel/commit/cf03578b117f92f0af23ae07a65c3e2eada5c30c which changes the NumberPattern.apply signature from

```py
def apply(self, value, locale, currency=None, force_frac=None):
```
to
```py
def apply(self, value, locale, currency=None, currency_digits=True):
```
chopping out the `force_frac` param.

Is this project following semver? Our build started to break on version 2.x. If this is not the case I can just fix the version on our project for now until we update the project.

Thanks

chessbr:
No problem @akx !
Thank you for the support!

akx:
Version 2.5.3, which is v2.5.1 + 7cedb84, 29da2d2, and edfb518, has been pushed to PyPI.

Once again, sorry about the screw-up! 😥

PR #538, opened 2017-10-16
Decimal refactor

kdeldycke:
This PR is refactoring the decimal handling code and:
* Adds extensive unittest of decimal-related rendering methods for percents, scientific notation and currencies.
* Deprecates `format_number()` in favor of `format_decimal()`.
* Fixes the way scientific numbers are rendered: patterns are now strictly respected, especially regarding the number of leading digits.

This PR is a preliminary work to prepare for #494, which adds more control over decimal quantization. The majority of these changes have already been extensively discussed in #494.

PR #494, opened 2017-04-20
Decimal quantization

kdeldycke:
This PR add a `decimal_quantization` parameter to `format_decimal()`, `format_currency()`, `format_percent()` and `format_scientific()`, to bypass the forced quantization on the fractional part. This PR definitely address #90 and is a continuation of #410.

This PR has been extensively reviewed and is waiting for #538 to be merged, so I can rebase that one on top of it.

PR #577, opened 2018-05-07
Restore force_frac to NumberPattern.apply() (as deprecated)

akx:
This retains backwards compatibility (at least as far as the function's
signature is concerned) with users that are using NumberPattern.apply()
directly.

See: https://github.com/python-babel/babel/issues/550
See: https://github.com/python-babel/babel/issues/569

Also refs #494 and #538.

### DIFF
--- a/babel/numbers.py
+++ b/babel/numbers.py
@@ -767,12 +767,38 @@ class NumberPattern(object):
         return value, exp, exp_sign
 
     def apply(
-            self, value, locale, currency=None, currency_digits=True,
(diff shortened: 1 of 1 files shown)


akx:
@kdeldycke Can I ask you to look over this real quick, as you're the decimal architect? :)

kdeldycke:
This PR is indeed good enough to temporarily restore backward-compatibility while we wait for removing that in a future release.

Let's merge! 👍

PR #585, opened 2018-06-04
numbers: implement currency formatting with long display names.

### LINKED ISSUE #578: Implement currency formatting with long display names.
Namely, things like:

    "123,456.78 US dollars"

(en locale)

or

    "dólares estadounidenses 123,456.78"

(es-GT locale)

The algorithm for doing this is described here: https://www.unicode.org/reports/tr35/tr35-numbers.html#Currencies and CLDR contains the data.

I'd be happy to attempt to implement this myself. Two issues I'm aware of:

1) What should the API be? An extra parameter to `format_currency`? A new `format_type` (this risks clashing with other format types already defined)?

2) It looks like this would require data not currently available in the pickled locale data. Specifically it requires the `unitPattern` data. I'm not sure how to go about adding this - apart from the `import_cldr.py` script, is there a specific process to follow?

For comparison, browsers currently implement this as part of [Intl.NumberFormat](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/NumberFormat) using options `{currencyDisplay: "name"}`. In a browser console you can do:

    > Intl.NumberFormat("en", {currencyDisplay:"name", style:"currency", currency:"USD"}).format(123456.78)
    "123,456.78 US dollars"

spookylukey:
Fixes #578

### DIFF
--- a/babel/numbers.py
+++ b/babel/numbers.py
@@ -442,6 +442,17 @@ def format_currency(
         ...
     UnknownCurrencyFormatError: "'unknown' is not a known currency format type"
 
+    You can also pass format_type='name' to use long display names. The order of
+    the number and currency name, along with the correct localized plural form
+    of the currency name, is chosen according to locale:
+
+    >>> format_currency(1, 'USD', locale='en_US', format_type='name')
+    u'1.00 US dollar'
+    >>> format_currency(1099.98, 'USD', locale='en_US', format_type='name')
+    u'1,099.98 US dollars'
+    >>> format_currency(1099.98, 'USD', locale='ee', format_type='name')
+    u'us ga dollar 1,099.98'
+
     By default the locale is allowed to truncate and round a high-precision
     number by forcing its format pattern onto the decimal part. You can bypass
     this behavior with the `decimal_quantization` parameter:
@@ -459,7 +470,12 @@ def format_currency(
     :param format_type: the currency format type to use
     :param decimal_quantization: Truncate and round high-precision numbers to
                                  the format pattern. Defaults to `True`.
+
     """
+    if format_type == 'name':
+        return _format_currency_long_name(number, currency, format=format,
+                                          locale=locale, currency_digits=currency_digits,
+                                          decimal_quantization=decimal_quantization)
     locale = Locale.parse(locale)
     if format:
         pattern = parse_pattern(format)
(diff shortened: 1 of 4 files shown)


spookylukey:
Some notes:

Step 1 of the [algorithm](https://www.unicode.org/reports/tr35/tr35-numbers.html#Currencies) involves checking for an exact match against number 0 or 1. However, none of the locale data actually has data defined for this. Since I would have been producing untestable code, I thought it best not to implement this. It could be added as a refinement when the need comes up.

Step 5 refers to applying `currencyDecimal` where appropriate. I have omitted this part, because babel doesn't currently support it all, and that affects other currency formatting, not just this new feature, and therefore ought to be addressed as its own issue. (It also would require a bit of re-plumbing to implement correctly, I think, and therefore would complicate this patch).

It would appear that there is just one affected locale which actually defines this currencyDecimal value - `fr_CH`, which defines it as `.` 

babel doesn't use it though:

    >>> print(format_currency(12345.678, 'USD', locale='fr_CH'))
    12 345,68 $US

Contrast this with `Intl.NumberFormat` (tested in Firefox and Chrome), which correctly has a `.` for the decimal separator instead of the `,`:

    > Intl.NumberFormat("fr-CH", {currencyDisplay:"symbol", style:"currency", currency:"USD"}).format(12345.678)
    "12 345.68 $US "

Whether you want to address this odd corner case is a matter for another issue I think.

akx:
Based on a cursory review, this looks great! Clean and simple :)
I'll give it a more thorough review this week, but it's looking good.

Mind making an issue out of that `currencyDecimal` thing, so it's not lost here in the comments?

scripts/import_cldr.py
```diff
@@ -903,6 +904,14 @@ def parse_currency_formats(data, tree):
                         currency_formats[type] = numbers.parse_pattern(pattern)
 
 
+def parse_currency_unit_patterns(data, tree):
+    currency_unit_patterns = data.setdefault('currency_unit_patterns', {})
+    for unit_pattern_elem in tree.findall('.//currencyFormats/unitPattern'):
```
akx: This should ignore non-`latn` currency formats at present, just in case there are both `latn` and non-`latn` `currencyFormats` elements within a single locale.

You can (rebase and) use the `_should_skip_number_elem()` function introduced in https://github.com/python-babel/babel/pull/579 to do that.

babel/numbers.py
```diff
@@ -475,6 +491,52 @@ def format_currency(
...
+    # locale data. So there is no point implementing that.
+    # Step 2.
+    if isinstance(number, string_types):
+        plural_category = locale.plural_form(float(number))
+    else:
+        plural_category = locale.plural_form(number)
+
+    # Step 3.
+    try:
+        unit_pattern = locale._data['currency_unit_patterns'][plural_category]
```
akx: I mildly like pulling things out of `locale._data` from outside the `locale` object, but we haven't been that strict about it before either.

scripts/import_cldr.py
spookylukey: Fixed in 4b5097f6064dda281ce7a1c5ef74e047cfdc558c

babel/numbers.py
spookylukey: I had a partial go at fixing this in 4c7e9b645cfec72672f836e04d8587ca66ba98c3 - it is at least more consistent now, and the resulting code is much flatter as well.

spookylukey:
Copying my message from the 2nd commit in case it gets a bit lost:

4c7e9b6 Simplify format_currency code by pulling out/using helpers.

In detail:

1. Use the already existing `get_currency_name` function which does  the plural form logic already.

2. Create a similar `get_currency_unit_pattern` function. This function could be useful if someone wanted to write something very similar to the `_format_currency_long_name` functionality but with some different customizations, so it is now publicly documented.

3. Simplify the `_format_currency_long_name` function - it is now much flatter and shorter.

4. Add more tests to ensure certain cases are really covered. (e.g. different unit patterns depending on the count)

An upshot of the changes is that we have reduced (and made more consistent) the number of places where we need to peek into `Locale._data` - `get_currency_name` and `get_currency_unit_pattern` are the only places that babel.numbers does this.

----  

In fact I already have a use case for having `get_currency_unit_pattern` as public API, so this seems better all round.

akx:
Thank you, @spookylukey! 😻

spookylukey:
Thanks for merging, and maintaining babel - it is one of the those projects that is often a thankless task when it comes to maintaining it, rarely making any headlines, but a dependency of a huge number of projects!

spookylukey:
@akx - do you have plans to do a release soon? I'm not pressuring you to do so, it would just be nice to know - there is another library I'm preparing a patch for that could be improved if I could use the feature you merged in this PR (and be able to refer to the specific babel version it was added in for dependencies). Thanks!

PR #588, opened 2018-06-10
Use https URLs throughout project where available

jdufresne:
Updated docs and code comments.


akx:
Looking good when the conflicts are fixed.

jdufresne:
Rebased to latest master. Thanks.

PR #590, opened 2018-06-16
Add a strict mode to `parse_decimal()`

### LINKED ISSUE #589: The `parse_decimal` function accepts irregular input
Examples:

```python
>>> parse_decimal('1,,000', locale='en_US')
1000
>>> parse_decimal('25.00', locale='de')
2500
```

Inputting `25.00` instead of `25,00` is an easy mistake to make, and it can be confusing that the software reads it as `2500` instead of `25`.

Maybe `parse_decimal` should be more strict and reject numbers formatted in a weird way?

This would require checking that the separators are positioned correctly before [replacing them](https://github.com/python-babel/babel/blob/v2.6.0/babel/numbers.py#L604-L605).

Changaco:
This branch adds a `strict` flag to the `parse_decimal` function. When set to `True` an exception is raised if the input number is formatted in a weird way. This is determined by comparing the string to the result of `format_decimal()`.

Closes #589.

### DIFF
--- a/babel/numbers.py
+++ b/babel/numbers.py
@@ -682,7 +682,7 @@ def parse_decimal(string, locale=LC_NUMERIC, strict=False):
     >>> parse_decimal('30.00', locale='de', strict=True)
     Traceback (most recent call last):
         ...
-    NumberFormatError: '30.00' is not a properly formatted decimal number
+    NumberFormatError: '30.00' is not a properly formatted decimal number. Did you mean '3.000'? Or maybe '30,00'?
 
     :param string: the string to parse
     :param locale: the `Locale` object or locale identifier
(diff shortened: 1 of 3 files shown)


babel/numbers.py
```diff
@@ -676,17 +676,36 @@ def parse_decimal(string, locale=LC_NUMERIC):
...
-                               .replace(get_decimal_symbol(locale), '.'))
+        r = decimal.Decimal(string.replace(group_symbol, '')
+                                  .replace(decimal_symbol, '.'))
     except decimal.InvalidOperation:
         raise NumberFormatError('%r is not a valid decimal number' % string)
+    if strict and group_symbol in string:
+        proper = format_decimal(r, locale=locale, decimal_quantization=False)
+        if string != proper and string.rstrip('0') != (proper + decimal_symbol):
+            raise NumberFormatError(
+                "%r is not a properly formatted decimal number" % string
```
akx: It would be super nice to have the expected value in the error here.

Maybe even

```python
class MisparsedNumberError(NumberFormatError):
  def __init__(self, message, original, parsed, expected):
    super(MisparsedNumberError, self).__init__(message)
    self.original = original
    self.parsed = parsed
    self.expected = expected

# ...

raise MisparsedNumberError(
  "%r is not properly formatted; parsed as %r, expected %r" % (string, r, proper),
  original=string,
  parsed=r,
  expected=proper,
)

```

so users could have more visibility into what went wrong.

babel/numbers.py
```diff
@@ -676,17 +676,36 @@ def parse_decimal(string, locale=LC_NUMERIC):
...
     :raise NumberFormatError: if the string can not be converted to a
                               decimal number
     """
     locale = Locale.parse(locale)
+    group_symbol = get_group_symbol(locale)
+    decimal_symbol = get_decimal_symbol(locale)
     try:
-        return decimal.Decimal(string.replace(get_group_symbol(locale), '')
-                               .replace(get_decimal_symbol(locale), '.'))
+        r = decimal.Decimal(string.replace(group_symbol, '')
```
akx: Could the name `r` be something more descriptive?

Changaco:
Ready for second review.

PR #592, opened 2018-06-18
Bumped version requirement on pytz

chrisbrake:
Here you use the utc attribute on the _pytz object: https://github.com/python-babel/babel/blob/23ca4bf2b187748e5e8372c6dae541fbcc5fbe5d/babel/util.py#L304

Which was added in this commit https://github.com/stub42/pytz/commit/ae82fbc71b374a5a5b32bf3bbb38ebb97af6e715

Corrected the version requirement to match the version where that attribute was added.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -61,7 +61,7 @@ setup(
         # This version identifier is currently necessary as
         # pytz otherwise does not install on pip 1.4 or
         # higher.
-        'pytz>=0a',
+        'pytz>=2015.7',
     ],
 
     cmdclass={'import_cldr': import_cldr},


chrisbrake:
The failing travis build failed with the error:

`FileNotFoundError: [Errno 2] No such file or directory: '/home/travis/build/python-babel/babel/cldr/core-33.0/common/supplemental/supplementalData.xml'`

I do not believe that is related to my change.

akx:
The Travis problem seems like a transient blip. I restarted the affected build, it should go through.

chrisbrake:
Thanks for restarting the build for me

PR #597, opened 2018-07-19
Small fixes to avoid stack traces due to badly formatted .po file

bryntruscott:
A couple of very small fixes to avoid stack traces due to errors in the .po file.

1. Line 186
If a keyword - e.g. 'msgid' appears on a line by itself, the 'len(keyword)' call used to raise an IndexError exception rather than reporting the error

2. Line 293
If the offending line in the .po file contains unicode characters - e.g. accents - the 'format()' call was raising a UnicodeEncodeError rather than reporting the location of the error.


akx:
Thank you!

PR #599, opened 2018-08-06
don't repeat suggestions in parse_decimal strict

### LINKED ISSUE #596: Repeated suggestions in parse_decimal strict mode
When you enter 0.00 in languages which use commas for delimiters using the strict mode of `parse_decimal`, you get two times the same suggestion.

<pre><samp>babel.numbers.NumberFormatError: '0.00' is not a properly formatted decimal number. Did you mean '0'? Or maybe '0'?</samp></pre>

Related to [issue 1177](https://github.com/liberapay/liberapay.com/issues/1177) of Liberapay.

fuzzmz:
Don't repeat suggestions for `0.00` in languages
which use commas as delimiters combined with strict
mode.

fixes #596

### DIFF
--- a/babel/numbers.py
+++ b/babel/numbers.py
@@ -689,6 +689,11 @@ def parse_decimal(string, locale=LC_NUMERIC, strict=False):
         ...
     NumberFormatError: '30.00' is not a properly formatted decimal number. Did you mean '3.000'? Or maybe '30,00'?
 
+    >>> parse_decimal('0.00', locale='de', strict=True)
+    Traceback (most recent call last):
+        ...
+    NumberFormatError: '0.00' is not a properly formatted decimal number. Did you mean '0'?
+
     :param string: the string to parse
     :param locale: the `Locale` object or locale identifier
     :param strict: controls whether numbers formatted in a weird way are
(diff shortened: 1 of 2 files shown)


babel/numbers.py
```diff
@@ -717,10 +722,16 @@ def parse_decimal(string, locale=LC_NUMERIC, strict=False):
                 ), suggestions=[proper])
             else:
                 proper_alt = format_decimal(parsed_alt, locale=locale, decimal_quantization=False)
-                raise NumberFormatError((
-                    "%r is not a properly formatted decimal number. Did you mean %r? Or maybe %r?" %
-                    (string, proper, proper_alt)
-                ), suggestions=[proper, proper_alt])
+                if proper_alt == '0':
```
akx: Hmm, could we just look at whether `proper_alt != proper`? That'd be bullet-proof...

fuzzmz: You're right, that's a way better check (well, doing `proper_alt == proper` as not to change the order of NumberFormatError's and make the diff larger).

akx:
Thank you (and sorry for the delay in merging)!

PR #600, opened 2018-08-15
add support to util.pathmatch for matching the start of a string

briancappello:
This PR adds support to `util.pathmatch` for forcing the match of the start of a string. This is useful for example if you want to place a `babel.cfg` file in a subdirectory and only extract messages from under that directory. The added (equivalent) syntax looks like so:

```ini
[python: ./**.py]
[python: ^**.py]
```


akx:
Looks good. Thank you!

PR #609, opened 2018-10-30
Using ABCs from collections instead of collections.abc is deprecated.

JulienPalard:
And it will stop working in Python 3.8 according to the warning.


akx:
Thank you! Very elegant patch.

tirkarthi:
It will be good to have a release since the upstream CPython PR is close to getting merged : https://github.com/python/cpython/pull/10596

Thanks!

PR #615, opened 2018-12-21
localedata.py: Do disk lookup once

lexh:
After this change, `localedata.locale_identifiers` will only read the available locale identifiers from disk once. Since the locale data files are distributed with the lib, there shouldn't be a need to scan the disk every time this is called.

I noticed a hot spot in this method when profiling a service that uses this library. Specifically I noticed that `babel.Locale.parse` performs very poorly when supplied bogus locales. I ran a small benchmark to demonstrate this:

```
import timeit

import babel


def wrapped():
    try:
        babel.Locale.parse("en_ZH")
    except babel.UnknownLocaleError:
        pass


print(timeit.timeit(wrapped, number=1000))
```

```
Before: 3.91633701324
After: 0.0591921806335
```

### DIFF
--- a/babel/localedata.py
+++ b/babel/localedata.py
@@ -24,7 +24,8 @@ _dirname = os.path.join(os.path.dirname(__file__), 'locale-data')
 _locale_identifiers = set(stem for stem, extension in [
     os.path.splitext(filename) for filename in os.listdir(_dirname)
 ] if extension == '.dat' and stem != 'root')
-_locale_identifiers_lower = set(s.lower() for s in _locale_identifiers)
+_locale_identifiers_normalization_map = {
+    s.lower(): s for s in _locale_identifiers}
 
 
 def normalize_locale(name):
@@ -35,9 +36,10 @@ def normalize_locale(name):
     """
     if not name or not isinstance(name, string_types):
         return None
-    name = name.strip().lower()
-    if name in _locale_identifiers_lower:
-        return name
+    try:
+        return _locale_identifiers_normalization_map[name.strip().lower()]
+    except KeyError:
+        return None
 
 
 def exists(name):
@@ -49,10 +51,11 @@ def exists(name):
     """
     if not name or not isinstance(name, string_types):
         return False
+
     if name in _cache:
         return True
-    file_found = os.path.exists(os.path.join(_dirname, '%s.dat' % name))
-    return True if file_found else bool(normalize_locale(name))
+
+    return normalize_locale(name) is not None
 
 
 def locale_identifiers():
@@ -89,11 +92,12 @@ def load(name, merge_inherited=True):
     :param merge_inherited: whether the inherited data should be merged into
                             the data of the requested locale
     :raise `IOError`: if no locale data file is found for the given locale
-                      identifer, or one of the locales it inherits from
+                      identifier, or one of the locales it inherits from
     """
     _cache_lock.acquire()
+    cache_key = 'root' if name == 'root' else normalize_locale(name)
     try:
-        data = _cache.get(name)
+        data = _cache.get(cache_key)
         if not data:
             # Load inherited data
             if name == 'root' or not merge_inherited:
@@ -114,7 +118,8 @@ def load(name, merge_inherited=True):
                     merge(data, pickle.load(fileobj))
                 else:
                     data = pickle.load(fileobj)
-            _cache[name] = data
+            _cache[cache_key] = data
+
         return data
     finally:
         _cache_lock.release()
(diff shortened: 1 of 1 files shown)


babel/localedata.py
```diff
@@ -47,10 +51,11 @@ def exists(name):
     """
     if not name or not isinstance(name, string_types):
         return False
+
     if name in _cache:
         return True
-    file_found = os.path.exists(os.path.join(_dirname, '%s.dat' % name))
-    return True if file_found else bool(normalize_locale(name))
```
lexh: The first condition is no longer needed, if the file exists on disk then `name` must be in `_locale_identifiers` and the call to `normalize_locale` will return a `str`.

babel/localedata.py
```diff
@@ -89,9 +92,10 @@ def load(name, merge_inherited=True):
     :param merge_inherited: whether the inherited data should be merged into
                             the data of the requested locale
     :raise `IOError`: if no locale data file is found for the given locale
-                      identifer, or one of the locales it inherits from
+                      identifier, or one of the locales it inherits from
     """
     _cache_lock.acquire()
+    name = 'root' if name == 'root' else normalize_locale(name)
```
lexh: The relationship between `Locale.__init__`, `Locale._data`, `localedata.exists` and `localedata.load` seemed off.

1) In `Locale.__init__`, we check if the passed in identifier is valid by calling `localedata.exists`, which normalizes the identifier to make the check (both before and after this PR). https://github.com/python-babel/babel/blob/b7dda2a0de98344426359087a3d3c239b74de548/babel/core.py#L167 If the check succeeds we return a `Locale` instance.

2) When the lazy property `Locale._data` is accessed, there is a call to `localedata.load`, the identifier passed in is not normalized. https://github.com/python-babel/babel/blob/b7dda2a0de98344426359087a3d3c239b74de548/babel/core.py#L364

This would allow the following unexpected behavior:
```
Python 2.7.6 (default, Nov 13 2018, 12:45:42)
[GCC 4.8.4] on linux2
>>> import babel
>>> a = babel.Locale('en')  # lowercase
>>> b = babel.Locale('EN')  # uppercase
>>> len(a._data). # just to force the lazy load
39
>>> len(b._data). # whoops!
  File "/.../venv/local/lib/python2.7/site-packages/babel/core.py", line 364, in _data
  File "/.../venv/local/lib/python2.7/site-packages/babel/localedata.py", line 113, in load
    with open(filename, 'rb') as fileobj:
IOError: [Errno 2] No such file or directory: '/.../venv/local/lib/python2.7/site-packages/babel/locale-data/EN.dat'
```

On case-insensitive filesystems, it's not as dramatic but still surprising.
```
Python 2.7.15 (default, Oct  2 2018, 11:47:18)
[GCC 4.2.1 Compatible Apple LLVM 10.0.0 (clang-1000.11.45.2)] on darwin
>>> import babel
>>> a = babel.Locale('en')  # lowercase
>>> b = babel.Locale('EN')  # uppercase
>>> len(a._data)  # just to force the lazy load
40
>>> len(b._data)
40
>>> babel.localedata._cache.keys()  # whoops! stored the same data twice
['en', 'root', 'EN']
```

This PR attempts to remediate by normalizing the identifier here.

babel/localedata.py
```diff
@@ -32,10 +36,10 @@ def normalize_locale(name):
     """
     if not name or not isinstance(name, string_types):
         return None
-    name = name.strip().lower()
-    for locale_id in chain.from_iterable([_cache, locale_identifiers()]):
-        if name == locale_id.lower():
-            return locale_id
+    try:
```
akx: This could just be rephrased as 

```python
return _locale_identifiers_normalization_map.get(name.strip().lower())
```

babel/localedata.py
```diff
@@ -14,14 +14,18 @@
...
 import threading
-from itertools import chain
 
 from babel._compat import pickle, string_types, abc
 
 
 _cache = {}
 _cache_lock = threading.RLock()
 _dirname = os.path.join(os.path.dirname(__file__), 'locale-data')
+_locale_identifiers = set(stem for stem, extension in [
```
akx: I'm a little hesitant about this happening at import time rather than the first time we need the data...

babel/localedata.py
akx: Yeah, this can happen... People usually shouldn't be using the `babel.Locale()` ctor by hand anyway, but `babel.Locale.parse()` instead.

We're acutely aware that's not the best API though. :)

babel/localedata.py
lexh: good point.

lexh:
moving to issue, might try again later

PR #617, opened 2018-12-30
Replace odict with Python's collection.OrderedDict

jdufresne:
The `odict` class duplicates `collection.OrderedDict` from Python's standard lib. Simplify the code by using builtin Python features.

https://docs.python.org/3/library/collections.html#collections.OrderedDict

### DIFF
--- a/babel/util.py
+++ b/babel/util.py
@@ -10,6 +10,7 @@
 """
 
 import codecs
+import collections
 from datetime import timedelta, tzinfo
 import os
 import re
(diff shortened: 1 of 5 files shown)


babel/util.py
```diff
@@ -220,79 +220,6 @@ def wraptext(text, width=70, initial_indent='', subsequent_indent=''):
     return wrapper.wrap(text)
 
 
-class odict(dict):
```
akx: Since the old `odict` is basically `collections.OrderedDict`, we could maybe do something like

```python
# TODO (Babel 3.x): Remove this re-export
odict = collections.OrderedDict
```

in case someone is still using this for some reason.

The other changes are fine, though - no need to use the `odict` alias internally.

jdufresne:
Thanks for the review. I've added the compatibility shim in the latest revision :slightly_smiling_face:.

akx:
@jdufresne Thank you for doing the work 🙌

PR #621, opened 2019-01-07
fixes #619 wrong weeknumber for 31.12.2018

### LINKED ISSUE #619: Weeknumber wrong for 2018-12-31
As mentioned in the docs babel.dates.format_date uses the ISO-Calendar when passing "YYYY" as format string.

Evaluating gives
```
babel.dates.format_date(datetime.datetime(2018,12,31
                                         ), format="w YYYY", locale='en_US')
>> u'53 2018'
```
The correct result should be 1 2019 by iso definitions.

Thanks for your reply

BT-sschmid:
The weeknumber was calculated to 53, but by definition the value
must compute to 1.

the fix will compute the weeknumber by using date.isocalendar if
locale.first_week_day == 0.

Also the computation of the year format 'YYYY' is replaced by isocalendar.

Tests added.

Fixes https://github.com/python-babel/babel/issues/619

### DIFF
--- a/babel/dates.py
+++ b/babel/dates.py
@@ -1506,8 +1506,13 @@ class DateTimeFormat(object):
 
         if 7 - first_day >= self.locale.min_week_days:
             week_number += 1
-
+        
         if self.locale.first_week_day == 0:
+            # Correct the weeknumber in case of iso-calendar usage (first_week_day=0). 
+            # If the weeknumber exceeds the maximum number of weeks for the given year
+            # we must count from zero.For example the above calculation gives week 53 
+            # for 2018-12-31. By iso-calender definition 2018 has a max of 52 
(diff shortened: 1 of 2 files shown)


babel/dates.py
```diff
@@ -1505,8 +1503,15 @@ def get_week_number(self, day_of_period, day_of_week=None):
         if first_day < 0:
             first_day += 7
         week_number = (day_of_period + first_day - 1) // 7
+
         if 7 - first_day >= self.locale.min_week_days:
             week_number += 1
+
+        if self.locale.first_week_day == 0:
```
akx: I'd love to see a comment explaining the rationale here, since the code (like all date code) is a little obtuse anyway :)

Issue #625, opened 2019-01-24
Use CLDR 34


mardy:
Please update to CLDR 34. This would probably fix issues #83 and #451.

akx:
Fixed in #626.

PR #626, opened 2019-01-24
Download CLDR 34.0

mardy:
Fixes issue #625

### DIFF
--- a/babel/numbers.py
+++ b/babel/numbers.py
@@ -448,7 +448,7 @@ def format_currency(
     >>> format_currency(1099.98, 'JPY', locale='en_US')
     u'\\xa51,100'
     >>> format_currency(1099.98, 'COP', u'#,##0.00', locale='es_ES')
-    u'1.100'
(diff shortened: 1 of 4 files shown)


babel/numbers.py
```diff
@@ -448,7 +448,7 @@ def format_currency(
     >>> format_currency(1099.98, 'JPY', locale='en_US')
     u'\\xa51,100'
     >>> format_currency(1099.98, 'COP', u'#,##0.00', locale='es_ES')
-    u'1.100'
+    u'1.099,98'
```
akx: This looked a little suspect at first, but seems to be legit:

* https://www.unicode.org/cldr/charts/34/delta/supplemental-data.html#COP
* https://unicode.org/cldr/trac/ticket/10788

[COP (the Colombian peso)](https://en.wikipedia.org/wiki/Colombian_peso) doesn't have cash money smaller than 1 peso, but for accounting, etc. purposes, one peso is still 100 centimos.

PR #633, opened 2019-03-01
frontend: Add  omit-header to update_catalog

cedk:
It is the same option as in extract_message but for when updating the
catalog.


akx:
Thank you!

PR #635, opened 2019-03-14
Fix conftest.py compatibility with pytest 4.3

hroncok:
While pytest in tox.ini is explicitly set to be 3.3.2, in Fedora 31 we are about to update to a newer version of pytest. In order to be able to test babel, we need pytest 4.3 support.

This adds support for pytest 4.3 without breaking support for 3.3.2.

Switching to 4.3 in tox.ini is left at the discretion of the maintainers.

### DIFF
--- a/.ci/appveyor.yml
+++ b/.ci/appveyor.yml
@@ -27,7 +27,7 @@ install:
   - "python --version"
   - "python -c \"import struct; print(struct.calcsize('P') * 8)\""
   # Build data files
-  - "pip install --upgrade pytest==3.3.2 pytest-cov==2.5.1 codecov freezegun==0.3.11"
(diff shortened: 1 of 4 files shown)


akx:
@hroncok Thank you! We could by all means upgrade to pytest 4, so if you feel like it, please add the other required changes too. :)

hroncok:
Should I pin pytest at 4.3.1?

akx:
> Should I pin pytest at 4.3.1?

That's fine. Thank you!

akx:
Oh, btw, the same changes will need to be done in the travis and coveralls configs too (due to Legacy Reasons). Would you mind doing that too, @hroncok? :)

hroncok:
sure, later today

hroncok:
amended

akx:
Thank you! 🎉

PR #638, opened 2019-04-12
Update for version portuguese/Brazil

juliocesark:


akx:
Sorry, but I can't accept a PR that removes the documentation index page and replaces it with a Portuguese version that has broken links.

Issue #179, opened 2015-08-02
Add Number Spelling


sils1297:
this bug is to track the commits of https://github.com/mitsuhiko/babel/pull/114

sils1297:
This would need to be done via CLDR. https://github.com/python-babel/babel/pull/114 contains a native implementation that might be helpful in order to understand spelling a bit.

PR #114, opened 2014-10-12
Number Spelling

blagasz:
Reinstantiated pull request to make it a clear changeset.

The spelling framework tries to be as generic as possible, but still structured upon my original non-generic implementation of the Hungarian spelling. The Hungarian language in general and the spelling in particular is very complicated, so I hope that the generic helper structure that is suitable for an easy implementation of the Hungarian spelling would be also suitable for all the other languages.

PR #660, opened 2019-07-26
Number spelling based on the CLDR's RBNF rules

blagasz:
After some years, I finally did what I kind of promised.

It is still a long way to go, including RBNF updates to
CLDR, but hopefully this implementation is a step
towards a good direction.

Any suggestions are welcome!

It is a pure Python engine for parsing RBNF rules.
The rules are incomplete in many cases, fractional
number spelling is hardly supported.

Based on an earlier discussion: https://github.com/python-babel/babel/pull/114 and referenced in https://github.com/python-babel/babel/issues/179

### DIFF
--- a/babel/numbers.py
+++ b/babel/numbers.py
@@ -24,6 +24,7 @@ import warnings
 
 from babel.core import default_locale, Locale, get_global
 from babel._compat import decimal, string_types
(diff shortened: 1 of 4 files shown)


akx:
Thank you! Let's continue in #682! :)

Issue #665, opened 2019-10-07
Can not initialize a catalog from a template with an empty string value for "Language" header


sinoroc:
[_GNU gettext_'s `xgettext`](https://www.gnu.org/software/gettext/manual/html_node/xgettext-Invocation.html) and [_lingua_'s `pot-create` ](https://pypi.org/project/lingua/) both create _PO_-template files containing a `Language` header with an empty string as its value (`"Language: \n"`).

_Babel_ does not seem to be able to handle this empty header and produces the error and stacktrace below.

Removing the `Language` header from the template file allows _Babel_ to initialize the catalogs without issue.

- Babel 2.7.0
- lingua 4.13
- xgettext 0.19.8.1
- Python 3.6.8

```
python3 setup.py init_catalog --domain project --input src/project/locale/project.pot --output-file src/project/locale/en/LC_MESSAGES/project.po --locale en
running init_catalog
creating catalog src/project/locale/en/LC_MESSAGES/project.po based on src/project/locale/project.pot
Traceback (most recent call last):
  File "setup.py", line 35, in <module>
    _do_setup()
  File "setup.py", line 30, in _do_setup
    version=version,
  File "/home/sinoroc/workspace/project/.tox/develop/lib/python3.6/site-packages/setuptools/__init__.py", line 145, in setup
    return distutils.core.setup(**attrs)
  File "/usr/lib/python3.6/distutils/core.py", line 148, in setup
    dist.run_commands()
  File "/usr/lib/python3.6/distutils/dist.py", line 955, in run_commands
    self.run_command(cmd)
  File "/usr/lib/python3.6/distutils/dist.py", line 974, in run_command
    cmd_obj.run()
  File "/home/sinoroc/workspace/project/.tox/develop/lib/python3.6/site-packages/babel/messages/frontend.py", line 622, in run
    catalog = read_po(infile, locale=self.locale)
  File "/home/sinoroc/workspace/project/.tox/develop/lib/python3.6/site-packages/babel/messages/pofile.py", line 377, in read_po
    parser.parse(fileobj)
  File "/home/sinoroc/workspace/project/.tox/develop/lib/python3.6/site-packages/babel/messages/pofile.py", line 308, in parse
    self._process_comment(line)
  File "/home/sinoroc/workspace/project/.tox/develop/lib/python3.6/site-packages/babel/messages/pofile.py", line 267, in _process_comment
    self._finish_current_message()
[...]

PR #666, opened 2019-10-07
Ignore 'Language' header in PO-templates

sinoroc:
The tools 'xgettext' from GNU gettext and 'pot-create' from lingua
both produce PO-template files containing a 'Language' header with
an empty string as value.

Babel has issues creating catalogs out of such templates.

It wants to parse the PO-template as a normal PO-catalog using the
locale of the target catalog, because it helps producing the correct
plurals. But it overwrites this locale when parsing the 'Language'
header and then it can not produce a valid catalog.

Ignoring the 'Language' header when it is known that the file being
processed is a template avoids this issue.

GitHub: #665


sinoroc:
Anything I can do to get this moving?

PR #671, opened 2019-10-18
Add install of pytz before import_cldr

### LINKED ISSUE #670: Minor Error in Living on the Edge instructions
The living on the edge instructions at http://babel.pocoo.org/en/latest/installation.html#living-on-the-edge fail because `python setup.py import_cldr` is after activating the venv but depends on `pytz` which isn't installed at that point.

Suggest adding `pip install pytz` after the `activate` and before the `setup` line.

GadgetSteve:
closes #670


akx:
Thank you!

Issue #141, opened 2015-02-10
Allow parsing of symbols to get to currency code.


slachinger:
Babel allows to "speak" any locale but no to understand other locales.

Similar to get_currency_name() and get_currency_symbol() from the currency code, there should be a method allowing to get the currency code from the symbol, e.g. 
```python
>>> from babel.numbers import get_currency_code
>>> get_currency_code('$', locale='en_US')
'USD'
>>> get_currency_code(u'\u20ac', locale='de_DE') 
'EUR'.
```
Even better would be a reverse method to format_currency(), e.g.
```python
>>> from babel.numbers import parse_currency
>>> parse_currency(u'123,45 \u20ac', locale='de_DE')
(123.45, 'EUR')
```
I see that "Locale.parse(locale).currency_symbols" already contains the required mapping for this.

Cheers, Simon

rojaraniiza:
---> 21 from babel.numbers import parse_currency
     22 # from babel.numbers import get_currency_code
ImportError: cannot import name 'parse_currency'
Can you please help me out to fix this?

akx:
@rojaraniiza The issue is there _is_ no such function yet. This is a proposal to implement one.

PR #677, opened 2019-12-01
Add symbol to currency code mapping

ddormer:
During cldr import the `currency_codes` mapping is created on `data` and
internally exposed by the `Locale.currency_codes` property.
The `numbers.get_currency_code` API is made available to convert a
currency symbol into a currency code.

Partially addresses issue #141

### DIFF
--- a/babel/core.py
+++ b/babel/core.py
@@ -531,6 +531,17 @@ class Locale(object):
         return self._data['currency_names']
 
     @property
+    def currency_codes(self):
+        """Mapping of currency symbols to codes.
+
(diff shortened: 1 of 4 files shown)


ericzolf:
I personally use the following workaround in the meantime:

```python
import babel.numbers as babelnum

CURRENCY_MAP = { babelnum.get_currency_symbol(x): x
                 for x in babelnum.list_currencies()
                 if x != babelnum.get_currency_symbol(x) }

def get_currency_code(currency_symbol):
    return CURRENCY_MAP.get(currency_symbol, currency_symbol)
```

PR #682, opened 2019-12-31
Number spelling

akx:
Thank you @blagasz for a great initial effort! I cleaned up things here and there (refactoring, performance issues, even some bugs).

There's also a smoke test to try and verify we get some result for every technically supported locale, but there are some infinite recursion cases involving year formatting I couldn't really figure out. These are marked XFAIL for the time being:

```
XFAIL tests/test_number_spelling.py::test_spelling_smoke[zh_Hant-year]
XFAIL tests/test_number_spelling.py::test_spelling_smoke[zh-year]
XFAIL tests/test_number_spelling.py::test_spelling_smoke[yue_Hans-year]
XFAIL tests/test_number_spelling.py::test_spelling_smoke[ja-year]
XFAIL tests/test_number_spelling.py::test_spelling_smoke[yue-year]
XFAIL tests/test_number_spelling.py::test_spelling_smoke[ak-year]
```

In addition, fractional formatting results are rather strange, but as the original PR's description said, it's still a work in progress.

Closes #660 (supersedes it)

### DIFF
--- a/babel/numbers.py
+++ b/babel/numbers.py
@@ -24,6 +24,7 @@ import warnings
 
 from babel.core import default_locale, Locale, get_global
 from babel._compat import decimal, string_types
+from babel.rbnf import RuleBasedNumberFormat
 
 try:
     # Python 2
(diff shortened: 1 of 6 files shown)


blagasz:
Thank you @akx for your improvements!

I was trying to figure out the recursion but I could not reproduce it in my original code.

I also found an error regarding the default ruleset for the `year=True` flag. How should I commit not to mess up your concept?

nagylzs:
Hello! I have my own implementation for number spelling  (Hungarian only), but I just found out that this may be part of babel. Looks like tests for unsupported python versions are failed only. How can I help to finish the implementation?

blagasz:
Hi, nice to have more people on board :)

I also started with my own implementation (#114) then the folks here convinced me that the CLDR RBNF is the way to go. I think the basic implementation of the RBNF rule engine is working now and @akx cleaned up the code.

I think we need to improve the tests and somehow reproduce the recursion error to address that, basic tests for English and Hungarian already run smoothly.

blagasz:
I did some additional work and created a new PR for that #1232, let's continue there!

Issue #128, opened 2014-12-11
API Docs: Catalog -> Message mapping not documented


rbu:
The `Catalog` API docs mention nowhere how one would get a list of all messages. Apparently, `__iter__` is supported. Is that part of the official API? It would be nice if this were mentioned somehwere.

yash14123:
On it .

lladhibhutall:
Is this issue still open? I would like to work on it.

PR #686, opened 2020-01-13
catalog.rst: Add __iter__ to Catalog documentation

stfab:
The declaration of \_\_iter__ under the special-members
directive makes it visible in the documentation.
The docstring describing \_\_iter__ already exists.

Closes https://github.com/python-babel/babel/issues/128


akx:
Thank you!

PR #691, opened 2020-01-27
Fix unicode printing error on Python 2 without TTY.

nh2:
Until now, on Python 2.7, `python setup.py test | cat` crashed in the test
runner with

    ======================================================================
    ERROR: test_abort_invalid_po_file (tests.messages.test_pofile.ReadPoTestCase)
    ----------------------------------------------------------------------
      File "src/babel/tests/messages/test_pofile.py", line 458, in test_abort_invalid_po_file
        output = pofile.read_po(buf, locale='fr', abort_invalid=False)
      File "src/babel/babel/messages/pofile.py", line 377, in read_po
        parser.parse(fileobj)
      File "src/babel/babel/messages/pofile.py", line 310, in parse
        self._process_message_line(lineno, line)
      File "src/babel/babel/messages/pofile.py", line 210, in _process_message_line
        self._process_keyword_line(lineno, line, obsolete)
      File "src/babel/babel/messages/pofile.py", line 222, in _process_keyword_line
        self._invalid_pofile(line, lineno, "Start of line didn't match any expected keyword.")
      File "src/babel/babel/messages/pofile.py", line 325, in _invalid_pofile
        print(u"WARNING: Problem on line {0}: {1}".format(lineno + 1, line))
    UnicodeEncodeError: 'ascii' codec can't encode character u'\xe0' in position 84: ordinal not in range(128)

The test suite would show this when printing the `à` in the test pofile contents

    Pour toute question, veuillez communiquer avec Fulano  à nadie@blah.com

But this bug is not confined to the test suite only. Any call to `read_po()`
with invalid .po file could trigger it in non-test code when
`sys.stdout.encoding` is `None`, which is the default for Python 2 when
`sys.stdout.isatty()` is false (as induced e.g. by `| cat`).
    
The fix is to `repr()` the line when printing the WARNING.

---

With th above, `python setup.py test | cat` passes (e.g. on my Ubuntu 18.04).

This was found as part of https://github.com/NixOS/nixpkgs/pull/75676.

### DIFF
--- a/babel/messages/pofile.py
+++ b/babel/messages/pofile.py
@@ -178,7 +178,7 @@ class PoFileParser(object):
             string = ['' for _ in range(self.catalog.num_plurals)]
             for idx, translation in self.translations:
                 if idx >= self.catalog.num_plurals:
-                    self._invalid_pofile("", self.offset, "msg has more translations than num_plurals of catalog")
+                    self._invalid_pofile(u"", self.offset, "msg has more translations than num_plurals of catalog")
                     continue
                 string[idx] = translation.denormalize()
             string = tuple(string)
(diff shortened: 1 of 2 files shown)


nh2:
I've improved the approach slightly with the latest force-push, adding a commit that makes explicit what encoding guarantees we have where.

PR #694, opened 2020-02-11
Improve extract performance via ignoring directory early during os.walk

stkao05:
Currently, the extraction code will do an `os.walk` to perform a deep file search. However, this file exploration could be very slow when there were directories that were deep and contain many files. Even if you have specified some directories to be ignored in the mapping file, the `os.walk` would explore these directories. 

The PR improves the extract process performance via making sure to skip exploring those ignore directory early during `os.walk`.


stkao05:
__Real-life scenario I have experience__

When you are working with front-end, typically you would have a `node_modules` directory in your codebase which contains source codes of all 3rd party lib (similar to Python's `/site-packages/`), and this directory typically is very large.

stkao05:
Ping @akx

babel/messages/extract.py
```diff
@@ -136,11 +136,22 @@ def extract_from_dir(dirname=None, method_map=DEFAULT_MAPPING,
         options_map = {}
 
     absname = os.path.abspath(dirname)
+
+    def dir_filter(method_map, dirname):
+        if dirname.startswith('.') or dirname.startswith('_'):
+            return False
+
+        absdir = os.path.join(root, dirname).replace(os.sep, '/')
```
akx, line 144: Since the logic for ignoring filenames uses

```
filename = relpath(filepath, dirpath)
```

I think this should also use a relative path to the root. Otherwise this might end up ignoring paths that happen to contain an ignored fragment outside the relative root.

That is, if your project lives in `/foobars/myproject/`, and you've ignored `*foobar*` (as it has a special meaning _within_ the `myproject` directory), and you invoke Babel from within `/foobars/myproject/`, this would ignore all files.

akx (changes requested):

babel/messages/extract.py
stkao05: Thanks for the feedbacks. Just applied your suggestion in https://github.com/python-babel/babel/pull/694/commits/252323ae3452ffcc35f2d19b73a9e83084fc6259

akx:
Hi @stkao05 – #832 landed today, so this would need to be rebased :)

PR #696, opened 2020-02-25
Update license.rst

TkTech:
The documentation refers to the license as the license for _flask_ instead of Babel.


akx:
Thank you!

Issue #700, opened 2020-04-01
Times are parsed inconsistently with respect to local time when they should be UTC


rachele-collin:
I recognised that the following tests are failing

* tests/messages/test_frontend.py::InitCatalogTestCase::test_correct_init_more_than_2_plurals
* tests/messages/test_frontend.py::InitCatalogTestCase::test_correct_init_singular_plural_forms
* tests/messages/test_frontend.py::InitCatalogTestCase::test_keeps_catalog_non_fuzzy
* tests/messages/test_frontend.py::InitCatalogTestCase::test_supports_no_wrap
* tests/messages/test_frontend.py::InitCatalogTestCase::test_supports_width
* tests/messages/test_frontend.py::InitCatalogTestCase::test_with_output_dir
* tests/messages/test_frontend.py::CommandLineInterfaceTestCase::test_init_more_than_2_plural_forms
* tests/messages/test_frontend.py::CommandLineInterfaceTestCase::test_init_singular_plural_forms
* tests/messages/test_frontend.py::CommandLineInterfaceTestCase::test_init_with_output_dir

when run with a timezone different to UTC, e.g. like so

```
UTC=UTC+6 pytest -vv -k test_init_more_than_2_plural_forms
```

The assertion printed states a difference in `POT-Creation-Date:` the values are `2007-04-01 15:30+0200` vs `2007-04-01 9:30+0200` (ignore the +0200 timezone as that is not the timezone of the system but more or less just copied from input to output).
The date and time is actually a hardcoded `2007-04-01 15:30+0200` in both the expected string as well as in the source pot-file.

I dug a little in the source code as at some point something (i.e. Babel ; who else?) must do some magic conversion to convert that string into a (wrong) time and back to a string. I think I found the place in `babel/messages/catalog.py` function

```python
def _parse_datetime_header(value):
    match = re.match(r'^(?P<datetime>.*?)(?P<tzoffset>[+-]\d{4})?$', value)

    tt = time.strptime(match.group('datetime'), '%Y-%m-%d %H:%M')
    ts = time.mktime(tt)
    dt = datetime.fromtimestamp(ts)

    # ... tzoffset handling ...
```
which as you can see does some wild conversation stuff back and forth to convert the string into a `datetime` object. The part to point out is the `time.mktime` call.
From the python documentation [0]:

> | From                        | To                      | Use                                                           |
> |---------------------------|----------------------|------------------------------------------------------|
[...]

PR #701, opened 2020-04-01
Parse string date times indepentent of time zone

rachele-collin:
Parsing a date time string (such as for ``POT-Creation-Date`` or
``PO-Revision-Date``) as UTC time, in particular independent of the
local time zone.

Closes: #700

### DIFF
--- a/babel/messages/catalog.py
+++ b/babel/messages/catalog.py
@@ -10,7 +10,6 @@
 """
 
 import re
-import time
 
 from cgi import parse_header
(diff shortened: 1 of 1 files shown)


rachele-collin:
Any chance this would be merged?

akx (approved):
Looks good. The docs say `mktime` is the inverse of `localtime`, and we're taking care of the timezone just after this, so a tz-naive parsing function is the right thing.

PR #706, opened 2020-04-15
Copy project, version info when updating from template

### LINKED ISSUE #705: 'update_catalog' doesn't update 'Project-Id-Version'
The distutils `extract_messages` command will generate a template file with an updated `Project-Id-Version` field, based on the version defined in e.g. `setuptools.setup()`. Unfortunately the `update_catalog` tool doesn't copy this across to the actual `po` files. I'm still figuring out gettext so I'm not sure if this is necessary/what it's used for, but it does _look_ wrong, at least.

stephenfin:
When updating catalogs using the 'update_catalog' command, one would expect to see the 'Project-Id-Version' header updated using the information extracted with distutils. Make it so.

Closes: #705

### DIFF
--- a/babel/messages/catalog.py
+++ b/babel/messages/catalog.py
@@ -822,8 +822,10 @@ class Catalog(object):
             # template's header
             self.header_comment = template.header_comment
 
-        # Make updated catalog's POT-Creation-Date equal to the template
-        # used to update the catalog
+        # Make updated catalog's Project-Id-Version, POT-Creation-Date equal to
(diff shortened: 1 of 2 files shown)


stephenfin:
Oh wow, I'd forgotten about this. It'd be great to land this but fair warning, I've totally forgotten the context on this by now (in my defense, it's been 2 COVID-ful years)

catalinaIlabaca:
@stephenfin Hello, did you came up with a solution? thanks

PR #710, opened 2020-05-04
stop using deprecated ElementTree methods "getchildren()" and "getiterator()"

FelixSchwarz:
Both methods were removed in Python 3.9 as mentioned in the release notes:

> Methods getchildren() and getiterator() of classes ElementTree and Element in
> the ElementTree module have been removed. They were deprecated in Python 3.2.
> Use iter(x) or list(x) instead of x.getchildren() and x.iter() or
> list(x.iter()) instead of x.getiterator().

### DIFF
--- a/scripts/import_cldr.py
+++ b/scripts/import_cldr.py
@@ -598,7 +598,7 @@ def parse_calendar_months(data, calendar):
         for width in ctxt.findall('monthWidth'):
             width_type = width.attrib['type']
             widths = ctxts.setdefault(width_type, {})
-            for elem in width.getiterator():
+            for elem in width.iter():
                 if elem.tag == 'month':
                     _import_type_text(widths, elem, int(elem.attrib['type']))
                 elif elem.tag == 'alias':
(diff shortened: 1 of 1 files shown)


scripts/import_cldr.py
```diff
@@ -880,7 +880,7 @@ def parse_interval_formats(data, tree):
             interval_formats[None] = elem.text
         elif elem.tag == "intervalFormatItem":
             skel_data = interval_formats.setdefault(elem.attrib["id"], {})
-            for item_sub in elem.getchildren():
+            for item_sub in iter(elem):
```
hroncok, line 883:
```suggestion
            for item_sub in elem:
```

I don't fully understand the difference between `element.iter()` and `iter(element)` but here I believe the `iter()` call is redundant.

FelixSchwarz: Good point, I tweaked my commit accordingly.

akx:
Closed via #712.

scripts/import_cldr.py
hroncok: I still have the impression that iterating over `element.iter()` (no arguments) and iterating over `element` will iterate over the children. Hence maybe the entire code can be simplified.

FelixSchwarz: I think you are right - is #713 what you suggested?

hroncok: Yes, but I am unsure. If the code is tested, good.

Issue #675, opened 2019-11-24
PyPy3 7.2.0 test failure: 


mgorny:
# Overview Description

Tests are failing with PyPy3 7.2.0.

# Steps to Reproduce

1. Install PyPy3 7.2.0
2. `tox -e pypy3`

# Actual Results

```
==================================================================== FAILURES =====================================================================
___________________________________________________________ test_format_current_moment ____________________________________________________________

monkeypatch = <_pytest.monkeypatch.MonkeyPatch object at 0x00007fa3a03ccf00>

    def test_format_current_moment(monkeypatch):
        import datetime as datetime_module
        frozen_instant = datetime.utcnow()
    
        class frozen_datetime(datetime):
    
            @classmethod
            def utcnow(cls):
                return frozen_instant
    
        # Freeze time! Well, some of it anyway.
        monkeypatch.setattr(datetime_module, "datetime", frozen_datetime)
>       assert dates.format_datetime(locale="en_US") == dates.format_datetime(frozen_instant, locale="en_US")

tests/test_dates.py:775: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _
babel/dates.py:728: in format_datetime
    datetime = _ensure_datetime_tzinfo(_get_datetime(datetime), tzinfo)
babel/dates.py:110: in _get_datetime
    return datetime_.utcnow()
/usr/lib/pypy3.6/lib-python/3/datetime.py:1535: in utcnow
    return cls.utcfromtimestamp(t)
/usr/lib/pypy3.6/lib-python/3/datetime.py:1523: in utcfromtimestamp
    return cls._fromtimestamp(t, True, None)
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _

cls = <class 'datetime.datetime'>, t = 1574616833.0, utc = True, tz = None

    @classmethod
    def _fromtimestamp(cls, t, utc, tz):
[...]

PR #714, opened 2020-05-10
tests/test_dates.py: Fix broken test

kerizane:
datetime was being incorrectly patched, so the unittest was failing

Closes https://github.com/python-babel/babel/issues/675


mgorny:
Ping. It would be nice to have this fixed.

akx:
This was resolved in 9a9d3c60ec05b346e87f2241b7ce976ed1a6bf0b (#754).

PR #720, opened 2020-07-16
allow header_comment to be passed as an option to extract_message

### LINKED ISSUE #82: Allow header_comment to be passed as an option to extract_message
In Trac and Apache Bloodhound we utilize [babel.extract_messages](http://babel.readthedocs.org/en/stable/setup.html#extract-messages). For Apache Bloodhound, we'd like to be able to specify the [header_comment](http://babel.readthedocs.org/en/stable/api/messages/catalog.html#babel.messages.catalog.Catalog.header_comment) parameter when creating the Catalog. We would like to add `header_comment` as an option for `extract_messages`.

I am planning to prepare a patch, but wanted to open the issue first in case there was any early feedback.

muhamed-hafez:
This fixes #82

### DIFF
--- a/babel/messages/frontend.py
+++ b/babel/messages/frontend.py
@@ -26,7 +26,7 @@ from io import StringIO
 from babel import __version__ as VERSION
 from babel import Locale, localedata
 from babel.core import UnknownLocaleError
-from babel.messages.catalog import Catalog
+from babel.messages.catalog import Catalog, DEFAULT_HEADER
 from babel.messages.extract import DEFAULT_KEYWORDS, DEFAULT_MAPPING, check_and_call_extract_file, extract_from_dir
 from babel.messages.mofile import write_mo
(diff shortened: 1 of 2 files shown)


babel/messages/frontend.py
```diff
@@ -327,6 +327,8 @@ class extract_messages(Command):
          'files or directories with commas(,)'),  # TODO: Support repetition of this argument
         ('input-dirs=', None,  # TODO (3.x): Remove me.
          'alias for input-paths (does allow files as well as directories).'),
+        ('header-comment=', None,
+         'header comment for the catelog'),
```
akx, line 331:
```suggestion
         'header comment for the catalog'),
```

akx (changes requested):

akx (approved):
I took the liberty of fixing the typo, and also fixing things so the default header is used by... well, default.

There's also a test case now.

Issue #723, opened 2020-07-31
RecursionError in LazyProxy class instead of AttributeError, when lazy function raise AttributeError


nikifkon:
# Overview Description
When `babel.support.LazyProxy` gets function that raise AttributeError, RecursionError raises instead of it

# Steps to Reproduce
```python
from babel.support import LazyProxy 
def func(): 
    raise AttributeError("test") 
LazyProxy(func).value
```

# Actual Results
`RecursionError: maximum recursion depth exceeded while calling a Python object`

# Expected Results
`AttributeError: test`

# Reproducibility
Always

# Additional Information:

PR #724, opened 2020-07-31
LazyProxy: Handle AttributeError in specified func

nikifkon:
Fixes https://github.com/python-babel/babel/issues/723

### DIFF
--- a/babel/support.py
+++ b/babel/support.py
@@ -256,7 +256,7 @@ class LazyProxy(object):
 
     def __getattr__(self, name):
         if self._attribute_error is not None:
-            raise AttributeError(self._attribute_error)
+            raise self._attribute_error
         return getattr(self.value, name)
 
     def __setattr__(self, name, value):
--- a/babel/support.py
+++ b/babel/support.py
@@ -165,7 +165,7 @@ class LazyProxy(object):
     Hello, universe!
     Hello, world!
     """
-    __slots__ = ['_func', '_args', '_kwargs', '_value', '_is_cache_enabled']
+    __slots__ = ['_func', '_args', '_kwargs', '_value', '_is_cache_enabled', '_attribute_error']
 
     def __init__(self, func, *args, **kwargs):
         is_cache_enabled = kwargs.pop('enable_cache', True)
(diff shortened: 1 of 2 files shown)


babel/support.py
```diff
@@ -249,6 +255,8 @@ def __delattr__(self, name):
         delattr(self.value, name)
 
     def __getattr__(self, name):
+        if self._attribute_error is not None:
+            raise AttributeError(self._attribute_error)
```
akx, line 259: Shouldn't this just `raise self._attribute_error`, since you're already stashing the exception itself?

tests/test_support.py
```diff
@@ -279,6 +279,17 @@ def first(xs):
         self.assertEqual(2, proxy.value)
         self.assertEqual(1, proxy_deepcopy.value)
 
+    def test_handle_atribute_error(self):
```
akx, line 282:
```suggestion
    def test_handle_attribute_error(self):
```

tests/test_support.py
```diff
@@ -279,6 +279,17 @@ def first(xs):
...
+
+        def raise_atribute_error():
```
akx, line 284:
```suggestion
        def raise_attribute_error():
```

tests/test_support.py
```diff
@@ -279,6 +279,17 @@ def first(xs):
...
+            raise AttributeError('message')
+
+        proxy = support.LazyProxy(raise_atribute_error)
```
akx, line 287:
```suggestion
        proxy = support.LazyProxy(raise_attribute_error)
```

akx (changes requested):

babel/support.py
nikifkon: Done in last commit. Thanks!

babel/support.py
```diff
@@ -175,11 +175,17 @@ def __init__(self, func, *args, **kwargs):
...
 
     @property
     def value(self):
         if self._value is None:
-            value = self._func(*self._args, **self._kwargs)
+            try:
+                value = self._func(*self._args, **self._kwargs)
+            except AttributeError as error:
+                object.__setattr__(self, '_attribute_error', error)
+                raise AttributeError()
```
akx, line 187: One more nitpick – why construct a new error? You could just `raise` to raise the same `AttributeError` just stored.
```suggestion
                raise
```

akx (approved):

PR #726, opened 2020-08-21
Added group_separator feature in number formatting

Abdullahjavednesar:
Added a new **group_separator** feature in number formatting. **group_separator** is a Boolean parameter default set as `True`, when `False`, it simply means the output should be free of group separator.
Example:
```
format_decimal(23434567890567.1224, locale='fr_CA', group_separator=False)
>>> '23434567890567,12'
```
Default behaviour remains unchanged.
Example:
```
format_decimal(23434567890567.1224, locale='en_US')
>>> '23,434,567,890,567.12
```

### DIFF
--- a/babel/numbers.py
+++ b/babel/numbers.py
@@ -373,7 +373,7 @@ def get_decimal_quantum(precision):
 
 
 def format_decimal(
-        number, format=None, locale=LC_NUMERIC, decimal_quantization=True):
+        number, format=None, locale=LC_NUMERIC, decimal_quantization=True, group_separator=True):
     u"""Return the given decimal number formatted for a specific locale.
 
     >>> format_decimal(1.2345, locale='en_US')
@@ -413,7 +413,7 @@ def format_decimal(
         format = locale.decimal_formats.get(format)
     pattern = parse_pattern(format)
     return pattern.apply(
-        number, locale, decimal_quantization=decimal_quantization)
+        number, locale, decimal_quantization=decimal_quantization, group_separator=group_separator)
 
 
 class UnknownCurrencyFormatError(KeyError):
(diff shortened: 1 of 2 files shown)


kdeldycke:
Looks nice @Abdullahjavednesar!

Can you add some test to demonstrates the use of explicit `group_separator=True`? That way we can make sure `group_separator` is properly recognize.

It might also be great to have the docstrings updated so we can have that new option documented for end-users.

Abdullahjavednesar:
Hi @kdeldycke, thanks for the review.

>Can you add some test to demonstrates the use of explicit group_separator=True? That way we can make sure group_separator is properly recognize.

Sure, will add more tests to cover those too.

>It might also be great to have the docstrings updated so we can have that new option documented for end-users.

Yes, makes sense, I'll update the docstrings.

Abdullahjavednesar:
Hi @kdeldycke @akx please have a look. If looks good, can this be merged?

kdeldycke:
LGTM! Thanks @Abdullahjavednesar for your contribution! I'm no official maintainer though, I just added a couple of features in the same spirit as yours, and as a fellow contributor I find it great! :)

Now we need a review and eventual merge decision from an official maintainer.

Abdullahjavednesar:
Thanks a lot @kdeldycke for your reviews, and those good words. Indeed Open source is a great platform to work and learn collaboratively :)

@akx a request if you could please review this :)

akx (approved):
Seems fine by me, though I'd like to know what the actual use case for this is :)

Abdullahjavednesar:
Thanks @akx for merging :)

We came across this use case where we had to export files with Amount column and amount should just have a fraction symbol and not the group symbol. I believe this would be the cleanest approach so I've added this feature to all applicable functions.

Thanks again @kdeldycke @akx :)

PR #729, opened 2020-09-25
allow header_comment to be passed as an option to extract_message

jabhishek87:
fixed issue #82 ,  allow header_comment to be passed as an option to extract_message


akx:
This is the exact same code as #720, only reattributed to another person.

PR #733, opened 2020-10-01
Started work on supporting CLDR 24

sonkelpesh:
Currently, the extraction code will do an os.walk to perform a deep file search. However, this file exploration could be very slow when there were directories that were deep and contain many files. Even if you have specified some directories to be ignored in the mapping file, the os.walk would explore these directories.

The PR improves the extract process performance via making sure to skip exploring those ignore directory early during os.walk.


akx:
Hi there! Pulling up another person's work from 2013 as a pull request doesn't seem to be very productive. Please let me know if closing this is in error somehow.

Issue #740, opened 2020-10-09
get_timezone_location does not work with python's ZoneInfo object


youtux:
`get_timezone_location` does not render correctly the ZoneInfo objects. These are part of python as of python 3.9 (and available as backport).

MCVE:
`$ pip install backports.zoneinfo babel`
```
# tz.py
from babel.dates import get_timezone_location
from backports import zoneinfo
import pytz

ams_native = zoneinfo.ZoneInfo("Europe/Amsterdam")
ams_pytz = pytz.timezone("Europe/Amsterdam")

print(get_timezone_location(ams_pytz))  # output: "Netherlands Time"
print(get_timezone_location(ams_native))  #  output: "Unknown Region (CEST) Time"
```
Python version: 3.6.11
babel version: 2.8.0

I think this can be easily fixed by checking the `timezone.key` attribute. Shall I make a PR?

akx:
Please do! :)

PR #741, opened 2020-10-11
Handle ZoneInfo objects in get_timezone_location, get_timezone_name

youtux:
Python 3.9 introduced ZoneInfo objects, and they are available as a [backport](https://pypi.org/project/backports.zoneinfo/) back to python 3.6.
This PR adds support for these objects.

Fixes https://github.com/python-babel/babel/issues/740

### DIFF
--- a/babel/dates.py
+++ b/babel/dates.py
@@ -76,6 +76,21 @@ def _get_dt_and_tzinfo(dt_or_tzinfo):
     return dt, tzinfo
 
 
+def _get_tz_name(dt_or_tzinfo):
+    """
+    Get the timezone name out of a time, datetime, or tzinfo object.
+
+    :rtype: str
+    """
+    dt, tzinfo = _get_dt_and_tzinfo(dt_or_tzinfo)
+    if hasattr(tzinfo, 'zone'):  # pytz object
+        return tzinfo.zone
+    elif hasattr(tzinfo, 'key') and tzinfo.key is not None:  # ZoneInfo object
+        return tzinfo.key
+    else:
+        return tzinfo.tzname(dt or datetime.utcnow())
+
+
 def _get_datetime(instant):
     """
     Get a datetime out of an "instant" (date, time, datetime, number).
@@ -500,15 +515,9 @@ def get_timezone_location(dt_or_tzinfo=None, locale=LC_TIME, return_city=False):
     :return: the localized timezone name using location format
 
     """
-    dt, tzinfo = _get_dt_and_tzinfo(dt_or_tzinfo)
     locale = Locale.parse(locale)
 
-    if hasattr(tzinfo, 'zone'):  # pytz object
-        zone = tzinfo.zone
-    elif hasattr(tzinfo, 'key') and tzinfo.key is not None:  # ZoneInfo object
-        zone = tzinfo.key
-    else:
-        zone = tzinfo.tzname(dt or datetime.utcnow())
+    zone = _get_tz_name(dt_or_tzinfo)
 
     # Get the canonical time-zone code
     zone = get_global('zone_aliases').get(zone, zone)
@@ -621,12 +630,7 @@ def get_timezone_name(dt_or_tzinfo=None, width='long', uncommon=False,
     dt, tzinfo = _get_dt_and_tzinfo(dt_or_tzinfo)
     locale = Locale.parse(locale)
 
-    if hasattr(tzinfo, 'zone'):  # pytz object
-        zone = tzinfo.zone
-    elif hasattr(tzinfo, 'key') and tzinfo.key is not None:  # ZoneInfo object
-        zone = tzinfo.key
-    else:
-        zone = tzinfo.tzname(dt)
+    zone = _get_tz_name(dt_or_tzinfo)
 
     if zone_variant is None:
         if dt is None:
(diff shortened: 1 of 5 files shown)


tests/test_dates.py
```diff
@@ -583,60 +600,113 @@ def test_get_timezone_gmt():
...
+            {"locale": "en_US", "return_zone": True},
+            u"America/Los_Angeles",
+        ),
+        ("America/Los_Angeles", {"width": "short", "locale": "en_US"}, u"PDT"),
+    ],
+)
+@pytest.mark.parametrize("timezone_getter", ["zoneinfo.ZoneInfo"], indirect=True)
+def test_get_timezone_name_time_zoneinfo(timezone_getter, tzname, params, expected):
+    """zoneinfo will correctly detect DST from the object.
+    FIXME: this test will only succeed in the winter.
```
youtux, line 702: how should we address this? Should I just test with a full datetime object instead?

youtux:
@akx could you have a look at this?

tests/test_dates.py
```diff
@@ -24,6 +24,23 @@
...
+        try:
+            import zoneinfo
+        except ImportError:
+            try:
+                from backports import zoneinfo
+            except ImportError:
+                pytest.skip("zoneinfo not available")
+        return zoneinfo.ZoneInfo
+    else:
+        raise NotImplementedError
```
akx, line 41: How about something like this, to precompute the tz implementations we have? This way tests wouldn't end up skipped half of the time.

```suggestion

tz_implementations = {"pytz.timezone": timezone}
try:
    import zoneinfo
except ImportError:
    try:
        from backports import zoneinfo
    except ImportError:
        zoneinfo = None
if zoneinfo:
     tz_implementations["zoneinfo.ZoneInfo"] = zoneinfo
      
@pytest.fixture(params=list(tz_implementations))
def timezone_getter(request): 
    return tz_implementations[request.param]
```

babel/dates.py
```diff
@@ -503,8 +503,10 @@ def get_timezone_location(dt_or_tzinfo=None, locale=LC_TIME, return_city=False):
     dt, tzinfo = _get_dt_and_tzinfo(dt_or_tzinfo)
     locale = Locale.parse(locale)
 
-    if hasattr(tzinfo, 'zone'):
+    if hasattr(tzinfo, 'zone'):  # pytz object
         zone = tzinfo.zone
+    elif hasattr(tzinfo, 'key') and tzinfo.key is not None:  # ZoneInfo object
+        zone = tzinfo.key
     else:
         zone = tzinfo.tzname(dt or datetime.utcnow())
```
akx, line 511: This stanza seems to be more or less the one in the below change fragment – could they be refactored into one function like `get_tz_name(tzinfo, default=None)`?

tests/test_dates.py
akx: I... guess? Getting a timezone from just a time-of-day seems weird anyway.

akx (changes requested):

tests/test_dates.py
youtux: Shall I just delete this test case?
Alternatively, I can try patching datetime.utcnow() to return a constant datetime, and test against that. But it seems that the datetime module is not patchable: https://stackoverflow.com/a/192857/714760. So it would be really hard to test this case.

youtux: I will remove this test.

tests/test_dates.py
youtux: @akx, can you check my answer and tell me if you then prefer to have them showing as skipped or not?

akx: Sorry for the delay. Yeah, you do have a point with the skipped tests there – we can probably leave this as-is.

akx:
@youtux, re: 

> When is support for these old pythons going to be removed? I frankly don't see a point in supporting EOL pythons.

Regarding this, #569 is probably still relevant. I'm afraid a lot of old code and systems rely on "just give me some Babel, I don't care if it's compatible", and if a release drops support for an older Python, breakage could be immense. I do agree we should drop support for EOL versions of Python, I just don't know what the graceful way to do it is.

akx (approved):

youtux:
> Regarding this, #569 is probably still relevant. I'm afraid a lot of old code and systems rely on "just give me some Babel, I don't care if it's compatible", and if a release drops support for an older Python, breakage could be immense. I do agree we should drop support for EOL versions of Python, I just don't know what the graceful way to do it is.

It should not be a problem if the new release will correctly declare which python versions are supported by using the [python_requires](https://packaging.python.org/guides/distributing-packages-using-setuptools/#python-requires) declaration in the setup.py/setup.cfg.
`pip` running on python2 will just pick whatever latest version of pytz still supports that. That's how most libraries handle the transition (e.g. https://github.com/pytest-dev/pytest-factoryboy/blob/910c4b0e2086c4944164bb3f631d4ac67d9d4283/setup.py#L31)

akx:
@youtux Yeah, good point. That should work (assuming, you know, that people install packages with a modern `pip` and not e.g. a prehistoric version of, say, `easy_install`...).

I think we should still have one more release with the current guaranteed compatibility, just to ship e.g. the new CLDR changes.

youtux:
> That should work (assuming, you know, that people install packages with a modern pip and not e.g. a prehistoric version of, say, easy_install...).

You can't save everyone :)

akx:
I know, but a package that still has 26,000 daily downloads on Py2 should take care of those users at least somewhat! :)

youtux:
I'm not sure it should. Popular projects like django and pytest already dropped support for python2 completely.
I don't think you should make it **your** problem to be honest.

PR #744, opened 2020-10-23
'else if' should be simplified to 'elif'

Bombear435:
also, the indentation changes a little bit

### DIFF
--- a/babel/messages/catalog.py
+++ b/babel/messages/catalog.py
@@ -806,24 +806,23 @@ class Catalog(object):
                 key = self._key_for(message.id, message.context)
                 if key in messages:
                     _merge(message, key, key)
-                else:
-                    if not no_fuzzy_matching:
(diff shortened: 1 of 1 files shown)


babel/messages/catalog.py
```diff
@@ -806,24 +806,23 @@ def _merge(message, oldkey, newkey):
...
+                                                fuzzy_candidates.keys(), 1)
+                    if matches:
+                        newkey = matches[0]
+                        newctxt = fuzzy_candidates[newkey]
+                        if newctxt is not None:
+                            newkey = newkey, newctxt
+                        _merge(message, newkey, key)
+                        continue
+
+                self[message.id] = message
```
akx, line 825: This seems like it used to be within the `else` block but outside the `if` block, so this change changes the semantics of the code.

PR #745, opened 2020-10-25
Update README.rst

sm-raiyyan:
Made it more concise.


akx:
This PR adds a typo and just removes a couple newlines. I'm sorry, but I'm deeming it invalid. Please let me know if you think that's not the case.

PR #746, opened 2020-10-25
Update README.rst

yourdad285:


akx:
This PR isn't a substantial change. I'm sorry, but I'm deeming it invalid. Please let me know if that shouldn't be the case.

PR #758, opened 2020-11-17
added Bengali(Bangladesh) to the locale aliases

mahihossain:


akx:
Hi @mahihossain – could you explain the rationale for this change? The alias table has remained unchanged since 2007, so I'm wondering why it would need a change now.

akx:
Closing as stale.

PR #761, opened 2020-11-29
fix #53 (now directories starting with '_' are not ignored)

### LINKED ISSUE #53: messages not extracted from folders starting with _ or .

NChechulin:
Fixed a small issue, when babel ignored directories starting with `_`.
I thought that it's still a good idea to ignore dirs whose entire name is `_`.

### DIFF
--- a/babel/messages/extract.py
+++ b/babel/messages/extract.py
@@ -139,7 +139,7 @@ def extract_from_dir(dirname=None, method_map=DEFAULT_MAPPING,
     for root, dirnames, filenames in os.walk(absname):
         dirnames[:] = [
             subdir for subdir in dirnames
-            if not (subdir.startswith('.') or subdir.startswith('_'))
+            if not (subdir.startswith('.') or subdir == '_')
         ]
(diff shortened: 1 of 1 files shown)


babel/messages/extract.py
```diff
@@ -139,7 +139,7 @@ def extract_from_dir(dirname=None, method_map=DEFAULT_MAPPING,
     for root, dirnames, filenames in os.walk(absname):
         dirnames[:] = [
             subdir for subdir in dirnames
-            if not (subdir.startswith('.') or subdir.startswith('_'))
+            if not (subdir.startswith('.') or subdir == '_')
```
gladykov, line 142: Your PR does not make sense.

```
>>> a = "_"
>>> a.startswith("_")
True
>>>
```

NChechulin: Please, consider reading the issue tread attached ([this one](https://github.com/python-babel/babel/issues/53)).
In my PR description, I didn't clarify what exactly I tried to fix - that's my fault.

PR #763, opened 2020-12-03
Add Python 3.8 & 3.9 support

eumiro:
Add py38 and py39 to `tox.ini`.

Would you consider dropping support for 2.7, 3.4, and 3.5 (all of them after EOL)?


akx:
This was done in #754 (which I forgot to merge). Thanks though!

PR #769, opened 2020-12-12
Use try-catch block in order to handle extracting f-strings

evonbevern:
Before this request, f-strings were causing the program to crash with an exception. After the try catch block is added in extract.py, f-strings will raise a NameError.


akx (changes requested):
Could you add a test case, please?

PR #773, opened 2021-02-12
Javascript lexer improvements

gitaarik:
Add option `parse_template_string` that improves parsing of javascript templates expressions, like:

    `${gettext('my text')}`

And it also works with nested expressions like:

    ${`some text ${gettext('my text')} more text`}

You have to enable template string extraction with an option in the [`option_map`](http://babel.pocoo.org/en/latest/api/messages/extract.html?highlight=options_map#babel.messages.extract.extract_from_dir):

    extract_from_dir(
        # ...
        options_map={
            '**': {'parse_template_string': True}
        }
    )

Then it should work.

On this branch I also fixed another issue that I was facing. In a JS library that I use, there was regex with a `/` inside a character class `[a-z]`. The babel JS lexer thought this closed the regex, but JavaScript allows this as a normal match character within a character class, and does not close the regex. I updated the regex so that it will cover this exception.

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -1,6 +1,14 @@
 Babel Changelog
 ===============
 
+Version 2.10.0
+--------------
+
+Improvements
+~~~~~~~~~~~~
+
+* Support for javascript template strings
+
 Version 2.9.1
 -------------
 
--- a/babel/__init__.py
+++ b/babel/__init__.py
@@ -21,4 +21,4 @@ from babel.core import UnknownLocaleError, Locale, default_locale, \
     negotiate_locale, parse_locale, get_locale_identifier
 
 
-__version__ = '2.9.1'
+__version__ = '2.10.0'
--- a/babel/messages/extract.py
+++ b/babel/messages/extract.py
@@ -17,6 +17,7 @@
     :license: BSD, see LICENSE for more details.
 """
 
+import io
 import os
 from os.path import relpath
 import sys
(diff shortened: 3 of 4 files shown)


gitaarik:
Hi @akx . I've been working on this feature for a while. In my setup I need these changes. However I have some problems when using my own fork in my project. So I have some questions regarding using a fork of babel in my project, and also what needs to be done to get this merged, so I don't have to use my fork anymore.

In my project, which uses [Pipenv](https://pipenv.pypa.io/en/latest/). Inside the `Pipfile`, I use a [VCS descriptor](https://pipenv.pypa.io/en/latest/basics/#a-note-about-vcs-dependencies) to specify my python-babel fork:

    babel = {editable = true, ref = "javascript-improvements-1", git = "https://github.com/gitaarik/babel"}

This works partly. When I run `python-babel`'s [`babel.messages.extract.extract_from_dir`](http://babel.pocoo.org/en/latest/api/messages/extract.html?highlight=extract_from_dir#babel.messages.extract.extract_from_dir), it finds `gettext()` functions that it didn't found before.

However, when I run my Django devserver, on *some* requests (not all of them, I'm not sure yet why it only happens on some requests), the request fails with a exception coming from `python-babel`:

    RuntimeError: The babel data files are not available. This usually happens because you are using a source checkout from Babel and you did not build the data files.  Just make sure to run "python setup.py import_cldr" before installing the library.

I noticed that when I go into the virtualenv directory (which is created by Pipenv and can be retrieved with `pipenv --venv`) and then from there go to `src/babel/`, I can run the command `python setup.py import_cld` to generate the locale files and then it seems to work.

I will have to look into the possibilities to integrate this in my project in a way that the locale files would be generated when I build and deploy my project. In the mean time it would be good for us if the changes would eventually be merged into the official package.

So I have two questions:
- Are there easier ways of using my fork in my project, for development and deployment?
- What would be needed to get these changes merged into the official package, so that we don't have to make things complicated with a fork?

gitaarik:
Closing this MR now, because I splitted this one up in 2 separate MRs:

- https://github.com/python-babel/babel/pull/792
- https://github.com/python-babel/babel/pull/791

Please reply in the appropriate new MRs.

PR #785, opened 2021-05-07
Implement `--init-missing` option for `pybabel update`

RuRo:
Currently, if an output `.po` file doesn't exist during `pybabel update`, the update fails when trying to open the missing file. This PR adds a boolean `--init-missing` option. If this option is enabled during an update and an output `.po` file is missing, it is created and initialized the same way it would be by `pybabel init`.

---

**Before this PR** you had to run two commands like

```sh
pybabel init   --domain=${DOM} --locale=${LOC} --input-file=pot/${DOM}.pot --output-dir=po/
pybabel update --domain=${DOM} --locale=${LOC} --input-file=pot/${DOM}.pot --output-dir=po/
```

if `po/${LOC}/LC_MESSAGES/${DOM}.po` didn't exist.

If the file **did** already exist, you had to only run the second command (otherwise, running `init` on an existing file overwrites it, removing any previous translations, which is not desirable).

---

**After this PR** you will be able to run a single command and get the expected behaviour, no matter if the output file existed previously or not:

```sh
pybabel update --init-missing --domain=${DOM} --locale=${LOC} --input-file=pot/${DOM}.pot --output-dir=po/
```

### DIFF
--- a/babel/messages/frontend.py
+++ b/babel/messages/frontend.py
@@ -668,6 +668,8 @@ class update_catalog(Command):
          'into several lines'),
         ('ignore-obsolete=', None,
          'whether to omit obsolete messages from the output'),
+        ('init-missing=', None,
+         'if any output files are missing, initialize them first'),
         ('no-fuzzy-matching', 'N',
          'do not use fuzzy matching'),
         ('update-header-comment', None,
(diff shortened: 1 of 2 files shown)


RuRo:
@akx Sorry to ping you, but what's a cowboy gotta do 'round here to get his PR reviewed 🤠?

akx (changes requested):
The code looks good, thank you.

I'd appreciate a non-trivial test that does test that this option also does initialize the files as it promises.

RuRo:
@akx Done.

akx (approved):
Looks good, thanks. Sorry for the delay in reviewing!

Issue #640, opened 2019-05-09
Babel gets confused by JavaScript regular expressions


remram44:
After a regular expression like `/>/`, Babel ignores all messages until the next one.

Mapping:
```
[javascript: **.js]
encoding = utf-8
extract_messages = gettext
```
Source file:
```
alert(gettext("First message"));
var expr = />/g;
alert(gettext("Second message"));
var expr2 = /'/g;
alert(gettext("Third message"));
```
Result POT:
```
"Generated-By: Babel 2.6.0\n"

#: test.js:1
msgid "First message"
msgstr ""

#: test.js:5
msgid "Third message"
msgstr ""
```

akx:
The JavaScript lexer in Babel is far from perfect, unfortunately...
If you feel like trying to hack at the lexer to fix this, by all means please do!

It might be a good idea for a future version to use e.g. https://github.com/Kronuz/esprima-python when it's available (i.e. an optional dependency) for more robust parsing.

remram44:
Thanks, I might take a look. I am low on time unfortunately right now.

remram44:
Seems that `/>` gets recognized as a JSX tag.

remram44:
I don't see how to fix this. The tokenizer can't possibly tell if `/>` is the start of a regexp or the end of a JSX tag, they are the same tokens. The Javascript grammar is too ambiguous for this to be decided at the tokenizer level.

A parser would be able to tell if it's current in a JSX tag or not and resolve the ambiguity, but there is no such parsing being done in Babel. Simply keeping track of JSX open/close tokens wouldn't be enough because regexps can appear inside JSX tags (`<hr class={ /\w+/.exec(var)[0] }/>`).

I think I'll have to leave that alone, sorry!

Issue #616, opened 2018-12-26
pybabel extract issue for javascript regex


cdsheen:
If a line contains p.replace(/>/g, '&gt;').repalce('/', ''), pybabel stop extracting msgs after it.

```
$ pybabel --version
pybabel 2.6.0

$ cat mapping.ini
[javascript: *.js]

$ cat test.js
x = _('string1');
y = _('string2');
p.replace(/>/g, '&gt;').repalce('/', '');
z = _('string3');

$  pybabel extract --omit-header -F mapping.ini test.js -o test.po      
extracting messages from test.js
writing PO template file to test.po

$ cat test.po
#: test.js:1
msgid "string1"
msgstr ""

#: test.js:2
msgid "string2"
msgstr ""

$
("string3" is not extracted)
```

Though the following lines works fine.

p.replace(/\\>/g, '&gt;').replace('/', '');   # okay if > is escaped
p.replace(/>/g, '&gt;');                         # okay if without slash (/) after
p.replace(/</g, '&lt;').replace('/', '');     # okay for <

PR #791, opened 2021-06-09
Improved javascript regex regocnizing for extracting js messages

gitaarik:
In a JS library that I use, there was regex with a `/` inside a character class `[a-z]`. The babel JS lexer thought this closed the regex, but JavaScript allows this as a normal match character within a character class, and does not close the regex. I updated the regex so that it will cover this exception.

This could also possibly fix these issues, or set the stage for further improvements regarding these issues:

- https://github.com/python-babel/babel/issues/640
- https://github.com/python-babel/babel/issues/616

How do I add unit tests for this? I haven't found clear documentation about this.


babel/__init__.py
```diff
@@ -21,4 +21,4 @@
     negotiate_locale, parse_locale, get_locale_identifier
 
 
-__version__ = '2.9.1'
+__version__ = '2.9.2'
```
akx, line 24: No need to do this change in this PR – it'll be done as part of a release.

CHANGES
```diff
@@ -1,6 +1,16 @@
...
 
+Version 2.9.2
+--------------
+
+Bugfixes
+~~~~~~~~~~~~
+
+* Regex for parsing Javascript regexes improved. Before this, the lexer
+  couldn't regocnize certain javascript regexes, breaking the parsing of JS
+  files. This is fixed now.
```
akx, line 12:
```suggestion
* Regex for parsing JavaScript regexes improved. Before this, the lexer
  couldn't recognize certain regexes, breaking the parsing of JS files.
```

CHANGES
```diff
@@ -1,6 +1,16 @@
 Babel Changelog
 ===============
 
+Version 2.9.2
+--------------
```
akx, line 5: I appreciate the changelog entry – I usually distill them by hand around release time, so this spares me some effort.

However, this'll probably end up in 2.10, so maybe just

```suggestion
Next version
------------
```
for now?

CHANGES
```diff
@@ -1,6 +1,16 @@
...
+
+Bugfixes
+~~~~~~~~~~~~
```
akx, line 8: Sphinx would otherwise complain about too many squigglies...
```suggestion
Bugfixes
~~~~~~~~
```

akx (changes requested):
Beyond the other comments, I'd appreciate tests showing how this improves parsing for the files previously accepted :)

gitaarik:
Hey @akx , could you please give me some guidance about how/where I should add unit tests for this? I haven't found clear documentation about this.

gitaarik:
@akx Got it now, I wasn't familiar with tox, but now I am, a little.

I added the unit test:

https://github.com/gitaarik/babel/blob/javascript-lexer-regex-parse-fix/tests/messages/test_js_extract.py#L156-L182

I would really appreciate it if this can be merged soon and that a new release could be made. Then the build process of our project can be greatly simplified :).

gitaarik:
I'm not completely understanding this result, and what, if anything, is wrong.

michaeldg:
@akx Is what you requested done now? can this be merged?

PR #793, opened 2021-06-12
Exclude venv folder from extraction

gladykov:
venv folder is currently processed. Added as excluded folder


tatarize:
This should not be hard-coded. If we are ignoring directories from extraction we should use the `.gitignore` file or require that the user specify these see: #813

PR #796, opened 2021-06-28
fix maltese plurals

Findus23:
Important disclaimer: I just came across babel (thanks for the great library btw.) and I know very little about localisation and nothing about Maltese, but I just came across this issue and thought it might be worth it to report it.

I created files with `pybabel init -i messages.pot -d . -l langcode` for a lot of languages and then tried to load them using `Translations.load()`.
But weirdly enough for `mt` I got `ValueError: invalid token in plural form: =`.
I looked at the pluralisation rule and while I know nothing about how it works, the syntax looks similar to C, so I tried to write a C program with this line pasted and that's when I noticed that the `=>` should probably be a `>=`.


akx:
Yep, this seems like it has been a typo in the commit changing that (38556ee3dfb23d22604ec726fef5e9bf1e58973a).

Thanks for the fix!

PR #808, opened 2021-09-12
Remove redundant _compat.py

hugovk:
Since dropping support for EOL Python versions in https://github.com/python-babel/babel/pull/754, we can now remove `_compat.py`.


akx:
Great, thanks!

PR #829, opened 2022-01-25
parse_locale(): upper-case variant tag to match file system

### LINKED ISSUE #814: `ca_ES_valencia` throws `FileNotFoundError` on `get_display_name()` but `ca_ES_VALENCIA` works fine
## Overview Description

The [BCP 47](https://datatracker.ietf.org/doc/html/rfc5646#section-2.1) standard recommends that lettercase should be discarded but babel does not seem to normalize the letter case of variant subtag leading to a spurious error when provided with `ca_ES_valencia` which throws `FileNotFoundError` on `get_display_name()` but `ca_ES_VALENCIA` works fine.

## Steps to Reproduce

```python
import babel
d = babel.Locale.parse('ca_ES_valencia')
d.get_display_name()
```

## Actual Results

```python
  File "[..]/site-packages/babel/core.py", line 381, in get_display_name
    retval = locale.languages.get(self.language)
  File "[..]/site-packages/babel/core.py", line 482, in languages
    return self._data['languages']
  File "[..]/site-packages/babel/core.py", line 364, in _data
  File "[..]/site-packages/babel/localedata.py", line 143, in load
    with open(filename, 'rb') as fileobj:
FileNotFoundError: [Errno 2] No such file or directory: '[..]/site-packages/babel/locale-data/ca_ES_valencia.dat'
```

## Expected Results

> `'català (Espanya, valencià)'`

As returned when using `ca_ES_VALENCIA`:

```python
import babel
d = babel.Locale.parse('ca_ES_VALENCIA')
d.get_display_name()
```

> `'català (Espanya, valencià)'`

Should find the correct file, or throw `babel.core.UnknownLocaleError` but never `FileNotFoundError`.

## Reproducibility

babel 2.9.1

## Additional Information

Please not that the language codes are (correctly) case-insensitive:

```python
data = babel.Locale.parse('ca_es')
data.get_display_name()
```
> `'català (Espanya)'`

```python
>>> data = babel.Locale.parse('ca_ES')
>>> data.get_display_name()
```
> `'català (Espanya)'`

akx:
Upper-casing the variant tag should be okay, since quoth [the spec](https://datatracker.ietf.org/doc/html/rfc5646#section-2.1.1):

>   At all times, language tags and their subtags, including private use
>  and extensions, are to be treated as case insensitive: there exist
>  conventions for the capitalization of some of the subtags, but these
>  MUST NOT be taken to carry meaning.

Fixes #814

### DIFF
--- a/babel/core.py
+++ b/babel/core.py
@@ -1048,6 +1048,12 @@ def parse_locale(identifier, sep='_'):
     ('zh', 'CN', None, None)
     >>> parse_locale('zh_Hans_CN')
     ('zh', 'CN', 'Hans', None)
+    >>> parse_locale('ca_es_valencia')
+    ('ca', 'ES', None, 'VALENCIA')
+    >>> parse_locale('en_150')
+    ('en', '150', None, None)
+    >>> parse_locale('en_us_posix')
+    ('en', 'US', None, 'POSIX')
 
     The default component separator is "_", but a different separator can be
     specified using the `sep` parameter:
@@ -1107,7 +1113,7 @@ def parse_locale(identifier, sep='_'):
     if parts:
         if len(parts[0]) == 4 and parts[0][0].isdigit() or \
                 len(parts[0]) >= 5 and parts[0][0].isalpha():
-            variant = parts.pop()
+            variant = parts.pop().upper()
 
     if parts:
         raise ValueError('%r is not a valid locale identifier' % identifier)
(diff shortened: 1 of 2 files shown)


akx:
@krassowski Does this look like a good approach to you?

krassowski:
Thank you, this would fix my issue. Is it guaranteed that future version of CLDR will follow the same naming (using capitals?).

For even better UX (preserving the capitalisation) one could consider rewriting the `load()` method to read the list of files in the `locale-data` directory first and then find the appropriate casing of the relevant `.dat` file (but I am not sure if it is worth the effort).

akx:
I'd imagine they won't change the casing anytime soon - maybe I should add a check for that in #826 anyway.

The loading code will need reworking later anyway (there are nebulous plans for a "Loader" interface in Babel 3, whenever that is).

Issue #824, opened 2022-01-03
Distutils is deprecated and will be remove in python 3.12


RomainBrault:
Distutils is used in many places.

Is there any hope that Babel will work with python 3.12?

See:

- https://docs.python.org/fr/3/library/distutils.html
- https://www.python.org/dev/peps/pep-0632/

akx:
Thanks for the heads-up.

Yes, Babel will likely work with Python 3.12 – we'll likely excise the `distutils` compatibility stuff by then.

RosanneZe:
Is this the same issue that causes the `SetuptoolsDeprecationWarning` when using the distutils/setuptools integration? I looked into using the cli instead of the integration, but that doesn't seem to pick up the configuration options I now have in a setup.cfg file (`[extract_messages]`, `[init_catalog]` etc).

akx:
@RosanneZe What warning is that in particular?

RosanneZe:
@akx This warning:
`.venv/lib/python3.8/site-packages/setuptools/installer.py:27: SetuptoolsDeprecationWarning: setuptools.installer is deprecated. Requirements should be satisfied by a PEP 517 installer.`

akx:
@RosanneZe That seems like an unrelated issue. Can you maybe open a new issue about it with a better repro case?

akx:
With #835, Setuptools is used as the first preference, and distutils is the fallback. 

A future version of Babel will likely drop the fallbacks.

Thanks for the heads-up, @RomainBrault!

PR #831, opened 2022-01-26
Provide a way of checking if the catalogs are up-to-date

kjagiello:
I was missing a way to, as a part of CI-pipeline, check if the translations files are up to date. This PRs attempts to implement it in form of a `--check` flag for `pybabel update`.

### DIFF
--- a/babel/messages/catalog.py
+++ b/babel/messages/catalog.py
@@ -139,6 +139,13 @@ class Message(object):
     def __ne__(self, other):
         return self.__cmp__(other) != 0
 
+    def is_identical(self, other):
+        """Checks whether messages are identical, taking into account all
+        properties.
+        """
+        assert isinstance(other, Message)
+        return self.__dict__ == other.__dict__
+
     def clone(self):
         return Message(*map(copy, (self.id, self.string, self.locations,
                                    self.flags, self.auto_comments,
@@ -837,3 +844,19 @@ class Catalog(object):
         if context is not None:
             key = (key, context)
         return key
+
+    def is_identical(self, other):
+        """Checks if catalogs are identical, taking into account messages and
+        headers.
+        """
+        assert isinstance(other, Catalog)
+        for key in self._messages.keys() | other._messages.keys():
+            message_1 = self.get(key)
+            message_2 = other.get(key)
+            if (
+                message_1 is None
+                or message_2 is None
+                or not message_1.is_identical(message_2)
+            ):
+                return False
+        return dict(self.mime_headers) == dict(other.mime_headers)
--- a/babel/messages/frontend.py
+++ b/babel/messages/frontend.py
@@ -41,14 +41,15 @@ try:
     distutils_log = log  # "distutils.log → (no replacement yet)"
 
     try:
-        from setuptools.errors import OptionError, SetupError
+        from setuptools.errors import OptionError, SetupError, BaseError
     except ImportError:  # Error aliases only added in setuptools 59 (2021-11).
-        OptionError = SetupError = Exception
+        OptionError = SetupError = BaseError = Exception
 
 except ImportError:
     from distutils import log as distutils_log
     from distutils.cmd import Command as _Command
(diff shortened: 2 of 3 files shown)


babel/messages/catalog.py
```diff
@@ -840,3 +855,16 @@ def _key_for(self, id, context=None):
         if context is not None:
             key = (key, context)
         return key
+
+    def is_eq(self, other):
```
kjagiello, line 859: Not really happy with having these `is_eq`, but haven't come up with anything better so far.

babel/messages/frontend.py
```diff
@@ -801,6 +821,18 @@ def run(self):
...
                 os.remove(tmpname)
 
+        if self.check:
+            for filename, up_to_date in check_status.items():
+                if up_to_date:
+                    self.log.info('Catalog %s is up to date.', filename)
+                else:
+                    self.log.warning('Catalog %s is out of date.', filename)
+            if not all(check_status.values()):
+                raise DistutilsError("Some catalogs are out of date.")
```
kjagiello, line 831: This results unfortunately in a traceback presented to the user. There seems to be no established way of providing an exit code to the user at the moment. Let me know if you want me to look into it.

akx: That's probably fine – #824 and all, we'll probably ditch the Distutils-ness compat sooner or later.

babel/messages/catalog.py
akx: Maybe a better name (as mused upon around the other `is_eq`) (and a docstring would be nice)...

akx (changes requested):
Thanks for the contribution, seems like a good addition. Some comments around naming and so on...

babel/messages/catalog.py
```diff
@@ -142,6 +142,21 @@ def __eq__(self, other):
...
+                obj.string,
+                obj.locations,
+                obj.flags,
+                obj.auto_comments,
+                obj.user_comments,
+                obj.previous_id,
+                obj.lineno,
+                obj.context,
+            )
+        return values_to_compare(self) == values_to_compare(other)
```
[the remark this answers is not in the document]
kjagiello: I was a bit hesitant about using `__dict__`, in case there would be any private properties or such that would cause problems, but looking at it now again it looks like we should be good.

babel/messages/catalog.py
```diff
@@ -840,3 +855,16 @@ def _key_for(self, id, context=None):
         if context is not None:
             key = (key, context)
         return key
+
+    def is_eq(self, other):
+        assert isinstance(other, Catalog)
+        for key in self._messages.keys() | other._messages.keys():
+            message_1 = self[key]
+            message_2 = other[key]
```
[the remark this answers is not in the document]
kjagiello: Fair point!

babel/messages/catalog.py
```diff
@@ -142,6 +142,13 @@ def __eq__(self, other):
     def __ne__(self, other):
         return self.__cmp__(other) != 0
 
+    def is_identical(self, other):
+        """Checks whether messages are identical, taking into account all
+        properties.
+        """
+        assert isinstance(other, Message)
+        return self.__dict__ == self.__dict__
```
akx, line 150:
```suggestion
        return self.__dict__ == other.__dict__
```
😅

akx (changes requested):
Typo in `is_identical`... :)

babel/messages/catalog.py
kjagiello: Haha oopsie... Thanks for being thorough! I have added a regression test for this.

kjagiello:
@akx Let me know if there is anything else that needs to be taken care of before we can merge this PR

akx (changes requested):
Sorry for the delay! Could you rebase this?

kjagiello:
Sure thing, done!

akx:
@kjagiello Tests are unhappy 😭

kjagiello:
This will hopefully do the trick!

akx (approved):
LGTM!

Issue #124, opened 2014-11-25
Excluding files from message extraction is under-documented (ignore)


rbu:
I was trying to exclude a folder from message extraction through an ignore rule in the mapping file, and ended up having to read the source code to find out how to do it. Some of the questions that are not answered in the documentation are:
- How to specify the ignore filter (A: [ignore: something/**.txt])
- How does Babel evaluate the priority in the mapping? (A: By first match, so ignore rules need to be listed first)
- What is the path that is matched against? (A: The file path _without the package name_, see #125)

rlam3:
Is there a way for us to ignore the entire contrib folder but at the same time have [python: **.py] not look there either? If I have the following tree:

-app
--mod1
--static
----js
----css
----node_modules

Id like to ignore the entire node_modules directory and not have

python:**.py check that directory. Is there a way for this? 

The closest I got was:
```
[ignore: static/node_modules/]
[jinja2: templates/**/**.html]
encoding = utf-8
extensions=jinja2.ext.autoescape,jinja2.ext.with_
[python: **.py]
```

However jinja2 and python are always checking into static even though it is ignored..

m-aciek:
Note: since 2.10 `extract` accepts `--ignore-dirs` option with space-separated glob patterns.

I've added a PR #963 with adding them to the documentation.

NewComer00:
For project using `pyproject.toml` as the configuration file, here is an example to exclude files:

```toml
[tool.babel]
[[tool.babel.mappings]]
method = "ignore"
pattern = ["tests/**.py", "dist/**.py", "build/**.py"]

[[tool.babel.mappings]]
method = "python"
pattern = "**.py"
encoding = "utf-8"
keywords = ["_", "_l"]
```

PR #832, opened 2022-01-27
Implement directory filter for extract

### LINKED ISSUE #53: messages not extracted from folders starting with _ or .

### LINKED ISSUE #402: Add `--ignore-dirs` or similar to `extract`

akx:
This PR:

* adds a `directory_filter` callback argument to `extract_from_dir` that defaults to the current behavior of ignoring dot and underscore directories
* adds a frontend `--ignore-dirs` (that can be repeated) to replace the default `.*` / `_*` filtering with other patterns

---

Fixes #53 – set `--ignore-dirs` to disable the default dot and underscore filters
Fixes #402 – implements it :)

Refs #124 – there is still no way to set this in the mapping configuration file, though (that's #694)
Refs #253 – one could now `--ignore-dirs=node_modules`, for instance.
Refs #694 – could be reimplemented on top of this `directory_filter`?

Closes #447 – supersedes it
Closes #563 – supersedes it
Closes #761 – you can set `--ignore-dirs='.*'`
Closes #813 – supersedes it (thanks for the wording for the CLI parameter!)
Closes #793 – you can set `--ignore-dirs='.* _* venv'`

### DIFF
--- a/babel/messages/extract.py
+++ b/babel/messages/extract.py
@@ -60,9 +60,22 @@ def _strip_comment_tags(comments, tags):
     comments[:] = map(_strip, comments)
 
 
-def extract_from_dir(dirname=None, method_map=DEFAULT_MAPPING,
-                     options_map=None, keywords=DEFAULT_KEYWORDS,
-                     comment_tags=(), callback=None, strip_comment_tags=False):
+def default_directory_filter(dirpath):
+    subdir = os.path.basename(dirpath)
+    # Legacy default behavior: ignore dot and underscore directories
+    return not (subdir.startswith('.') or subdir.startswith('_'))
+
+
+def extract_from_dir(
+    dirname=None,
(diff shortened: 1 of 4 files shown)


akx:
@kinshukdua Can I ask you to review this (as the last person to attempt implementing this in #813)? :) Just so I haven't made any extra silly mistakes here...

kinshukdua:
@akx Yes sure, I'll have a look at it!

kinshukdua (approved):
LGTM!

PR #837, opened 2022-01-31
Fix duplicate locations when writing without lineno

ljodal:
If the same translation appears multiple times in the same file,
duplicate locations would be written to the .po file when using
`write_po(..., include_lineno=False)`.


akx (approved):
Good catch, thanks!

Issue #759, opened 2020-11-27
Format strings not marked correctly


youtux:
This is a problem affecting the extraction of string from python files.

While format strings like `"Hello %(name)s"` are correctly extracted and marked with the `#, python-format` marker, strings like `"Hello {name}"` are not marked, and translators may not realize that and change the formatting part, resulting in KeyErrors when actually evaluating the string.
`xgettext` will mark them correctly, with the `#, python-brace-format` marker.

# How to reproduce

```python
# asd.py
_("old style hello %(name)s")

_("new style hello {name}")
```

Run pybabel extraction:
```
$ pybabel extract asd.py -o asd.babel.po
extracting messages from asd.py
writing PO template file to asd.po
```

The `asd.babel.po` contains these entries:
```
...
#: asd.py:1
#, python-format
msgid "old style hello %(name)s"
msgstr ""

#: asd.py:3
msgid "new style hello {name}"
msgstr ""
```

The last entry is not marked as a format string.

Run the following command to check what instead `xgettext` does:
```
$ xgettext -L Python --keyword=_ asd.py -o - > asd.xgettext.po
```

The `asd.xgettext.po` contains these entries:
```
#: asd.py:1
#, python-format
msgid "old style hello %(name)s"
msgstr ""

#: asd.py:3
#, python-brace-format
msgid "new style hello {name}"
msgstr ""
```
Note that the last string is marked as `#, python-brace-format`.

# Use case
I have this problem when automatically translating using podebug to generate [catchy unicode strings](http://docs.translatehouse.org/projects/translate-toolkit/en/latest/commands/option_rewrite.html#unicode) for development purposes. The tool will not see that `"hello {name}"` is a format string, and it will translate it to `"ħḗŀŀǿ {ƞȧḿḗ}"`, causing a KeyError when calling `_("hello {name}").format(name="Alessio")`.
This can be a problem when using tools like Transifex that protect you from touching part of the string that you shouldn't. If they are not marked correctly, the tool can't do much.

ronanpaixao:
Also, even if marking the .po file manually, `pybabel compile` doesn't handle the `#, python-brace-format` properly:

```
(...)
#: somefile.py:1363
#, python-brace-format
msgid "Percentil 95% da Validação 1: {}"
msgstr "95% percentile of Validation 1: {}"
```

`pybabel compile -d . -i translations.po -l en -f --statistics`

```
122 of 122 messages (100%) translated in translations.po
[...]

PR #838, opened 2022-01-31
Allow extractors to provide message flags

### LINKED ISSUE #35: python-format flag incorrectly forced on all extractors

ljodal:
This adds a sixth value to the tuple returned by the extractor functions
which should be a set of flags.

Which flags should be applied to a message should be determined by the
extractor, as it depends on e.g. the file format being parsed. For
example "%s" should have the python-format flag if it was parsed from a
Python file and the c-format flag if it was extracted from a C file.

The logic of detecting python-format flags is also moved to the Python
extractor in this PR.

NOTE:

This is partially a breaking change. Backwards compability is maintained
with extractors that return 5-tuples instead of 6-tuples, but the
interface Babel exposes for extracting messages always returns 6-tuples.
I don't see a good way around this.

Removing the python-format detection from Message.__init__ is also a
breaking change, but that could potentially be put back for now.

Fixes #35

### DIFF
--- a/babel/messages/catalog.py
+++ b/babel/messages/catalog.py
@@ -27,18 +27,6 @@ from babel.util import distinct, LOCALTZ, FixedOffsetTimezone, _cmp
 __all__ = ['Message', 'Catalog', 'TranslationError']
 
 
-PYTHON_FORMAT = re.compile(r'''
-    \%
-        (?:\(([\w]*)\))?
-        (
-            [-#0\ +]?(?:\*|[\d]+)?
-            (?:\.(?:\*|[\d]+))?
-            [hlL]?
-        )
-        ([diouxXeEfFgGcrs%])
-''', re.VERBOSE)
-
-
 def _parse_datetime_header(value):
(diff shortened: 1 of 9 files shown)


ljodal:
Let me know if this PR has any future and if so I'll clean it up, fix the last remaining tests and add a few new tests.

Based on this I also have working code fixing #759, support for the python-brace-format flag :)

chuckyblack:
I need this feature for custom extractor. 
Without it the custom extractor cannot set flags.
The older version of babel had this feature, but it was removed.

stianjensen:
This feature would still be great!

Issue #836, opened 2022-01-31
Missing trailing semicolon in Plural-Forms header entry


QuentinFchx:
## Overview Description
Hello,

According to the [spec](https://www.gnu.org/software/gettext/manual/html_node/Plural-forms.html#Plural-forms) it seems a trailing semicolon is missing from the current ["Plural-Forms" catalog header](https://github.com/python-babel/babel/blob/9d6803a2700df9228f0b6a9614ec7a44a3ad27bb/babel/messages/catalog.py#L477)

```
Plural-Forms: nplurals=2; plural=(n != 1)
vs
Plural-Forms: nplurals=2; plural=(n != 1);
```

The code that generates the header is located [here](https://github.com/python-babel/babel/blob/9d6803a2700df9228f0b6a9614ec7a44a3ad27bb/babel/messages/catalog.py#L532)

This is breaking tools such as [gettext.js](https://github.com/guillaumepotier/gettext.js/) ([which relies on the presence of a trailing semicolon](https://github.com/guillaumepotier/gettext.js/blob/4358fdc6913055864e60177e1072d46be91c3b56/lib/gettext.js#L102)).
I understand that the tools must adapt, but in this case it seems that the issue comes from the generated format.

## Steps to Reproduce

1. Create a catalog for a locale with a "Plural-forms" header (tested with babel 2.9.1)

## Actual Results

`Plural-Forms: nplurals=2; plural=(n != 1)`

## Expected Results

`Plural-Forms: nplurals=2; plural=(n != 1);`

akx:
I think it would be prudent for `gettext.js` to also add a semicolon before that return; an extra semicolon in the JavaScript function body doesn't cause any issues, but would be compatible with Babel's `Plural-Forms` generation.

In any case, a PR is welcome to add the missing semicolon! :)

PR #841, opened 2022-02-14
catalog.py: Fixing header by adding semicolon

harshi1122:
catalog.py: Fixing header by adding semicolon 
Closes https://github.com/python-babel/babel/issues/836

I hope this solves it. I am new to this and would love guidance, if any


akx (changes requested):
Adding a semicolon to the end of the statement doesn't change the output.

PR #842, opened 2022-02-20
Adapt parse_date to handle ISO dates

ericzolf:
As the subject says: the changes should allow for transparent parsing of ISO dates, I added corresponding tests

### DIFF
--- a/babel/dates.py
+++ b/babel/dates.py
@@ -1179,12 +1179,17 @@ class ParseError(ValueError):
 def parse_date(string, locale=LC_TIME, format='medium'):
     """Parse a date from a string.
 
-    This function uses the date format for the locale as a hint to determine
-    the order in which the date fields appear in the string.
+    This function first tries to interpret the string as ISO-8601
+    date format, then uses the date format for the locale as a hint to
+    determine the order in which the date fields appear in the string.
 
     >>> parse_date('4/1/04', locale='en_US')
     datetime.date(2004, 4, 1)
     >>> parse_date('01.04.2004', locale='de_DE')
+    datetime.date(2004, 4, 1)
+    >>> parse_date('2004-04-01', locale='en_US')
+    datetime.date(2004, 4, 1)
+    >>> parse_date('2004-04-01', locale='de_DE')
     datetime.date(2004, 4, 1)
 
     :param string: the string containing the date
@@ -1195,7 +1200,16 @@ def parse_date(string, locale=LC_TIME, format='medium'):
     if not numbers:
         raise ParseError("No numbers were found in input")
 
-    # TODO: try ISO format first?
+    # we try ISO-8601 format first, meaning similar to formats
+    # extended YYYY-MM-DD or basic YYYYMMDD
+    iso_alike = re.match(r'^(\d{4})-?([01]\d)-?([0-3]\d)$',
+                         string, flags=re.ASCII)  # allow only ASCII digits
+    if iso_alike:
+        try:
+            return date(*map(int, iso_alike.groups()))
+        except ValueError:
+            pass  # a locale format might fit better, so let's continue
+
     format_str = get_date_format(format=format, locale=locale).pattern.lower()
     year_idx = format_str.index('y')
     month_idx = format_str.index('m')


babel/dates.py
```diff
@@ -1177,12 +1181,19 @@ def parse_date(string, locale=LC_TIME, format='medium'):
...
     year = numbers[indexes['Y']]
+    day = numbers[indexes['D']]
+    month = numbers[indexes['M']]
     if len(year) == 2:
-        year = 2000 + int(year)
-    else:
-        year = int(year)
-    month = int(numbers[indexes['M']])
-    day = int(numbers[indexes['D']])
+        # check if we don't have an ISO kind of format
```
akx, line 1187: There's a comment on line 1168 with "TODO: try ISO format first" – maybe this logic should be around there. Also, I'm not sure it makes much sense to read any of the indexes from a localized format for the very much standardized ISO order...

akx (changes requested):

babel/dates.py
ericzolf: > There's a comment on line 1168 with "TODO: try ISO format first" – maybe this logic should be around there.

I can remove the comment, I just forgot.

> Also, I'm not sure it makes much sense to read any of the indexes from a localized format for the very much standardized ISO order...

I thought about this but then I thought that it's probably best to be more flexible and accept also formats which only look like ISO e.g. with slashes as separator rather than dashes.
This said, I don't care too much, I do use properly ISO :-) Just let me know.

ericzolf: OK, I pushed a new commit with a more explicit handling of the ISO format at the beginning of the function. Let me know which one you prefer.

babel/dates.py
```diff
@@ -1161,7 +1165,12 @@ def parse_date(string, locale=LC_TIME, format='medium'):
     if not numbers:
         raise ParseError("No numbers were found in input")
 
-    # TODO: try ISO format first?
+    # we try ISO format first YYYY-MM-DD
+    if len(numbers) == 3:
+        ints = list(map(int, numbers))
+        if ints[0] > 31 and ints[1] <= 12 and ints[2] <= 31:
+            return date(*ints)
```
akx, line 1172: I think this should be wrapped in a try-expect. If the atoms aren't interpretible as integers, this would raise an exception. Additionally since the format is `YYYY`, shouldn't `ints[0] > 31` be `ints[0] >= 1000`? At that point you could also do something like

```
if ints[0] >= 1000 and 1 <= ints[1] <= 12 and 1 <= ints[2] <= 31
```
for better validation.

akx (changes requested):

babel/dates.py
ericzolf: I think, the issue is that we're not aligned on the exact purpose of the code: I was trying to identify everything looking "ISO-ish" without trying to be aligned exactly with the specs. In this perspective, I still think that my code is correct (ISO-ish being a year followed by a month followed by a day). So let me know if I rather should aim at ISO-strictness, where indeed my code would need to look very different.
PS: all my comments will appear at once because I wrongly started a review...

babel/dates.py
```diff
@@ -1161,7 +1165,12 @@ def parse_date(string, locale=LC_TIME, format='medium'):
     if not numbers:
         raise ParseError("No numbers were found in input")
 
-    # TODO: try ISO format first?
+    # we try ISO-alike format first, meaning similar to YYYY-MM-DD
+    if len(numbers) == 3:
```
akx, line 1169: I think we should at the very least check that the first number-string is 4-digited:
```suggestion
    if len(numbers) == 3 and len(numbers[0]) == 4:
```
or maybe even more specifically
```suggestion
    if [len(n) for n in numbers] == [4, 2, 2]:  # 3 numbers, "ISO-shaped"
```
...

babel/dates.py
```diff
@@ -1161,7 +1165,12 @@ def parse_date(string, locale=LC_TIME, format='medium'):
...
+        ints = list(map(int, numbers))
+        if ints[0] > 31 and ints[1] <= 12 and ints[2] <= 31:
```
akx, line 1171: I think this could use a comment illuminating the idea behind this check in any case.

```suggestion
        # Only assume the date is an ISO 8601 date if
        # * the "year" component is >= 31 (so it can't be a day-of-month)
        # * the "month" component is in range
        # * the "day" component is in range
        if ints[0] > 31 and 1 <= ints[1] <= 12 and 1 <= ints[2] <= 31:
```

akx (changes requested):
Beyond the inline comments, the docstring for the function should also be augmented to note that despite the `locale` hint given, it will attempt ISO 8601 -style parsing initially (which, honestly, feels a bit wrong but it's nevertheless probably practical)...

babel/dates.py
ericzolf: The new regex and comment should make everything clear

babel/dates.py
ericzolf: I replaced the whole thing through a single pattern matching, and else let datetime.date function do the heavy lifting validation. This is shorter, more expressive, and allow me to support as well the basic YYYYMMDD format of ISO.

ericzolf:
I think, I've resolved all points. At the end, the ISO-validation is stricter but it also allows to support the basic ISO format YYYYMMDD (without dashes), so I think it's a good deal.

babel/dates.py
```diff
@@ -1161,7 +1166,15 @@ def parse_date(string, locale=LC_TIME, format='medium'):
     if not numbers:
         raise ParseError("No numbers were found in input")
 
-    # TODO: try ISO format first?
+    # we try ISO-8601 format first, meaning similar to formats
+    # extended YYYY-MM-DD or basic YYYYMMDD
+    iso_alike = re.match(r'^(\d{4})-?([01]\d)-?([0-3]\d)$', string)
```
jun66j5, line 1171: This pattern matches string with non-ascii characters. I think we should accept only-ascii string by adding `re.ASCII` or using `[0-9]` instead of `\d`.

```pycon
>>> string = '২০২৩-1২-3১'
>>> re.match(r'^(\d{4})-?([01]\d)-?([0-3]\d)$', string)
<re.Match object; span=(0, 10), match='২০২৩-1২-3১'>
```

ericzolf: I added re.ASCII and rebased/squashed the whole PR.

akx (approved):
Looks good. Sorry for the delay, and again thanks to @jun66j5 for comments & review!

PR #843, opened 2022-02-23
Prefer setuptools imports to distutils imports

akx:
Considering distutils is being deprecated, this fixes up things so imports from `setuptools` are used where available instead.

The non-conditional imports have been around for 6 to 17 years, so they should be safe in conservative situations too.

Refs #824


akx:
For people working around e.g. Fedora packaging (@FelixSchwarz, @kloczek?), do you foresee this change causing trouble?

FelixSchwarz:
Fedora has `setuptools` available so I don't think this will be an issue. In general Fedora seems to be pretty much at  the (b)leading edge of Python versioning so if something is deprecated upstream, chances are pretty good that Fedora's Python packaging (in general) was already adapted to support the new approach.

PR #845, opened 2022-02-26
Fix output of --list-locales to not use bytes reprs

morganwahl:
This adds a unit test for the `--list-locales` argument, and updates its implementation to not output bytes reprs.

In other words, goes from this:

```
...
b'fr_CH           French (Switzerland)'
b'fr_CI           French (C\xc3\xb4te d\xe2\x80\x99Ivoire)'
b'fr_CM           French (Cameroon)'
...
```

to this

```
...
fr_CH           French (Switzerland)
fr_CI           French (Côte d’Ivoire)
fr_CM           French (Cameroon)
...
```

### DIFF
--- a/babel/messages/frontend.py
+++ b/babel/messages/frontend.py
@@ -22,7 +22,6 @@ from collections import OrderedDict
 from configparser import RawConfigParser
 from datetime import datetime
 from io import StringIO
-from locale import getpreferredencoding
 
 from babel import __version__ as VERSION
 from babel import Locale, localedata
@@ -906,10 +905,7 @@ class CommandLineInterface(object):
             format = u'%%-%ds %%s' % (longest + 1)
             for identifier in identifiers:
                 locale = Locale.parse(identifier)
-                output = format % (identifier, locale.english_name)
-                print(output.encode(sys.stdout.encoding or
-                                    getpreferredencoding() or
-                                    'ascii', 'replace'))
+                print(format % (identifier, locale.english_name))
             return 0
 
         if not args:
--- a/tests/messages/test_frontend.py
+++ b/tests/messages/test_frontend.py
@@ -756,6 +756,17 @@ usage: pybabel command [options] [args]
 pybabel: error: no valid command or option passed. try the -h/--help option for more information.
 """, sys.stderr.getvalue().lower())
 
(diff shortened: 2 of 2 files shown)


babel/messages/frontend.py
```diff
@@ -893,9 +893,16 @@ def run(self, argv=None):
...
+                # Prevent errors on printing by doing an encode/decode
+                # roundtrip with errors='replace'.
+                encoding = (
+                    sys.stdout.encoding
+                    or getpreferredencoding()
+                    or 'ascii'
+                )
+                print(
+                    output.encode(encoding, errors='replace').decode(encoding)
+                )
```
akx, line 905: I think in these Py3-only days where we also have `PYTHONUTF8` for Windows, this could just be

```suggestion
                print(output)
```

tests/messages/test_frontend.py
```diff
@@ -761,6 +761,19 @@ def test_usage(self):
 pybabel: error: no valid command or option passed. try the -h/--help option for more information.
 """, sys.stderr.getvalue().lower())
 
+    def test_list_locales(self):
+        """
+        Test the command with the --list-locales arg.
+        """
+        result = self.cli.run(sys.argv + ['--list-locales'])
+        self.assertEqual(0, result)
```
akx, line 769: Prefer regular assertions in new tests:
```suggestion
        assert result == 0
```

tests/messages/test_frontend.py
```diff
@@ -761,6 +761,19 @@ def test_usage(self):
...
+        output = sys.stdout.getvalue()
+        for line in output.splitlines():
+            self.assertRegex(
+                line,
+                r"[0-9A-Za-z_]{2,}\s+[0-9A-Za-z,&\\\-\.() ]+",
+            )
```
akx, line 775: I think we don't need to test this this strictly. How about something like
```suggestion
        output = sys.stdout.getvalue()
        # Known locale occurs in output
        assert 'fr_CH' in output
        assert 'French (Switzerland)' not in output
        # No bytes repr markers in output
        assert "\nb'" not in output
```

akx (approved):
I folded in my review comments. Thank you for the contribution!

PR #847, opened 2022-03-12
bug fix with doctest

xelA-S:
#836 bug fix tested using doctest


akx:
Thank you! This was superseded by #848, which has correct test coverage too.

PR #848, opened 2022-03-28
Plural-Forms: Fix missing trailing semicolon

farhan5900:
Adds missing semicolon in the code that generates the header that is `Catalog.plural_forms` as well as in `plural.to_gettext` function. Also modifies all the concerning test cases as well as test data files.

Closes https://github.com/python-babel/babel/issues/836


akx (approved):
Looks good, thank you!

Issue #866, opened 2022-04-27
"B" symbol in pattern of dayperiod is not supported


jun66j5:
## Overview Description

(Originally reported in https://trac.edgewall.org/ticket/13482, running unit tests in Trac with Babel 2.10.1)

Since Babel 2.10.0 (CLDR 41), `B` symbol is used in datetime pattern for `zh_Hant` locale, however Babel doesn't support the `B` symbol.

## Steps to Reproduce

```
>>> from datetime import datetime
>>> import pytz
>>> import babel
>>> from babel import dates
>>> babel.__version__
'2.10.1'
>>> dates.get_time_format('medium', 'zh_Hant')
<DateTimePattern 'Bh:mm:ss'>
>>> t1 = datetime(2010, 8, 28, 1, 45, 56, 0, pytz.UTC)
>>> t2 = datetime(2010, 8, 28, 13, 45, 56, 0, pytz.UTC)
>>> s1 = dates.format_datetime(t1, 'medium', t1.tzinfo, 'zh_Hant')
>>> s1
'2010年8月28日 B1:45:56'
>>> s2 = dates.format_datetime(t2, 'medium', t1.tzinfo, 'zh_Hant')
>>> s2
'2010年8月28日 B1:45:56'
>>> t1 == t2
False
>>> s1 == s2
True
```

## Actual Results

The `B` symbol is not formatted.

## Expected Results

The `B` symbol should be formatted to `上午` (am) or `下午` (pm).

## Reproducibility

100%

## Additional Information

n/a

jun66j5:
Only `zh_Hant` and `zh_Hant_TW` use `B` symbol for time format in Babel 2.10.1 (CLDR 41).

```pycon
>>> import babel
>>> babel.__version__
'2.10.1'
>>> from babel.localedata import locale_identifiers
>>> from babel import dates
>>> for locale in locale_identifiers():
...   for format_ in formats:
...     pattern = dates.get_time_format(format=format_, locale=locale)
...     if 'b' in pattern.pattern or 'B' in pattern.pattern:
...       print(locale, width, pattern)
...
zh_Hant_TW full Bh:mm:ss [zzzz]
zh_Hant_TW full Bh:mm:ss [z]
zh_Hant_TW full Bh:mm:ss
zh_Hant_TW full Bh:mm
zh_Hant full Bh:mm:ss [zzzz]
zh_Hant full Bh:mm:ss [z]
zh_Hant full Bh:mm:ss
[...]

PR #869, opened 2022-04-29
Add support for `b` and `B` period symbols in time format

jun66j5:
Fix for #866

### DIFF
--- a/babel/dates.py
+++ b/babel/dates.py
@@ -1456,23 +1456,23 @@ class DateTimeFormat(object):
         """
         Return period from parsed datetime according to format pattern.
 
-        >>> format = DateTimeFormat(time(19, 42), 'en_US')
+        >>> format = DateTimeFormat(time(13, 42), 'fi_FI')
         >>> format.format_period('a', 1)
-        u'PM'
-
+        u'ip.'
         >>> format.format_period('b', 1)
-        u'evening'
-
-        >>> format.format_period('B', 1)
-        u'in the evening'
+        u'iltap.'
+        >>> format.format_period('b', 4)
+        u'iltapäivä'
+        >>> format.format_period('B', 4)
+        u'iltapäivällä'
+        >>> format.format_period('B', 5)
+        u'ip.'
 
         >>> format = DateTimeFormat(datetime(2022, 4, 28, 6, 27), 'zh_Hant')
         >>> format.format_period('a', 1)
         u'上午'
-
         >>> format.format_period('b', 1)
         u'清晨'
-
         >>> format.format_period('B', 1)
         u'清晨'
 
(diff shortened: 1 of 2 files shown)


babel/dates.py
```diff
@@ -1453,10 +1452,44 @@ def format_day_of_year(self, num):
...
+
+        >>> format.format_period('B', 1)
+        u'清晨'
+
+        :param char: pattern format character ('a', 'b', 'B')
+        :param num: count of format character
+
+        """
+        widths = [{3: 'abbreviated', 4: 'wide', 5: 'narrow'}[max(3, num)],
+                  'wide', 'narrow', 'abbreviated']
```
akx, line 1484: Just so I'm following, the idea here is to take a preferred width according to `num`, followed by fallbacks?

🤔  Passing in a `num` >= 6 will crash here, but that's fine – those are out of spec anyway.

babel/dates.py
```diff
@@ -1453,10 +1452,44 @@ def format_day_of_year(self, num):
...
+        if char == 'a':
+            period = {0: 'am', 1: 'pm'}[int(self.value.hour >= 12)]
```
akx, line 1486: I know this was from the original code, but now that we're touching this segment, this should probably be changed to the more obvious
```suggestion
            period = 'pm' if int(self.value.hour >= 12) else 'am'
```

babel/dates.py
```diff
@@ -1453,10 +1452,44 @@ def format_day_of_year(self, num):
...
+
+        >>> format = DateTimeFormat(time(19, 42), 'en_US')
+        >>> format.format_period('a', 1)
+        u'PM'
+
+        >>> format.format_period('b', 1)
+        u'evening'
+
+        >>> format.format_period('B', 1)
+        u'in the evening'
```
akx, line 1467: Should there be some tests for `BBB`/`bbb` or other widths in general?

akx (changes requested):
Thank you, looks pretty good! Some minor comments within :)

babel/dates.py
jun66j5:
```suggestion
            period = 'pm' if self.value.hour >= 12 else 'am'
```
I think `int(...)` is not needed if changing the code.

babel/dates.py
jun66j5: > Just so I'm following, the idea here is to take a preferred width according to `num`, followed by fallbacks?

Yes. But the fallback is not needed if all widths of `dayPeriod` in CLDR are defined.

I'll check CLDR database.

> 🤔 Passing in a `num` >= 6 will crash here, but that's fine – those are out of spec anyway.

Other `format_*` methods have the same issue. It is defined to allow 1-5 characters for `b` and `B` in `PATTERN_CHARS`. Actually, `BBBBBB` (num: 6) is not allowed.

```pycon
>>> dates.format_time(time(13), 'BBBBBB h:mm', locale='en_US')
  File "/home/jun66j5/src/babel.git/babel/dates.py", line 833, in format_time
    return parse_pattern(format).apply(time, locale)
  File "/home/jun66j5/src/babel.git/babel/dates.py", line 1667, in parse_pattern
    raise ValueError('Invalid length for field: %r'
ValueError: Invalid length for field: 'BBBBBB'
```

akx (approved):
Thank you for the contribution and the fast review turnaround, looks great! :)

Issue #870, opened 2022-05-01
get_period_id() incorrectly handles day period rule crossing 0:00 since Babel 2.4.0


jun66j5:
## Overview Description

According to https://github.com/unicode-org/cldr/blob/release-41/common/supplemental/dayPeriods.xml#L15-L22, there is `dayPeriodRule` crossing 0:00 for `en` locale.

```xml
<dayPeriodRules locales="en">
        ...
        <dayPeriodRule type="night1" from="21:00" before="06:00"/>      <!-- night -->
</dayPeriodRules>
```

I noticed `get_period_id()` incorrectly handles this day period rule.

```pycon
>>> dates.get_period_id(time(23), locale=en_US)
'pm'  # => should be 'night1'
>>> dates.get_period_id(time(1), locale=en_US)
'am'  # => should be 'night1'
```

## Steps to Reproduce

```pycon
>>> from datetime import time
>>> from babel import Locale, dates
>>> en_US = Locale.parse('en_US')
>>> dates.get_period_id(time(23), locale=en_US)
'pm'  # => should be 'night1'
>>> dates.get_period_id(time(1), locale=en_US)
'am'  # => should be 'night1'
>>> en_US = Locale.parse('en_US')
>>> for h in range(24):
...   t = time(h)
...   print(t, dates.get_period_id(t, locale=en_US))
...
00:00:00 midnight
01:00:00 am
02:00:00 am
03:00:00 am
04:00:00 am
05:00:00 am
06:00:00 morning1
07:00:00 morning1
08:00:00 morning1
09:00:00 morning1
10:00:00 morning1
11:00:00 morning1
12:00:00 noon
13:00:00 afternoon1
14:00:00 afternoon1
15:00:00 afternoon1
16:00:00 afternoon1
17:00:00 afternoon1
18:00:00 evening1
19:00:00 evening1
20:00:00 evening1
21:00:00 pm
22:00:00 pm
23:00:00 pm
```

## Actual Results

```#!pycon
>>> dates.get_period_id(time(23), locale=en_US)
'pm'
>>> dates.get_period_id(time(1), locale=en_US)
'am''
```

## Expected Results

`dates.get_period_id(time(23), locale=en_US)` and `dates.get_period_id(time(1), locale=en_US)` should return `'night1'`.

Also, we should add unit tests for `get_period_id()`.

## Reproducibility

The issue occurs with Babel 2.4.0 and later, however doesn't with 2.3.0 - 2.3.4.

## Additional Information

[...]

PR #871, opened 2022-05-01
Fix get_period_id() with `dayPeriodRule` across 0:00

jun66j5:
Fix for #870.


akx (approved):
Looks great, thank you! I would have added a comment noting that the first `if` statement takes care of rules that straddle days, but it's evident from the commit message. 😄

Issue #868, opened 2022-04-28
format_currency throws a KeyError when using the "es_AR" locale


pedrogut:
## Overview Description
`format_currency` throws a KeyError for any multiple of 1000000 when using the "es_AR" locale.
## Steps to Reproduce
```
from babel.numbers import format_currency
format_currency(1000000, "ARS", locale="es_AR")
```

## Actual Results
```
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/home/pedro/Documents/project/venv/lib/python3.10/site-packages/babel/numbers.py", line 533, in format_currency
    return pattern.apply(
  File "/home/pedro/Documents/project/venv/lib/python3.10/site-packages/babel/numbers.py", line 1028, in apply
    get_currency_name(currency, value, locale))
  File "/home/pedro/Documents/project/venv/lib/python3.10/site-packages/babel/numbers.py", line 130, in get_currency_name
    return plural_names[currency][plural_form]
  File "/home/pedro/Documents/project/venv/lib/python3.10/site-packages/babel/localedata.py", line 238, in __getitem__
    orig = val = self._data[key]
KeyError: 'many'
```

## Expected Results
'$\xa01.000.000,00'
## Additional Information
This works fine for other values such as: 100000, 999999 and 1000001. But fails for any other multiple of 1000000 (2000000, 3000000, 4000000, etc.)

Olunusib:
I was checking around and this bug seems to be from the different ways plurals are represented in different languages. In Espanyol for example, 'many' is supported but it's not in English (only 'one' and 'other') seems to be supported for this currency (and many other currencies in the 'es' region). 

This is not a problem with currencies like RUR since they have the same plural representation for all languages.

A possible quick fix is to fallback on the ordinal value if the plural value is not supported. A possible long-term fix will be normalizing plural representation in different languages for the same currency. This heavily depends on CLDR.

jun66j5:
According to https://unicode.org/reports/tr35/tr35-numbers.html#4-currencies:

> If no matching displayName element was found in the previous step, then look for a displayName element that matches count="other", starting in the current locale and then following the locale fallback chain up to, but not including root.

However, this step is not implemented in `format_currency()`.

I posted pull request #872 for this issue.

akx:
Fixed in #872. Thank you, @jun66j5!

PR #872, opened 2022-05-07
Fallback count="other" format in format_currency()

jun66j5:
Proposed fix for #868.


akx (approved):
Looks good, thank you!

PR #877, opened 2022-05-12
Support for hex escapes in JavaScript strings

codepainters:
Currently Babel only supports `\u`-style Unicode escapes in JavaScript string literals.

Unfortunately, I have to deal witth JS files (WebPack-generated) that contain `\x`-style escapes. I believe Babel should support it, as it's perfectly legitimate JS syntax.


akx (approved):
Looks good, thank you!

codepainters:
Thank you!

PR #887, opened 2022-06-15
Fix incorrect iso week-years (`w`)

xmo-odoo:
This is related to #885 but doesn't fix the non-ISO issues, probably (not actually tested).

The issue this fixes is that while #621 fixed some cases it broke others, specifically the case where a day of year A is part of a week of year A-1, *but* A-1 has 53 weeks and A has 52.

See last commit for a complete explanation of the process.

- The first commit is just a convenience because the current `tox.ini` doesn't forward parameters to pytest which is frustrating.
- The second commit adds an exhaustive-ish parametrized test to the test suite: using a different method of computing week numbers (which I assume is correct but might not be), it checks every date within 10 days of January 1st, from year 1900 to 2100, to verify that babel formats the correct iso-week-year and iso-week.
- The last commit fixes the issue.

Note that this is only a fix (somewhat dubious) for incorrect ISO weeks. It does not fix:

- non-iso weeks, which is the issue described in #885
- month-weeks, which are almost certainly broken in all sorts of ways at least if we go by TR 35 semantics (to the extent that they're worth anything).

I figure it might be useful to expand the exhaustive test to other known locales, or even to all combinations of `(min_week_days, first_week_day)` though ensuring the oracle is correct under all conditions would be less trivial, here I was able to check no a normal calendar to see if test failures were genuine (aka the oracle was correct and babel not). I don't think there's a calendar out there where the first DOW is wednesday and the first week of the year must have 5 days or some such nonsense.


xmo-odoo:
@akx any issue?

akx:
Hi @xmo-odoo, so sorry for the delay. 

I took a look at this originally and left it to simmer in a background thread and eventually forgot. I'm pretty hesitant about a calendar-related change (because times are hard and this could probably affect someone's business).

I'd love to see a reference to some other system implementing things the same way – in fact using `isocalendar` for the tests would be a great way to do that, since I do trust Python's stdlib to have gotten ISO weeks right..?

xmo-odoo:
> I'd love to see a reference to some other system implementing things the same way – in fact using `isocalendar` for the tests would be a great way to do that, since I do trust Python's stdlib to have gotten ISO weeks right..?

The second commit should be a pretty exhaustive test for the range [1900, 2100]. Using a known and perfectly reliable computation method (first iso week is the one which contains Jan 4th) seemed like an ideal oracle for tests. Unless it's not a proper definition of ISO week and has caveats?

PR #895, opened 2022-07-29
Fix exception on iOS in _get_localzone

lotuspar:
On iOS the macOS specific part of the function just instantly throws an exception due to no systemsetup file being found

This PR just wraps that part in a try block and continues on a FileNotFoundError

### DIFF
--- a/babel/localtime/_unix.py
+++ b/babel/localtime/_unix.py
@@ -65,17 +65,21 @@ def _get_localzone(_root='/'):
     # since it knows where the zone files are that should be a bit
     # better than reimplementing the logic here.
     if sys.platform == 'darwin':
-        c = subprocess.Popen(['systemsetup', '-gettimezone'],
-                             stdout=subprocess.PIPE)
-        sys_result = c.communicate()[0]
-        c.wait()
-        tz_match = _systemconfig_tz.search(sys_result)
-        if tz_match is not None:
-            zone_name = tz_match.group(1)
-            try:
-                return pytz.timezone(zone_name)
-            except pytz.UnknownTimeZoneError:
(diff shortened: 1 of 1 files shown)


babel/localtime/_unix.py
```diff
@@ -65,17 +65,21 @@ def _get_localzone(_root='/'):
...
+            tz_match = _systemconfig_tz.search(sys_result)
+            if tz_match is not None:
+                zone_name = tz_match.group(1)
+                try:
+                    return pytz.timezone(zone_name)
+                except pytz.UnknownTimeZoneError:
+                    pass
+        # iOS doesn't come with systemsetup
+        except FileNotFoundError: 
+            pass
```
lukasjuhrich, line 82: Instead of having a huge `try`-block, you can make it more specific moving the non-critical things to the `else`-block of a try-except:
```suggestion
        try:
            c = subprocess.Popen(['systemsetup', '-gettimezone'],
                                stdout=subprocess.PIPE)
        # iOS doesn't come with systemsetup
        except FileNotFoundError: 
            pass
        else:
            sys_result = c.communicate()[0]
            c.wait()
            tz_match = _systemconfig_tz.search(sys_result)
            if tz_match is not None:
                zone_name = tz_match.group(1)
                try:
                    return pytz.timezone(zone_name)
                except pytz.UnknownTimeZoneError:
                    pass        
```
it helps identifying which part we expect to throw the exception, and avoids catching any unexpected instances of this exceptions – which probably should stay unhandled so that the bug becomes visible and one gets a nice backtrace.

_(not a maiintainer, just randomly browsed through PRs)_

PR #896, opened 2022-07-30
Remove `__nonzero__` methods

sobolevn:
`__nonzero__` is not used in python3, instead we now use `__bool__`.
Since `setup.py` declares to support `python >= 3.6`, I guess it is safe to remove these methods.

`setup.py`: https://github.com/python-babel/babel/blob/d8584405a66a79d79bed44b5d80df0bc89f8eee3/setup.py#L59


akx (approved):
Makes sense. These have just been missed when removing Python 2 support. :)

Thanks!

PR #898, opened 2022-08-09
Update Python versions in dev docs

rnestler:
Also remove mentions of Python 2 workarounds, since they aren't applicable anymore.


akx (approved):

akx:
Good catch, thanks!

PR #905, opened 2022-09-23
Remove superfluous `__unicode__` declarations

lukasjuhrich:
The `__unicode__` protocol is not used in python3, and furthermore the `unicode()` builtin does not exist anymore.

I found this while browsing through the source code and given that babel requires ≥3.6 I thought I could just open a PR.


akx:
Thanks!

PR #909, opened 2022-10-17
Add support for compact decimal formats

### LINKED ISSUE #908: Is it possible to use the short number formats from CLDR?
CLDR contains some short number formats, for example in English

1,000 can become 1K, 12,100,000 can become 12.1M.

https://github.com/unicode-org/cldr-json/blob/main/cldr-json/cldr-numbers-modern/main/en/numbers.json#L61-L88

Similarly, other languages can use this with their own characters

eg. 32,500,000 = 3250万

https://github.com/unicode-org/cldr-json/blob/main/cldr-json/cldr-numbers-modern/main/ja/numbers.json#L51-L66

Is there a method in Babel that allows us to use these formats for decimal numbers?

----

It seems that the data is loaded under "compact_decimal_formats" although there are no utility functions for using them.

As per comment @ https://github.com/python-babel/babel/blob/v2.10.3/scripts/import_cldr.py#L768-L774

```py
>>>Locale('en', 'US')._data["compact_decimal_formats"]["short"]["one"]["1000"]
<NumberPattern '0K'>
```

DenverCoder1:
Adds support for compact decimal formats.

Implementation based on details found at https://www.unicode.org/reports/tr35/tr35-45/tr35-numbers.html#Compact_Number_Formats

```py
>>> format_decimal(12345, locale='en_US', compact="short")
u'12K'
>>> format_decimal(12345, locale='en_US', compact="long")
u'12 thousand'
>>> format_decimal(12345, locale='en_US', compact="short", compact_fraction_digits=2)
u'12.35K'
>>> format_decimal(1234567, locale='ja_JP', compact="short")
u'123万'
```

Closes #908

### DIFF
--- a/babel/numbers.py
+++ b/babel/numbers.py
@@ -375,8 +375,12 @@ def get_decimal_quantum(precision):
     return decimal.Decimal(10) ** (-precision)
 
 
+class UnknownCompactFormat(KeyError):
+    """Exception raised when an unknown compact format is requested."""
+
+
 def format_decimal(
-        number, format=None, locale=LC_NUMERIC, decimal_quantization=True, group_separator=True):
+        number, format=None, locale=LC_NUMERIC, decimal_quantization=True, group_separator=True, compact=None, compact_fraction_digits=0):
     u"""Return the given decimal number formatted for a specific locale.
 
     >>> format_decimal(1.2345, locale='en_US')
@@ -408,6 +412,14 @@ def format_decimal(
     u'12345,67'
     >>> format_decimal(12345.67, locale='en_US', group_separator=True)
     u'12,345.67'
+    >>> format_decimal(12345, locale='en_US', compact="short")
+    u'12K'
+    >>> format_decimal(12345, locale='en_US', compact="long")
+    u'12 thousand'
+    >>> format_decimal(12345, locale='en_US', compact="short", compact_fraction_digits=2)
+    u'12.35K'
+    >>> format_decimal(1234567, locale='ja_JP', compact="short")
+    u'123万'
 
     :param number: the number to format
     :param format:
@@ -416,8 +428,30 @@ def format_decimal(
                                  the format pattern. Defaults to `True`.
     :param group_separator: Boolean to switch group separator on/off in a locale's
                             number format.
+    :param compact: Compact format to use ("short" or "long"). Defaults to `None`.
+    :param compact_fraction_digits: Number of fraction digits to use in compact
+                                    format. Defaults to `0`. If this is set to
+                                    a value greater than `0`, the `decimal_quantization`
+                                    will be treated as `False`.
     """
     locale = Locale.parse(locale)
+    if compact:
+        try:
+            compact_format = locale.compact_decimal_formats[compact]
+            for magnitude in sorted([int(m) for m in compact_format["other"]], reverse=True):
+                if number >= magnitude:
+                    format = compact_format["other"][str(magnitude)]
+                    pattern = parse_pattern(format).pattern
+                    if pattern != "0" and number >= 1000:
+                        number = number / (magnitude / (10 ** (pattern.count("0") - 1)))
+                    if float(number) == 1.0 and "one" in compact_format:
+                        format = compact_format["one"][str(magnitude)]
+                    if compact_fraction_digits > 0:
+                        decimal_quantization = False
+                        number = round(number, compact_fraction_digits)
+                    break
+        except KeyError as e:
+            raise UnknownCompactFormat("%r is not a known compact format" % e.args[0])        
     if not format:
         format = locale.decimal_formats.get(format)
     pattern = parse_pattern(format)
--- a/tests/test_numbers.py
+++ b/tests/test_numbers.py
@@ -121,6 +121,33 @@ class FormatDecimalTestCase(unittest.TestCase):
         assert numbers.format_currency(101299.98, 'EUR', locale='en_US', group_separator=True, format_type='name') == u'101,299.98 euros'
         assert numbers.format_percent(251234.1234, locale='sv_SE', group_separator=True) == u'25\xa0123\xa0412\xa0%'
 
+    def test_compact(self):
+        assert numbers.format_decimal(1, locale='en_US', compact="short") == u'1'
+        assert numbers.format_decimal(999, locale='en_US', compact="short") == u'999'
+        assert numbers.format_decimal(1000, locale='en_US', compact="short") == u'1K'
+        assert numbers.format_decimal(9000, locale='en_US', compact="short") == u'9K'
+        assert numbers.format_decimal(9123, locale='en_US', compact="short", compact_fraction_digits=2) == u'9.12K'
+        assert numbers.format_decimal(10000, locale='en_US', compact="short") == u'10K'
+        assert numbers.format_decimal(10000, locale='en_US', compact="short", compact_fraction_digits=2) == u'10K'
+        assert numbers.format_decimal(1000000, locale='en_US', compact="short") == u'1M'
+        assert numbers.format_decimal(9000999, locale='en_US', compact="short") == u'9M'
+        assert numbers.format_decimal(9000900099, locale='en_US', compact="short", compact_fraction_digits=5) == u'9.0009B'
+        assert numbers.format_decimal(1, locale='en_US', compact="long") == u'1'
+        assert numbers.format_decimal(999, locale='en_US', compact="long") == u'999'
+        assert numbers.format_decimal(1000, locale='en_US', compact="long") == u'1 thousand'
+        assert numbers.format_decimal(9000, locale='en_US', compact="long") == u'9 thousand'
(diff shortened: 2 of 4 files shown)


jun66j5:
If the given number is negative, the formatted text is not compact.

```
>>> numbers.format_decimal(123456789, compact='short', locale='en_US')
'123M'
>>> numbers.format_decimal(123456789, compact='short', locale='ja')
'1億'
>>> numbers.format_decimal(-123456789, compact='short', locale='en_US')
'-123,456,789'
>>> numbers.format_decimal(-123456789, compact='short', locale='ja')
'-123,456,789'
```

jun66j5:
Several locales have different formats between `count="one"` and `count="other"` (e.g. `mk`, `ps`, ...). I think the `count="one"` pattern should be handled.

```pycon
>>> from babel.core import Locale
>>> from babel.localedata import locale_identifiers
>>> def show_different_formats():
...   for identifier in sorted(locale_identifiers()):
...     locale = Locale.parse(identifier)
...     fmts = locale.compact_decimal_formats['short']
...     if 'one' in fmts and 'other' in fmts:
...       for s in sorted(set(fmts['one']) & set(fmts['other'])):
...         fmt_one = fmts['one'][s]
...         fmt_other = fmts['other'][s]
...         if fmt_one.pattern != fmt_other.pattern:
...           print(identifier, repr(s), repr(fmt_one.pattern), repr(fmt_other.pattern))
...
>>> show_different_formats()
bn '10000000000' '00\xa0শত\xa0কো' '00শত\xa0কো'
bn_BD '10000000000' '00\xa0শত\xa0কো' '00শত\xa0কো'
bn_IN '10000000000' '00\xa0শত\xa0কো' '00শত\xa0কো'
mk '100000000000' "000\xa0мј'.'" "000\xa0ми'.'"
mk_MK '100000000000' "000\xa0мј'.'" "000\xa0ми'.'"
ps '100000000000' '000G' '000B'
ps_AF '100000000000' '000G' '000B'
ps_PK '100000000000' '000G' '000B'
sw '1000000' '0M;-0M' '0M'
sw '10000000' '00M;-00M' '00M'
sw '100000000' '000M;-000M' '000M'
sw '1000000000000' '0T;-0T' '0T'
sw '10000000000000' '00T;-00T' '00T'
sw '100000000000000' '000T;-000T' '000T'
sw_CD '1000000' '0M;-0M' '0M'
sw_CD '10000000' '00M;-00M' '00M'
sw_CD '100000000' '000M;-000M' '000M'
sw_CD '1000000000000' '0T;-0T' '0T'
sw_CD '10000000000000' '00T;-00T' '00T'
sw_CD '100000000000000' '000T;-000T' '000T'
sw_TZ '1000000' '0M;-0M' '0M'
sw_TZ '10000000' '00M;-00M' '00M'
sw_TZ '100000000' '000M;-000M' '000M'
sw_TZ '1000000000000' '0T;-0T' '0T'
sw_TZ '10000000000000' '00T;-00T' '00T'
sw_TZ '100000000000000' '000T;-000T' '000T'
sw_UG '1000000' '0M;-0M' '0M'
sw_UG '10000000' '00M;-00M' '00M'
sw_UG '100000000' '000M;-000M' '000M'
sw_UG '1000000000000' '0T;-0T' '0T'
sw_UG '10000000000000' '00T;-00T' '00T'
sw_UG '100000000000000' '000T;-000T' '000T'
```

DenverCoder1:
> Several locales have different formats between `count="one"` and `count="other"` (e.g. `mk`, `ps`, ...). I think the `count="one"` pattern should be handled.

Unless I understood incorrectly, I believe it does already when number is 1 after dividing.

DenverCoder1:
In the latest commit, the parameter `compact` was renamed to `format_type` to make it more similar to the parameters to `format_currency`.

`format_type` can be `None` to use the format `locale.decimal_formats[None]`, or `"short"` / `"long"` to use the respective compact decimal format.

(Note: compact format currencies are not implemented by this PR, they appear to partially implemented already and broken and can be corrected in a later PR).

jun66j5:
I think we could add doc tests for the plural form like:

```diff
index 7e0ddd63a..fc2164183 100644
--- a/babel/numbers.py
+++ b/babel/numbers.py
@@ -417,6 +417,10 @@ def format_decimal(
     u'12.35K'
     >>> format_decimal(1234567, locale='ja_JP', format_type="short")
     u'123万'
+    >>> format_decimal(2345678, locale='mk', format_type='long')
+    u'2 милиони'
+    >>> format_decimal(21098765, locale='mk', format_type='long')
+    u'21 милион'

     :param number: the number to format
     :param format:
```

DenverCoder1:
> I think we could add doc tests for the plural form 

Added :+1:

babel/numbers.py
```diff
@@ -416,15 +429,54 @@ def format_decimal(
...
+                                    a value greater than `0`, the `decimal_quantization`
+                                    will be treated as `False`.
     """
     locale = Locale.parse(locale)
-    if not format:
-        format = locale.decimal_formats.get(format)
+    if format_type in ("short", "long"):
+        number, format = _get_compact_format(number, format_type, locale, compact_fraction_digits)
+        # if compact_fraction_digits is set, we don't want to truncate the fraction digits
+        decimal_quantization = False if compact_fraction_digits > 0 else decimal_quantization
```
akx, line 443: I'd say just use a good old `if` – and no need to look at whether the value is strictly positive either; we just care about it being or not being set.

```suggestion
        # if compact_fraction_digits is set, we don't want to truncate the fraction digits
        if compact_fraction_digits:
            decimal_quantization = False
```

babel/numbers.py
```diff
@@ -416,15 +429,54 @@ def format_decimal(
...
+            # equal to the number of 0's in the pattern minus 1
+            number = number / (magnitude / (10 ** (pattern.count("0") - 1)))
+            # round to the number of fraction digits requested
+            number = round(number, compact_fraction_digits)
+            # if the remaining number is singular, use the singular format
+            plural_form = locale.plural_form(abs(number))
+            plural_form = plural_form if plural_form in compact_format else "other"
+            format = compact_format[plural_form][str(magnitude)]
+            break
+    format = format if format is not None else locale.decimal_formats.get(None)
```
akx, line 476: I think this could be clearer as a regular `if`.
```suggestion
    if format is None:  # Did not find a format, fall back.
        format = locale.decimal_formats.get(None)
```
or similar..?

tests/test_numbers.py
```diff
@@ -121,6 +121,39 @@ def test_group_separator(self):
         assert numbers.format_currency(101299.98, 'EUR', locale='en_US', group_separator=True, format_type='name') == u'101,299.98 euros'
         assert numbers.format_percent(251234.1234, locale='sv_SE', group_separator=True) == u'25\xa0123\xa0412\xa0%'
 
+    def test_compact(self):
```
akx, line 124: Just to make sure: is there a test here that sees how `decimal_quantization` interacts with this formatting?

babel/numbers.py
```diff
@@ -416,15 +429,54 @@ def format_decimal(
                                  the format pattern. Defaults to `True`.
     :param group_separator: Boolean to switch group separator on/off in a locale's
                             number format.
+    :param format_type: Format to use (`None`, "short" or "long"). The standard decimal
+                        format is `None`. Defaults to `None`.
+    :param compact_fraction_digits: Number of fraction digits to use in "short" or "long"
+                                    format. Defaults to `0`. If this is set to
+                                    a value greater than `0`, the `decimal_quantization`
+                                    will be treated as `False`.
```
akx, line 437: I'm a bit on the fence about these being additional arguments to `format_decimal`, especially given `format_type` being such a generic term.

A separate function – `format_compact_decimal(number, *, type, digits=0, locale=None)` might be a better API? (Would there be a reason to keep `decimal_quantization` and `group_separator` in that API?)

akx (changes requested):
Thank you for getting to work on this! The code looks mainly great, just a couple thoughts here and there, and maybe we should discuss whether to overload the `format_decimal` API further.

(Also thanks, @jun66j5, for thoroughly reviewing this earlier!)

tests/test_numbers.py
DenverCoder1: I don't think it needs one since the argument is not used with compact format types. If `compact_fraction_digits` is set, it is `True`, otherwise it is effectively `False` even if manually set.

babel/numbers.py
DenverCoder1: I've seen other APIs that only have one method for number formatting, for example [GlobalizeJS](https://github.com/globalizejs/globalize#number-module). I agree that making it separate would probably be cleaner since some of the parameters are unused.

DenverCoder1: I've updated the API, let me know what you think :+1:

tests/test_numbers.py
akx: 👍 That's what I thought; thanks for the clarification.

akx (changes requested):
Looks quite excellent! Just a couple of tiny comments to go.

akx (approved):
Great, thank you!

PR #912, opened 2022-10-26
align license with OSI template

lsmith77:
## Overview Description

I am not sure why but github isn't detecting your use of the BSD-3-Clause properly.

I am trying to automate an overview of the licenses included in my projects and currently I cannot detect the license using dependabots database.

For example here it works:
https://github.com/pallets/click/blob/main/LICENSE.rst

I suspect it is due to including the text `All rights reserved.` and/or other differences which isn't part of the license template https://opensource.org/licenses/BSD-3-Clause

Maybe also due to formatting but I kept things formatted to a similar character length per line as in the current version.

Would you accept a PR this to hopefully fix Github's ability to detect the license and to bring your license inline with the template?


lsmith77:
```
➜  babel git:(patch-1) licensee detect .
License:        BSD-3-Clause
Matched files:  LICENSE
LICENSE:
  Content hash:  a961b19cc6921d510e29a13b0ba1a826fcffe41c
  Attribution:   Copyright (c) 2013-2022 by the Babel Team, see AUTHORS for more information.
  Confidence:    100.00%
  Matcher:       Licensee::Matchers::Exact
  License:       BSD-3-Clause
```

akx (approved):
It looks like "All Rights Reserved" has inadvertently been there forever (00f16f91a878676a55aa1e6e85b5e1092f7cf8e9) and doesn't really belong in the file.

The change
```diff
-The name of the author
+Neither the name of the copyright holder nor the names of its contributors
```
is also benign: https://github.com/spdx/license-list-XML/issues/399

> discussed on legal call today; we agree that these two wordings are identical and they should be marked up to both match BSD-3 clause.

Thank you!

Issue #715, opened 2020-06-04
Pybabel should raise an error when it encounters f-strings


fabiommendes:
I mistakenly put something like that in my code:

```
_(f"Name: {x}")
```

and Babel fails (as in it crashes with an exception) to extract strings from this file. The reason why it crashes is that the extract_python function relies on eval(token_code) to obtain the string from its source code representation. While this is legitimate in general, it does not work for f-strings (evaluation will almost certainly crash due to missing variables).

I understand that including f-strings inside the _() function is a mistake, I think Babel should warn the user about the incorrect usage instead of simply crashing. This can be easily fixed by simply wrapping the offending eval() call into a try/except block:

```python
# extract_python() in babel.messages.extract.py
            elif tok == STRING:
                # Unwrap quotes in a safe manner, maintaining the string's
                # encoding
                # https://sourceforge.net/tracker/?func=detail&atid=355470&
                # aid=617979&group_id=5470
                code = compile('# coding=%s\n%s' % (str(encoding), value),
                               '<string>', 'eval', future_flags)
                try:
                    value = eval(code, {'__builtins__': {}}, {})
                except NameError:
                    continue
                else:
                    if PY2 and not isinstance(value, text_type):
                        value = value.decode(encoding)
                    buf.append(value)
``` 

It still accepts f-strings with no interpolation (which is a bad style, but do not cause any damage).

evonbevern:
Hi, I can send in a pull request for this. Will be done in the next few days.

iranzo:
I've recently falled into this one, as per python 3.9 I was moving to using fstrings and for one of the files it fails with error. With the previous version of the file, using regular "%s" substitutions it worked, but when running pyupgrade over the code and getting some fstrings instead, it started failing.

Dreamsorcerer:
> I've recently falled into this one, as per python 3.9 I was moving to using fstrings and for one of the files it fails with error.

Just to clarify, while it should probably produce a reasonable warning, rather than an exception, the problem in your code is the fact that you were formatting a string _before_ sending it for translation.

i.e. `_("Hello, {name}".format(name="Joe"))` will try to lookup a translation for the string `"Hello, Joe"`, which will obviously fail, unless you've added a translation for every single user individually..
Instead you need to format a string which has already been translated like `_("Hello, {name}").format(name="Joe")`. This looks up the translation for `"Hello, {name}"`, which will be found, and then formats the resulting string.
If done correctly, pyupgrade would be unable to convert to an f-string.

[...]

PR #915, opened 2022-10-31
Use `ast` instead of `eval` for string extraction

akx:
This is safer (as we don't actually execute anything), and allows us to parse f-strings too.

Closes #769 (supersedes it)
Refs #715 (doesn't add an error yet, but doesn't crash on f-strings)

### DIFF
--- a/babel/messages/extract.py
+++ b/babel/messages/extract.py
@@ -15,7 +15,7 @@
     :copyright: (c) 2013-2022 by the Babel Team.
     :license: BSD, see LICENSE for more details.
 """
-
+import ast
 import os
 from os.path import relpath
 import sys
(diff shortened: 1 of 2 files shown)


babel/messages/extract.py
```diff
@@ -516,6 +511,29 @@ def extract_python(fileobj, keywords, comment_tags, options):
...
+    if not isinstance(code, ast.Expression):
+        return None
+    body = code.body
+    if isinstance(body, ast.Str):
+        return body.s
+    if isinstance(body, ast.JoinedStr):  # f-string
+        if all(isinstance(node, ast.Str) for node in body.values):
+            return ''.join(node.s for node in body.values)
+        if all(isinstance(node, ast.Constant) for node in body.values):
+            return ''.join(str(node.value) for node in body.values)
```
DenverCoder1, line 532: From what I can tell from trying things out, it seems like the AST always combines all the strings into a single constant. If that's the case maybe this could be simplified?

```py
>>> value='"123" f"456" "789"'
>>> ast.dump(compile(f'# coding={str(encoding)}\n{value}', '<string>', 'eval', ast.PyCF_ONLY_AST))
"Expression(body=JoinedStr(values=[Constant(value='123456789')]))"
```

I'm not too familiar with all the use cases of this function, but from looking at the examples and docstrings, it seems like it should work.

akx: Yeah, that's the case, but I suspect that's just an optimization in the parser that we shouldn't necessarily rely on. (Sure, it's not likely that it will be un-optimized anytime soon, but perhaps if Babel is used with a Python implementation that doesn't do that..?)

magdapoppins (approved):
Looks reasonable!

akx:
I've requested a couple of extra reviews off-GitHub from colleagues and such, and I suppose this is good to go :)

We'll get limited f-string support and less `eval`.

PR #922, opened 2022-11-02
docs(changes): Fix broken issue links

DenverCoder1:
Some links in CHANGES.rst are not linking directly to the corresponding issues properly. The extra `#` sign was causing it to be treated as a url hash on the issue listing page.

image


akx (approved):
🤦 Should've been more careful distilling the changelog.

Thanks!

PR #926, opened 2022-11-02
feat: Support for short compact currency formats

### LINKED ISSUE #923: `format_currency` with `standard:short` format does not format correctly
## Overview Description

Currently there isn't a way to properly format currencies in compact form by locale.

The data is partially imported to `locale.currency_formats["standard:short"]`

```py
>>> list(Locale("en_US").currency_formats)
['standard', 'accounting', 'standard:short']
```

But it does not store it for all magnitudes, rather just the last one

```py
>>> Locale("en_US").currency_formats["standard:short"]
<NumberPattern '¤000T'>
```

This leads to `standard:short` format not formatting numbers correctly

```py
>>> numbers.format_currency(1.98, 'USD', format_type="standard:short")
'$001.98T'
```

## Additional Information

The `import_cldr.py` needs to be updated to correctly store the patterns for all magnitudes and count types ("one", "other")

It may also make sense to make a new function, for example, `format_compact_currency` to make the API similar to newly added `format_compact_decimal`.

DenverCoder1:
Added a way to format currency with standard "short" currency formats from the CLDR data.

Closes #923

```py
>>> format_compact_currency(123456789, 'USD', locale='en_US', fraction_digits=2)
u'$123.46M'

>>> format_compact_currency(-123456789, 'USD', locale='en_US', fraction_digits=2)
u'-$123.46M'

>>> Format("en_US").compact_currency(123456789, 'USD', fraction_digits=2)
u'$123.46M'
```

`"short"` is currently the only length type that seems to appear in the data.

I'm not 100% sure about setting `"¤0"` as a fallback. As far as I can tell, CLDR doesn't provide a way for formatting currency without decimal places (short) for numbers less than 1000. But this format (`"¤0"`) works with the formats I've checked and tested.

Note: `"standard:short" `no longer appears in `locale.currency_formats` (see #923). This was never documented or tested and did not format things properly in the first place, so I don't think this should be considered a breaking change.


babel/numbers.py
```diff
@@ -624,6 +625,36 @@ def _format_currency_long_name(
...
+    :param format_type: the compact format type to use. Defaults to "short".
+    :param locale: the `Locale` object or locale identifier
+    :param fraction_digits: Number of digits after the decimal point to use. Defaults to `0`.
+    """
+    locale = Locale.parse(locale)
+    try:
+        compact_format = locale.compact_currency_formats[format_type]
+    except KeyError as e:
+        raise UnknownCurrencyFormatError(
+            "%r is not a known currency format type" % format_type) from e
```
akx, line 649: You can use an f-string in new code.

```suggestion
        raise UnknownCurrencyFormatError(f"{format_type!r} is not a known currency format type") from e
```

scripts/import_cldr.py
```diff
@@ -915,10 +915,6 @@ def parse_currency_formats(data, tree):
             curr_length_type = length_elem.attrib.get('type')
             for elem in length_elem.findall('currencyFormat'):
                 type = elem.attrib.get('type')
-                if curr_length_type:
-                    # Handle `<currencyFormatLength type="short">`, etc.
-                    # TODO(3.x): use nested dicts instead of colon-separated madness
```
akx, line 920: In hindsight the `(3.x)` in the TODO was probably too conservative. After all we've always noted the format dicts may change between version.

👍

akx (changes requested):

scripts/import_cldr.py
DenverCoder1: Yeah, and also as mentioned, the short format was always returning the format for 100 trillions, so it's probably safe to assume it's not actively in use

babel/numbers.py
```diff
@@ -624,6 +625,35 @@ def _format_currency_long_name(
...
 
+def format_compact_currency(number, currency, *, format_type="short", locale=LC_NUMERIC, fraction_digits=0):
+    u"""Format a number as a currency value in compact form.
+
+    >>> format_compact_currency(12345, 'USD', locale='en_US')
+    u'$12K'
+    >>> format_compact_currency(123456789, 'USD', locale='en_US', fraction_digits=2)
+    u'$123.46M'
+    >>> format_compact_currency(123456789, 'EUR', locale='de_DE', fraction_digits=1)
+    u"123,5\xa0Mio'.'\xa0€"
```
akx, line 636: This is actually incorrectly formatted – using `print()` to clean out the NBSPs:
```
>>> print(u"123,5\xa0Mio'.'\xa0€")
123,5 Mio'.' €
```
The output should be
```
123,5 Mio. €
```
i.e. the literal segment `'.'` should get correctly interpreted.

Good choice of doctest case 😁

akx (changes requested):

DenverCoder1:
As an alternative to using `"¤0"` as a fallback, maybe it would make sense to take the smallest magnitude for each locale and strip out anything that's not "¤" or "0", or spaces? That way, if the the currency symbol should go afterwards, it would format in the right order.

babel/numbers.py
```diff
@@ -1112,6 +1112,11 @@ def apply(
             retval = retval.replace(u'¤¤', currency.upper())
             retval = retval.replace(u'¤', get_currency_symbol(currency, locale))
 
+        # remove single quotes around text, except for doubled single quotes
+        # which are replaced with a single quote
+        retval = re.sub(r"(^|[^'])'(?!')|'(?<!')($|[^'])", r'\1\2', retval)
+        retval = retval.replace("''", "'")
```
jun66j5, line 1118: I think the regular expression is too complex and buggy like:

```pycon
>>> numbers.format_decimal(42, "'$'''0", locale='en')
"$''42"  # shoud be "$'42"
```

```suggestion
        retval = re.sub(r"'([^']*)'", lambda m: m.group(1) or "'", retval)
```

DenverCoder1: Looks good :+1:

akx (approved):
Great stuff, thanks @DenverCoder1 and @jun66j5!

PR #930, opened 2022-11-13
Cast number to Decimal in _get_compact_format

### LINKED ISSUE #929: Support for decimal.Decimal and str with format_compact_decimal
## Overview Description

Currently only `float` (or compatible types including `int`) work with `format_compact_decimal`.

Using `decimal.Decimal` in `format_compact_decimal` results in `unsupported operand type(s) for /: 'decimal.Decimal' and 'float'`

Using `str` results in `bad operand type for abs(): 'str'`

## Steps to Reproduce

```py
>>> from babel.numbers import *
>>> import decimal
>>> format_compact_decimal(decimal.Decimal("1000"))
  File "babel/numbers.py", line 450, in format_compact_decimal
    number, format = _get_compact_format(number, format_type, locale, fraction_digits)
  File "babel/numbers.py", line 472, in _get_compact_format
    number = number / (magnitude / (10 ** (pattern.count("0") - 1)))
TypeError: unsupported operand type(s) for /: 'decimal.Decimal' and 'float'
>>> format_compact_decimal("1000")
  File "babel/numbers.py", line 450, in format_compact_decimal
    number, format = _get_compact_format(number, format_type, locale, fraction_digits)
  File "babel/numbers.py", line 463, in _get_compact_format
    if abs(number) >= magnitude:
TypeError: bad operand type for abs(): 'str'
```

## Expected Results

Just like format_decimal, it should probably return the correctly formatted number.

## Additional Information

Possibly we can cast it to a float or even int in the function to get the compact format

DenverCoder1:
Fixes #929 

Division does not work with decimal.Decimal. This uses `//` so integer division is performed instead.

`abs` and `round` don't work with strings. This converts the number to a Decimal so that `_get_compact_format` won't raise a ValueError as long as the number supports Decimal conversion. Also has a fix for how [rounding modes](https://babel.pocoo.org/en/latest/numbers.html#rounding-modes) are treated.

This adds support for `decimal.Decimal` and `str` types to the number parameter of `format_compact_*` functions.

```py
>>> from babel.numbers import *
>>> import decimal
>>> format_compact_decimal(decimal.Decimal("1000"))
"1K"
>>> format_compact_decimal("1000")
"1K"
```

### DIFF
--- a/babel/numbers.py
+++ b/babel/numbers.py
@@ -455,6 +455,7 @@ def _get_compact_format(number, compact_format, locale, fraction_digits=0):
     https://www.unicode.org/reports/tr35/tr35-45/tr35-numbers.html#Compact_Number_Formats.
     """
     format = None
+    number = float(number)
     for magnitude in sorted([int(m) for m in compact_format["other"]], reverse=True):
         if abs(number) >= magnitude:
             # check the pattern using "other" as the amount
--- a/babel/numbers.py
+++ b/babel/numbers.py
@@ -426,7 +426,7 @@ def format_compact_decimal(number, *, format_type="short", locale=LC_NUMERIC, fr
     >>> format_compact_decimal(12345, format_type="long", locale='en_US')
     u'12 thousand'
     >>> format_compact_decimal(12345, format_type="short", locale='en_US', fraction_digits=2)
-    u'12.35K'
+    u'12.34K'
     >>> format_compact_decimal(1234567, format_type="short", locale="ja_JP")
     u'123万'
     >>> format_compact_decimal(2345678, format_type="long", locale="mk")
@@ -454,8 +454,9 @@ def _get_compact_format(number, compact_format, locale, fraction_digits=0):
     The algorithm is described here:
     https://www.unicode.org/reports/tr35/tr35-45/tr35-numbers.html#Compact_Number_Formats.
     """
+    if not isinstance(number, decimal.Decimal):
+        number = decimal.Decimal(str(number))
     format = None
-    number = float(number)
     for magnitude in sorted([int(m) for m in compact_format["other"]], reverse=True):
         if abs(number) >= magnitude:
             # check the pattern using "other" as the amount
@@ -466,7 +467,7 @@ def _get_compact_format(number, compact_format, locale, fraction_digits=0):
                 break
             # otherwise, we need to divide the number by the magnitude but remove zeros
             # equal to the number of 0's in the pattern minus 1
-            number = number / (magnitude / (10 ** (pattern.count("0") - 1)))
+            number = number / (magnitude // (10 ** (pattern.count("0") - 1)))
             # round to the number of fraction digits requested
             number = round(number, fraction_digits)
             # if the remaining number is singular, use the singular format
--- a/tests/test_numbers.py
+++ b/tests/test_numbers.py
@@ -154,6 +154,8 @@ class FormatDecimalTestCase(unittest.TestCase):
         assert numbers.format_compact_decimal(-123456789, format_type='long', locale='en_US') == u'-123 million'
         assert numbers.format_compact_decimal(2345678, locale='mk', format_type='long') == u'2 милиони'
         assert numbers.format_compact_decimal(21098765, locale='mk', format_type='long') == u'21 милион'
+        assert numbers.format_compact_decimal(decimal.Decimal(123456789), format_type='short', locale='en_US') == u'123M'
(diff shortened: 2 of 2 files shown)


akx:
Initial pre-review thought: I'm pretty hesitant about the possible loss of precision w/ casting to `float`...

DenverCoder1:
As far as I can tell, in compact format, the fractional part never even shows up in any of the formats, so I think even truncating it to an int would be fine.

tests/test_numbers.py
```diff
@@ -154,6 +154,8 @@ def test_compact(self):
         assert numbers.format_compact_decimal(-123456789, format_type='long', locale='en_US') == u'-123 million'
         assert numbers.format_compact_decimal(2345678, locale='mk', format_type='long') == u'2 милиони'
         assert numbers.format_compact_decimal(21098765, locale='mk', format_type='long') == u'21 милион'
+        assert numbers.format_compact_decimal(decimal.Decimal(123456789), format_type='short', locale='en_US') == u'123M'
+        assert numbers.format_compact_decimal('123456789', format_type='short', locale='en_US') == u'123M'
```
jun66j5, line 158: I don't think the test passing `str` is needed.

The `format_compact_decimal` uses `NumberPattern.apply` as the same of `format_decimal`. docstring of the `apply` says `:type value: decimal.Decimal|float|int`. `str` instance is not expected.

Actually, `test_format_decimal` doesn't have tests passing `str` instance to `format_decimal`

jun66j5:
I think we could use `//` operator instead of casting to float.

```diff
index 8baf110b8..907ae23f6 100644
--- a/babel/numbers.py
+++ b/babel/numbers.py
@@ -465,7 +465,7 @@ def _get_compact_format(number, compact_format, locale, fraction_digits=0):
                 break
             # otherwise, we need to divide the number by the magnitude but remove zeros
             # equal to the number of 0's in the pattern minus 1
-            number = number / (magnitude / (10 ** (pattern.count("0") - 1)))
+            number = number / (magnitude // (10 ** (pattern.count("0") - 1)))
             # round to the number of fraction digits requested
             number = round(number, fraction_digits)
             # if the remaining number is singular, use the singular format
```

tests/test_numbers.py
DenverCoder1: `str` works differently in `format_compact_decimal` than in `format_decimal` since it gets passed through `_get_compact_format` which previously did not support string because of the `abs()` call

jun66j5:
I noticed the `format_compact_decimal` ignores [rounding modes](https://babel.pocoo.org/en/latest/numbers.html#rounding-modes). We should cast to `Decimal` before applying `round()` to the `number`.

```diff
index 8baf110b8..6c9ab24b0 100644
--- a/babel/numbers.py
+++ b/babel/numbers.py
@@ -426,7 +426,7 @@ def format_compact_decimal(number, *, format_type="short", locale=LC_NUMERIC, fr
     >>> format_compact_decimal(12345, format_type="long", locale='en_US')
     u'12 thousand'
     >>> format_compact_decimal(12345, format_type="short", locale='en_US', fraction_digits=2)
-    u'12.35K'
+    u'12.34K'
     >>> format_compact_decimal(1234567, format_type="short", locale="ja_JP")
     u'123万'
     >>> format_compact_decimal(2345678, format_type="long", locale="mk")
@@ -454,6 +454,8 @@ def _get_compact_format(number, compact_format, locale, fraction_digits=0):
     The algorithm is described here:
     https://www.unicode.org/reports/tr35/tr35-45/tr35-numbers.html#Compact_Number_Formats.
     """
+    if not isinstance(number, decimal.Decimal):
+        number = decimal.Decimal(str(number))
     format = None
     for magnitude in sorted([int(m) for m in compact_format["other"]], reverse=True):
         if abs(number) >= magnitude:
```

DenverCoder1:
Thanks for taking a look! Suggested applied :+1:

jun66j5:
I found other issue in `format_compact_decimal` while investigating this issue. Filed in #931.

babel/numbers.py
```diff
@@ -426,7 +426,7 @@ def format_compact_decimal(number, *, format_type="short", locale=LC_NUMERIC, fr
     >>> format_compact_decimal(12345, format_type="long", locale='en_US')
     u'12 thousand'
     >>> format_compact_decimal(12345, format_type="short", locale='en_US', fraction_digits=2)
-    u'12.35K'
+    u'12.34K'
```
akx, line 429: This changing probably depends on the ambient decimal context. But that's okay 👍

akx (approved):
Good stuff!

PR #932, opened 2022-11-17
Fix compact singular formats and patterns with no numbers

### LINKED ISSUE #931: format_compact_decimal incorrectly works when the pattern has no 0 characters
## Overview Description

`format_compact_decimal` incorrectly works when the pattern has no 0 characters. Also, it doesn't handle `count="1"` rule.

## Steps to Reproduce

```pycon
>>> from babel import numbers
>>> numbers.format_compact_decimal(1000, format_type='long', locale='it')
'mille1'  # should be 'mille'
>>> numbers.format_compact_decimal(1234, format_type='long', locale='it')
'mille1'  # should be '1 mila'
>>> numbers.format_compact_decimal(1000, format_type='long', locale='fr')
'1 millier'  # should be 'mille'
>>> numbers.format_compact_decimal(1234, format_type='long', locale='fr')
'1 millier'  # correct
```

## Additional Information

* https://github.com/unicode-org/cldr/blob/release-41/common/main/it.xml#L5210
* https://github.com/unicode-org/cldr/blob/release-41/common/main/fr.xml#L7743
* [Explicit 0 and 1 rules](https://unicode.org/reports/tr35/tr35-numbers.html#Explicit_0_1_rules)

DenverCoder1:
Fixes #931

* Adds support for patterns with no numbers
* In compact format, only uses "one" format if the number pluralizes to "one" *before* rounding
* In compact format, [explicit "1" formats](https://unicode.org/reports/tr35/tr35-numbers.html#Explicit_0_1_rules) can now be applied (there [seem to be none](https://gist.github.com/DenverCoder1/9c34fb012cefdc75f8d53724c24fe260) with explicit "0" formats)

```py
>>> numbers.format_compact_decimal(1000, format_type='long', locale='it')
'mille'  # previously 'mille1'
>>> numbers.format_decimal(1, format="mille")
'mille'  # previously 'mille1'
>>> format_compact_decimal(21098765, format_type="long", locale="mk")
'21 милиони'  # previously '21 милион' (would still be singular if the number were 21000000)
>>> numbers.format_compact_decimal(1000, format_type='long', locale='fr')
'mille'  # previously '1 millier'
```

### DIFF
--- a/babel/numbers.py
+++ b/babel/numbers.py
@@ -474,8 +474,6 @@ def _get_compact_format(number, compact_format, locale, fraction_digits=0):
                 plural_form = "other"
             if number == 1 and "1" in compact_format:
                 plural_form = "1"
-            elif number == 0 and "0" in compact_format:
-                plural_form = "0"
             format = compact_format[plural_form][str(magnitude)]
             number = rounded
             break
--- a/babel/numbers.py
+++ b/babel/numbers.py
@@ -17,6 +17,7 @@
 # TODO:
 #  Padding and rounding increments in pattern:
 #  - https://www.unicode.org/reports/tr35/ (Appendix G.6)
+from __future__ import annotations
 import decimal
 import re
 from datetime import date as date_, datetime as datetime_
@@ -967,7 +968,8 @@ def parse_pattern(pattern):
 class NumberPattern:
 
     def __init__(self, pattern, prefix, suffix, grouping,
-                 int_prec, frac_prec, exp_prec, exp_plus, number_pattern=None):
+                 int_prec, frac_prec, exp_prec, exp_plus,
+                 number_pattern: str | None = None):
         # Metadata of the decomposed parsed pattern.
         self.pattern = pattern
         self.prefix = prefix
--- a/babel/numbers.py
+++ b/babel/numbers.py
@@ -431,7 +431,7 @@ def format_compact_decimal(number, *, format_type="short", locale=LC_NUMERIC, fr
     u'123万'
     >>> format_compact_decimal(2345678, format_type="long", locale="mk")
     u'2 милиони'
-    >>> format_compact_decimal(21098765, format_type="long", locale="mk")
+    >>> format_compact_decimal(21000000, format_type="long", locale="mk")
     u'21 милион'
 
     :param number: the number to format
@@ -467,11 +467,17 @@ def _get_compact_format(number, compact_format, locale, fraction_digits=0):
             # equal to the number of 0's in the pattern minus 1
             number = number / (magnitude / (10 ** (pattern.count("0") - 1)))
             # round to the number of fraction digits requested
-            number = round(number, fraction_digits)
+            rounded = round(number, fraction_digits)
             # if the remaining number is singular, use the singular format
             plural_form = locale.plural_form(abs(number))
-            plural_form = plural_form if plural_form in compact_format else "other"
+            if plural_form not in compact_format:
+                plural_form = "other"
+            if number == 1 and "1" in compact_format:
+                plural_form = "1"
+            elif number == 0 and "0" in compact_format:
+                plural_form = "0"
             format = compact_format[plural_form][str(magnitude)]
+            number = rounded
             break
     return number, format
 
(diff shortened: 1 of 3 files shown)


babel/numbers.py
```diff
@@ -431,7 +431,7 @@ def format_compact_decimal(number, *, format_type="short", locale=LC_NUMERIC, fr
     u'123万'
     >>> format_compact_decimal(2345678, format_type="long", locale="mk")
     u'2 милиони'
-    >>> format_compact_decimal(21098765, format_type="long", locale="mk")
+    >>> format_compact_decimal(21000000, format_type="long", locale="mk")
     u'21 милион'
```
akx, line 435: What happened to the previous number to test with?

babel/numbers.py
```diff
@@ -467,11 +467,15 @@ def _get_compact_format(number, compact_format, locale, fraction_digits=0):
...
             number = number / (magnitude / (10 ** (pattern.count("0") - 1)))
             # round to the number of fraction digits requested
-            number = round(number, fraction_digits)
+            rounded = round(number, fraction_digits)
             # if the remaining number is singular, use the singular format
             plural_form = locale.plural_form(abs(number))
-            plural_form = plural_form if plural_form in compact_format else "other"
+            if plural_form not in compact_format:
+                plural_form = "other"
+            if number == 1 and "1" in compact_format:
```
akx, line 475: Maybe `elif`? 
```suggestion
            elif number == 1 and "1" in compact_format:
```

babel/numbers.py
```diff
@@ -1112,7 +1117,7 @@ def apply(
 
         retval = ''.join([
             self.prefix[is_negative],
-            number,
+            number if self.number_pattern != '' else '',
```
akx, line 1120: If the distinction between `None` and `''` isn't significant, maybe just
```suggestion
            (number if self.number_pattern else ''),
```

babel/numbers.py
```diff
@@ -957,17 +961,18 @@ def parse_precision(p):
...
                          int_prec, frac_prec,
-                         exp_prec, exp_plus)
+                         exp_prec, exp_plus, number)
 
 
 class NumberPattern:
 
     def __init__(self, pattern, prefix, suffix, grouping,
-                 int_prec, frac_prec, exp_prec, exp_plus):
+                 int_prec, frac_prec, exp_prec, exp_plus, number_pattern=None):
```
akx, line 970: Could you add a type hint for the new argument? We haven't had type hints much anywhere yet, but it'll happen sooner or later.

akx (changes requested):

babel/numbers.py
DenverCoder1: It's checking if plural before rounding now so 21.0 would be singular but 21.098765 would not even if it's going to be truncated.

Assumption based on @jun66j5's comment

https://github.com/python-babel/babel/issues/931#issuecomment-1318845308

babel/numbers.py
DenverCoder1: If the number were `1`, the plural form could initially be "one", if "one" is not in `compact_format`, it will become "other", but if "1" is, it should still use "1".

babel/numbers.py
DenverCoder1: Ok, just to be sure, this assumes we are not concerned about people manually constructing `NumberPattern` on their own, since that could make it a breaking change.

Eg. if someone does `NumberPattern(pattern, (pos_prefix, neg_prefix),
                         (pos_suffix, neg_suffix), grouping,
                         int_prec, frac_prec,
                         exp_prec, exp_plus)` (note, no number pattern), then `number_pattern` will be `None` and this `NumberPattern` won't show numbers anymore even though it used to.

Differentiating here is the reason I used None rather than empty string as the default.

babel/numbers.py
DenverCoder1: I actually just type-hinting pretty much all modules except `messages` and some of `support` that in types-babel. Maybe I could make a PR to copy them all over?

akx (approved):
Thanks!

Issue #937, opened 2022-12-23
Unnatural formatting of negative currency


michaelfm1211:
## Overview Description
When using `babel.numbers.format_currency`, the negative currency is formatted with the negative sign before the currency name, like `'-$100.00'` and `-CHF 100.00`. See [this Stack Overflow question](https://stackoverflow.com/q/74894618/13376511) for more examples. In some locales this doesn't make much of a change (ex: in en_US both -$100 and $-100 looks OK), however, in others it doesn't make as much sense (ex: in de_CH the difference between -CHF 100.50 and CHF -100.50 is more noticeable).
## Steps to Reproduce

Run `format_currency()` with a custom format and pass a negative value for the currency. For example:
```python
format_currency(value, 'CHF', '¤¤ #,##0.00', locale='de_CH')
```

## Actual Results
`'-CHF 100.50'`
## Expected Results
`'CHF -100.50'`
## Reproducibility
This is the default behavior.

PR #938, opened 2022-12-23
Add negative sign after currency names (#937)

michaelfm1211:
This PR changes `babel.numbers.parse_pattern` to move the negative sign to the end of the prefix if the format string includes a placeholder for a currency name.  It updates some tests.

### DIFF
--- a/babel/numbers.py
+++ b/babel/numbers.py
@@ -916,7 +916,7 @@ def parse_pattern(pattern):
         neg_prefix, _, neg_suffix = _match_number(neg_pattern)
     else:
         pos_prefix, number, pos_suffix = _match_number(pos_pattern)
-        neg_prefix = f"-{pos_prefix}"
+        neg_prefix = f"-{pos_prefix}" if '¤' not in pattern else f'{pos_prefix}-'
         neg_suffix = pos_suffix
     if 'E' in number:
         number, exp = number.split('E', 1)
--- a/tests/test_numbers.py
+++ b/tests/test_numbers.py
@@ -427,7 +427,7 @@ def test_format_compact_currency():
     assert numbers.format_compact_currency(999, 'USD', locale='en_US', format_type="short") == u'$999'
     assert numbers.format_compact_currency(123456789, 'USD', locale='en_US', format_type="short") == u'$123M'
     assert numbers.format_compact_currency(123456789, 'USD', locale='en_US', fraction_digits=2, format_type="short") == u'$123.46M'
-    assert numbers.format_compact_currency(-123456789, 'USD', locale='en_US', fraction_digits=2, format_type="short") == u'-$123.46M'
+    assert numbers.format_compact_currency(-123456789, 'USD', locale='en_US', fraction_digits=2, format_type="short") == u'$-123.46M'
     assert numbers.format_compact_currency(1, 'JPY', locale='ja_JP', format_type="short") == u'￥1'
     assert numbers.format_compact_currency(1234, 'JPY', locale='ja_JP', format_type="short") == u'￥1234'
     assert numbers.format_compact_currency(123456, 'JPY', locale='ja_JP', format_type="short") == u'￥12万'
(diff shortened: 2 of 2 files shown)


DenverCoder1:
I don't know about all locales, but at least in the US, I find it much more common to see "-$123.46M" than "$-123.46M".

Is there a source for this format being preferred?

michaelfm1211:
This PR is to address [this issue](https://github.com/python-babel/babel/issues/937) that I opened (sorry if that wasn't as clear as it should've been). The rationale behind it is that "-$123.46M" is normal, but "-CHF 123.46M" isn't. With the change, it would fix the latter to "CHF -123.46M".

DenverCoder1:
It seems to me that even with this change it still outputs 'CHF-12,345.00'. And it would best to make it so that it does not affect the outputs which are already correct such as the '-$123.46M' case.

```py
>>> format_currency(-12345, 'CHF', locale="en_US")
'CHF-12,345.00'
```

DenverCoder1:
There may be other solutions, but I might suggest to put this back to how it was
```diff
-        neg_prefix = f"-{pos_prefix}" if '¤' not in pattern else f'{pos_prefix}-'
+        neg_prefix = f"-{pos_prefix}"
```

And instead change where it replaces the currency symbols to guess the minus sign location based on the currency symbol. For example:

```py
# if the currency symbol is the same as the currency code
# and the value is negative, move the currency symbol
# to be before the number instead of before the currency code
if currency_symbol == currency_code and is_negative:
    retval = retval.replace(r'-¤', r'¤ -')
```

```diff
         if u'¤' in retval:
-            retval = retval.replace(u'¤¤¤',
-                                    get_currency_name(currency, value, locale))
-            retval = retval.replace(u'¤¤', currency.upper())
-            retval = retval.replace(u'¤', get_currency_symbol(currency, locale))
+            currency_name = get_currency_name(currency, value, locale)
+            currency_code = currency.upper()
+            currency_symbol = get_currency_symbol(currency, locale)
+            retval = retval.replace(u'¤¤¤', currency_name)
+            retval = retval.replace(u'¤¤', currency_code)
+            # if the currency symbol is the same as the currency code
+            # and the value is negative, move the currency symbol
+            # to be before the number instead of before the currency code
+            if currency_symbol == currency_code and is_negative:
+                retval = retval.replace(r'-¤', r'¤ -')
+            retval = retval.replace(u'¤', currency_symbol)
```

This allows changing the behavior only in the problematic cases, being the ones where the symbol (¤) is the three letter abbreviation and not a currency sign like "$".

Feel free to share your thoughts on this.

michaelfm1211:
I'm not sure if that change would be effective in all cases, especially when using the full currency names. For example, with your suggested change:
```python
>>> format_currency(-100.50, 'CHF', '¤¤ #,##0.00', locale='de_CH')
'-CHF 100.50'
>>> format_currency(-100.50, 'CHF', '¤¤¤  #,##0.00', locale='de_CH')
'-Schweizer Franken  100.50'
```
However, the code from the original PR prevents the odd negative sign before letters:
```python
>>> format_currency(-100.50, 'CHF', '¤¤ #,##0.00', locale='de_CH')
'CHF -100.50'
>>> format_currency(-100.50, 'CHF', '¤¤¤  #,##0.00', locale='de_CH')
'Schweizer Franken  -100.50'
```

Let me know if you observed any other results, or if I am at all wrong about this

I should also note that this PR doesn't affect the default format because the `parse_pattern()` is only called when a custom format is not provided to `format_currency()`. Also, custom formats with a negative subpattern are not affected either. So when using the default format with or without the suggestion, the correct value is returned:
```python
>>> format_currency(-100.50, 'CHF', locale='de_CH')
'CHF-100.50'
>>> format_currency(-100.50, 'USD', locale='en_US')
'$-100.50'
```

DenverCoder1:
At least in my opinion, in order for this to be a positive change, it needs to be able to differentiate with complete accuracy between a currency label being text vs. being a symbol since the default behavior for `format_currency(-100.50, 'USD', locale='en_US')` should definitely stay as `-$100.50` to avoid making an unexpected change to projects that rely on that.

As it would need to work with all currencies and all locales, it seems like it may not be so simple to differentiate here, in which case I think leaving it with `format_currency(-100.50, 'CHF', locale='de_CH') == '-CHF100.50'` may be the most practical approach. There is always the `format='¤¤ #,##0.00;¤¤ -#,##0.00'` workaround from [StackOverflow](https://stackoverflow.com/a/74894787/11608064) that can be used as a workaround after all.

PR #939, opened 2022-12-29
javascript extract improvements

johanneswilm:
- Improved javascript template string expression extracting
- handle line numbers + add tests for JS template strings


johanneswilm:
Hey @akx I have address the issues you mentioned in https://github.com/python-babel/babel/pull/792 . There are now tests, the style is more in line with your code and line numbers are handled correctly. Would it be possible to merge this or do you need me to do something more? 
PS: I will try to contact you in some other way as well, given that you wrote somewhere here that you muted github notifications.

babel/messages/jslexer.py
```diff
@@ -151,7 +151,7 @@ def unquote_string(string):
     return u''.join(result)
 
 
-def tokenize(source, jsx=True, dotted=True, template_string=True):
+def tokenize(source, jsx=True, dotted=True, template_string=True, lineno=1):
```
akx, line 154: Could you update the docstring for `lineno`?

babel/messages/extract.py
```diff
@@ -544,7 +545,11 @@ def extract_javascript(fileobj, keywords, comment_tags, options):
     :param options: a dictionary of additional options (optional)
                     Supported options are:
                     * `jsx` -- set to false to disable JSX/E4X support.
-                    * `template_string` -- set to false to disable ES6 template string support.
+                    * `template_string` -- if `True`, supports gettext(`key`)
+                    * `parse_template_string` -- if `True` will parse the
+                                                 contents of javascript
+                                                 template strings.
+    :param lineno: the line number to start from (optional)
```
akx, line 552: This should probably be clarified somehow to note that it doesn't change how parsing works, but just offsets the `lineno` in the yielded items?
```suggestion
    :param lineno: line number offset (for parsing embedded fragments)
```

Also, maybe it should be in `options` all in all, so as not to change the extractor function signature?

babel/messages/extract.py
```diff
@@ -665,3 +674,41 @@ def extract_javascript(fileobj, keywords, comment_tags, options):
...
+def parse_template_string(template_string, keywords, comment_tags, options, lineno):
+    """Parse JavaScript template string.
+
+    :param template_string: the template string to be parsed
+    :param keywords: a list of keywords (i.e. function names) that should be
+                     recognized as translation functions
+    :param comment_tags: a list of translator tags to search for and include
+                         in the results
+    :param options: a dictionary of additional options (optional)
+    :param lineno: the line number to start from
```
akx, line 688:
```suggestion
    :param lineno: starting line number (optional)
```

babel/messages/extract.py
```diff
@@ -665,3 +674,41 @@ def extract_javascript(fileobj, keywords, comment_tags, options):
             funcname = token.value
 
         last_token = token
+
+
+def parse_template_string(template_string, keywords, comment_tags, options, lineno):
```
akx, line 679:
```suggestion
def parse_template_string(template_string, keywords, comment_tags, options, lineno=1):
```

akx (changes requested):
Looks pretty good. Some minor changes requested...

babel/messages/jslexer.py
johanneswilm: Done!

johanneswilm:
@akx OK, all changes made as you requested.

As for:

> Also, maybe it should be in `options` all in all, so as not to change the extractor function signature?

This is up to you to decide. I can see arguments doing either (it is an option, but it's different from the other ones). Feel free to make that change if you think it's better to do it that way. I have left it as it was for now.

Please let me know if you need me to do anything else so that we can get this merged ASAP!

akx (approved):
LGTM. Thanks for the contribution!

PR #940, opened 2023-01-05
Implement zoneinfo support and make pytz optional

### LINKED ISSUE #716: Drop pytz as a dependency
Since the introduction of [PEP 495](https://www.python.org/dev/peps/pep-0495/), `pytz`'s unusual API has become [more of a liability than an asset](https://blog.ganssle.io/articles/2018/03/pytz-fastest-footgun.html). With the introduction of [PEP 615](https://www.python.org/dev/peps/pep-0615/) (the standard library `zoneinfo` module), there is really no good reason to continue using `pytz` at this point, and it is probably a good time to transition away from it. This is one of the more popular projects using `pytz`, and I think it would be a good idea to start transitioning away. 

I will note that Stuart Bishop, author and maintainer of `pytz`, is [planning to make `pytz` a thin wrapper around `zoneinfo`](https://discuss.python.org/t/pep-615-support-for-the-iana-time-zone-database-in-the-standard-library/3468/16) at some point, and so it is probably a good idea to get ahead of the curve on this.

I took a quick look to see how easy it would be to do, and I found that the `pytz`-specific integration is a bit deeper than I had hoped, and a straightforward adoption of `zoneinfo.ZoneInfo` would run into the following problems:

1. This project still supports Python 2.7, and [`backports.zoneinfo`](https://pypi.org/project/backports.zoneinfo/) only supports Python 3.6+.
2. The [`get_timezone` function](https://github.com/python-babel/babel/blob/e7e4265d9a037ac38bba99f8513fb9e48a1081ba/babel/dates.py#L176) explicitly returns `pytz` zones, which (as we noted), have a different API than standard `tzinfo` zones.
3. The [`get_next_timezone_transition`](https://github.com/python-babel/babel/blob/e7e4265d9a037ac38bba99f8513fb9e48a1081ba/babel/dates.py#L197) function relies on private implementation details of `pytz`, and it is not possible to write analogous code for `zoneinfo`, which stores transition information in opaque C structs. Presumably this will become a problem in the future when `pytz` becomes a thin wrapper around `zoneinfo` *anyway*.

I think that we can solve problem 1 easily enough by adding a compatibility function that uses `dateutil.tz` in Python 2.7 only — `dateutil.tz` has similar functionality to `zoneinfo.ZoneInfo`.

Depending on how you feel about breaking changes and how much you intend to explicitly support the `pytz` interface, I think you can solve problem 2 with a thin wrapper class deprecating the `pytz` interface, like this:

```python
class TzWrapper(tzinfo):
    def __init__(base_tzinfo):
        self._base_tzinfo = base_tzinfo

    def utcoffset(self, dt):
        return self._base_tzinfo.utcoffset(dt)

    def dst(self, dt):
        return self._base_tzinfo.dst(dt)

    def tzname(self, dt):
        return self._base_tzinfo.tzname(dt)

    def fromutc(self, dt):
        dt_base = dt.replace(tzinfo=self._base_tzinfo)
        dt_out = self._base_tzinfo.fromutc(dt_base)
        return dt_out.replace(tzinfo=self)

    def localize(self, dt, is_dst=False):
        raise_pytz_warning()
        # Some more involved code to map is_dst to fold goes here
        ...
        return dt.replace(tzinfo=self, fold=fold)

    def normalize(self, dt):
        raise_pytz_warning()
        return dt.astimezone(self)
```

Where `raise_pytz_warning()` would raise a warning telling people to stop calling `localize` and `normalize`, as the wrapper class will be dropped in a future version.

Problem 3 is the hardest to solve and *also* is likely to affect you regardless of whether or not you choose to adopt `backports.zoneinfo`. There's no public interface on any time zone provider that allows for this, so I think your options are:

1. Deprecate and then remove this function entirely.
2. Make the case for why this functionality is important, in which case we can add it to `zoneinfo` for Python 3.10, and add it to `backports.zoneinfo` now.
3. Write a TZif parser (this is actually easier than it sounds), and try to find the corresponding TZif file (this is fragile, unfortunately) to get the next transition.
4. Don't use *any* existing time zone implementations and write your own that exposes this.

ds-cbo:
Fixes #716

### DIFF
--- a/babel/dates.py
+++ b/babel/dates.py
@@ -18,7 +18,12 @@
 
 import re
 import warnings
-import pytz as _pytz
+
+try:
+    import zoneinfo
+except ModuleNotFoundError:
+    zoneinfo = None
+    import pytz as _pytz
 
 from datetime import date, datetime, time, timedelta
 from bisect import bisect_right
@@ -42,6 +47,23 @@ LC_TIME = default_locale('LC_TIME')
 date_ = date
 datetime_ = datetime
 time_ = time
+
+
+def localize(tz, dt):
+    """Support localizing with both pytz and zoneinfo tzinfos"""
+    # nothing to do
+    if dt.tzinfo is tz:
+        return dt
+
+    if hasattr(tz, 'localize'):  # pytz
+        return tz.localize(dt)
+    elif dt.tzinfo is None:
+        # convert naive to localized
+        return dt.replace(tzinfo=tz)
+    else:
+        # convert timezones
+        return dt.astimezone(tz)
+
 
 
 def _get_dt_and_tzinfo(dt_or_tzinfo):
@@ -138,7 +160,7 @@ def _ensure_datetime_tzinfo(datetime, tzinfo=None):
 
     If a tzinfo is passed in, the datetime is normalized to that timezone.
 
-    >>> _ensure_datetime_tzinfo(datetime(2015, 1, 1)).tzinfo.zone
+    >>> _get_tz_name(_ensure_datetime_tzinfo(datetime(2015, 1, 1)))
     'UTC'
 
     >>> tz = get_timezone("Europe/Stockholm")
@@ -172,8 +194,10 @@ def _get_time(time, tzinfo=None):
         time = datetime.utcnow()
     elif isinstance(time, (int, float)):
         time = datetime.utcfromtimestamp(time)
+
     if time.tzinfo is None:
         time = time.replace(tzinfo=UTC)
+
     if isinstance(time, datetime):
         if tzinfo is not None:
             time = time.astimezone(tzinfo)
@@ -187,23 +211,34 @@ def _get_time(time, tzinfo=None):
 
 def get_timezone(zone=None):
     """Looks up a timezone by name and returns it.  The timezone object
-    returned comes from ``pytz`` and corresponds to the `tzinfo` interface and
-    can be used with all of the functions of Babel that operate with dates.
+    returned comes from ``zoneinfo`` is available or ``pytz`` if not.
+    It corresponds to the `tzinfo` interface and can be used with all of
+    the functions of Babel that operate with dates.
 
     If a timezone is not known a :exc:`LookupError` is raised.  If `zone`
     is ``None`` a local zone object is returned.
 
     :param zone: the name of the timezone to look up.  If a timezone object
-                 itself is passed in, mit's returned unchanged.
+                 itself is passed in, it's returned unchanged.
     """
     if zone is None:
         return LOCALTZ
     if not isinstance(zone, str):
         return zone
-    try:
-        return _pytz.timezone(zone)
-    except _pytz.UnknownTimeZoneError:
-        raise LookupError(f"Unknown timezone {zone}")
+
+    if zoneinfo:
+        try:
+            return zoneinfo.ZoneInfo(zone)
+        except zoneinfo.ZoneInfoNotFoundError:
+            pass
+
+    else:
+        try:
+            return _pytz.timezone(zone)
+        except _pytz.UnknownTimeZoneError:
+            pass
+
+    raise LookupError(f"Unknown timezone {zone}")
 
 
 def get_next_timezone_transition(zone=None, dt=None):
@@ -456,7 +491,7 @@ def get_timezone_gmt(datetime=None, width='long', locale=LC_TIME, return_z=False
     >>> get_timezone_gmt(dt, locale='en', width='iso8601_short')
     u'+00'
     >>> tz = get_timezone('America/Los_Angeles')
-    >>> dt = tz.localize(datetime(2007, 4, 1, 15, 30))
+    >>> dt = localize(tz, datetime(2007, 4, 1, 15, 30))
     >>> get_timezone_gmt(dt, locale='en')
     u'GMT-07:00'
     >>> get_timezone_gmt(dt, 'short', locale='en')
@@ -588,7 +623,7 @@ def get_timezone_name(dt_or_tzinfo=None, width='long', uncommon=False,
     u'PST'
 
     If this function gets passed only a `tzinfo` object and no concrete
-    `datetime`,  the returned display name is indenpendent of daylight savings
+    `datetime`,  the returned display name is independent of daylight savings
     time. This can be used for example for selecting timezones, or to set the
     time of events that recur across DST changes:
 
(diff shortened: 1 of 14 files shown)


babel/dates.py
```diff
@@ -18,7 +18,12 @@
 
 import re
 import warnings
-import pytz as _pytz
+
+try:
+    import zoneinfo
+except ModuleNotFoundError:
+    zoneinfo = None
+    import pytz as _pytz
```
akx, line 26: I think we should support both zoneinfo and pytz, just in case someone is relying on pytz behavior even on Python 3.9+ – they could just install `pytz` and Babel would prefer it and work like before?

While we're at it, could get rid of the `_pytz` alias:

```suggestion
try:
    import pytz
except ModuleNotFoundError:
    pytz = None

try:
    import zoneinfo
except ModuleNotFoundError:
    zoneinfo = None
```

babel/dates.py
```diff
@@ -187,23 +211,34 @@ def _get_time(time, tzinfo=None):
...
+        except zoneinfo.ZoneInfoNotFoundError:
+            pass
+
+    else:
+        try:
+            return _pytz.timezone(zone)
+        except _pytz.UnknownTimeZoneError:
+            pass
+
+    raise LookupError(f"Unknown timezone {zone}")
```
akx, line 241: In accordance with the above "support both pytz and zoneinfo" thing, this should prefer pytz... and we might as well keep track of the original exception:
```suggestion
    exc = None
    if pytz:
        try:
            return pytz.timezone(zone)
        except pytz.UnknownTimeZoneError as exc:
            pass

    if zoneinfo:
        try:
            return zoneinfo.ZoneInfo(zone)
        except zoneinfo.ZoneInfoNotFoundError as exc:
            pass

    raise LookupError(f"Unknown timezone {zone}") from exc
```

babel/localtime/_helpers.py
```diff
@@ -0,0 +1,43 @@
+try:
+    import zoneinfo
+except ModuleNotFoundError:
+    zoneinfo = None
+    import pytz
```
akx, line 5: Same thing here, let's support both:
```suggestion
try:
    import pytz
except ModuleNotFoundError:
    pytz = None

try:
    import zoneinfo
except ModuleNotFoundError:
    zoneinfo = None
```

babel/localtime/_helpers.py
```diff
@@ -0,0 +1,43 @@
...
+        try:
+            return zoneinfo.ZoneInfo(tzenv)
+        except zoneinfo.ZoneInfoNotFoundError:
+            pass
+
+    else:
+        try:
+            return pytz.timezone(tzenv)
+        except pytz.UnknownTimeZoneError:
+            pass
```
akx, line 24: Same here, let's prefer pytz, then go for zoneinfo:
```suggestion
    if pytz:
        try:
            return pytz.timezone(tzenv)
        except pytz.UnknownTimeZoneError:
            pass

    if zoneinfo:
        try:
            return zoneinfo.ZoneInfo(tzenv)
        except zoneinfo.ZoneInfoNotFoundError:
            pass
```

babel/localtime/_helpers.py
```diff
@@ -0,0 +1,43 @@
...
+
+    return None
+
+def _get_tzinfo_or_raise(tzenv):
+    tzinfo = _get_tzinfo(tzenv)
+    if tzinfo is None:
+        raise LookupError(
+            f"Can not find timezone {tzenv}. \n"
+            "Please use a timezone in the form of Continent/City"
```
akx, line 33: Maybe something like
```suggestion
            f"Can not find timezone {tzenv}. "
            "Timezone names are generally in the form `Continent/City`."
```
because there are `Etc/` timezones that aren't continent-y.

babel/localtime/_helpers.py
```diff
@@ -0,0 +1,43 @@
...
+        )
+    return tzinfo
+
+
+def _get_tzinfo_from_file(tzfilename):
+    with open(tzfilename, 'rb') as tzfile:
+        if zoneinfo:
+            return zoneinfo.ZoneInfo.from_file(tzfile)
+        else:
+            return pytz.tzfile.build_tzinfo('local', tzfile)
```
akx, line 43:
```suggestion
        if pytz:
            return pytz.tzfile.build_tzinfo('local', tzfile)
        if zoneinfo:
            return zoneinfo.ZoneInfo.from_file(tzfile)
        raise RuntimeError("pytz or zoneinfo not available")
```

docs/dev.rst
```diff
@@ -46,10 +46,11 @@ Unicode is a big deal in Babel.  Here is how the rules are set up:
...
 
-Generally all timezone support in Babel is based on pytz which it just
-depends on.  Babel should assume that timezone objects are pytz based
-because those are the only ones with an API that actually work correctly
-(due to the API problems with non UTC based timezones).
+Generally all timezone support in Babel is based on zoneinfo, with pytz
+supported as optional dependency. Babel should assume that timezone objects
+are either zoneinfo- or pytz-based because those are the only ones with an
+API that actually works correctly (due to the API problems with non UTC based
+timezones).
```
akx, line 53: In accordance with the above changes:

```suggestion
Babel's timezone support relies on either ``pytz`` or ``zoneinfo``; if ``pytz``
is installed, it is preferred over ``zoneinfo``.  Babel should assume that any
timezone objects can be from either of these modules.
```

docs/dates.rst
```diff
@@ -276,11 +276,11 @@ class, which you need appropriate implementations for to actually use in your
...
-Babel uses `pytz`_ for real timezone support which includes the
-definitions of practically all of the time-zones used on the world, as
-well as important functions for reliably converting from UTC to local
-time, and vice versa.  The module is generally wrapped for you so you can
-directly interface with it from within Babel:
+Babel uses either `zoneinfo`_ or `pytz`_ for real timezone support which
+include the definitions of practically all of the time-zones used on the world,
+as well as important functions for reliably converting from UTC to local time,
+and vice versa.  The module is generally wrapped for you so you can directly
+interface with it from within Babel:
```
akx, line 283:
```suggestion
Babel uses either `zoneinfo`_ or `pytz`_ for timezone support. If pytz is installed,
it is preferred over the standard library's zoneinfo.

You can directly interface with either of these modules from within Babel:
```

docs/installation.rst
```diff
@@ -4,10 +4,13 @@ Installation
 ============
 
 Babel is distributed as a standard Python package fully set up with all
-the dependencies it needs.  It primarily depends on the excellent `pytz`_
-library for timezone handling.  To install it you can use ``pip``.
+the dependencies it needs.  It optionally depends on the `pytz`_ library
+for timezone handling on Python versions where `zoneinfo`_ is not available.
+To install it you can use ``pip``.
```
akx, line 9:
```suggestion
the dependencies it needs.  On Python versions where the standard library`zoneinfo`_ 
module is not available, `pytz`_  needs to be installed for timezone support.
If `pytz`_  is installed, it is preferred over the standard library `zoneinfo`_  module
where possible.
```

tests/test_dates.py
```diff
@@ -16,27 +16,38 @@
...
+        zoneinfo = None
+
+try:
+    import pytz
+except ModuleNotFoundError:
+    pytz = None
 
 from babel import dates, Locale
-from babel.dates import NO_INHERITANCE_MARKER
+from babel.dates import NO_INHERITANCE_MARKER, get_timezone, localize, _get_tz_name, LOCALTZ
```
akx, line 35: Hmm... I don't think we should use `get_timezone` in these tests at all, but use the `timezone_getter` fixture wherever we would?

akx (changes requested):
Thanks for the PR! ❤️ 

Some change requests, a big one being that `pytz` should IMO be supported alongside `zoneinfo` and preferred where available for backward compatibility reasons.
I'm all ears if there's something I'm missing and that'd be a terrible idea. :D

akx (approved):
Looks good, thank you! 😄

PR #945, opened 2023-01-11
Unify datetime imports

akx:
Various modules in Babel imported `datetime` in various ways, sometimes needing aliases and whatnot; this commit unifies everything to `import datetime`.

### DIFF
--- a/babel/support.py
+++ b/babel/support.py
@@ -71,7 +71,7 @@ class Format:
 
     def datetime(
         self,
-        datetime: datetime.date | datetime.datetime | None = None,
+        datetime: datetime.date | None = None,
         format: _PredefinedTimeFormat | str = 'medium',
     ) -> str:
         """Return a date and time formatted according to the given pattern.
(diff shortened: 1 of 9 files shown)


akx:
@DenverCoder1, could you take a reviewy look? :)

babel/support.py
```diff
@@ -71,18 +69,25 @@ def date(self, date: _date | None = None, format: _PredefinedTimeFormat | str =
         """
         return format_date(date, format, locale=self.locale)
 
-    def datetime(self, datetime: _date | None = None, format: _PredefinedTimeFormat | str = 'medium') -> str:
+    def datetime(
+        self,
+        datetime: datetime.date | datetime.datetime | None = None,
```
DenverCoder1, line 74: Doesn't really need both date and datetime since datetime is a subclass of date
```py
>>> issubclass(datetime.datetime, datetime.date)
True
```
```suggestion
        datetime: datetime.date | None = None,
```

DenverCoder1 (commented):
Looks good to me

Issue #946, opened 2023-01-17
Do not discard locale modifiers (@variants)


madduck:
Locale modifiers ("<span>@</span>variants") are described in the [GNU gettext documentation](https://www.gnu.org/software/gettext/manual/html_node/Locale-Names.html) like this:

> The ‘<span>@</span>variant’ can denote any kind of characteristics that is not already implied by the language ll and the country CC. […] It can also denote a dialect of the language, …

This seems useful for various purposes, and yet, the Babel source code [completely discards](https://github.com/python-babel/babel/blob/master/babel/core.py#L1164) the modifiers, saying "we don't care about them":

```
In [1]: import babel

In [2]: babel.parse_locale('de_DE@formal')
Out[2]: ('de', 'DE', None, None)
```

This is especially problematic, because "private use subtags", as [defined by RFC4646](https://datatracker.ietf.org/doc/html/rfc4646.html#section-2.2.7) are not properly handled by Babel, and neither are the [private use region subtags 'AA', 'ZZ', and in the ranges 'QM'-'QZ' and 'XA'-'XZ'](https://datatracker.ietf.org/doc/html/rfc4646.html#section-4.5), which basically leaves an implementer no choice to extend translations beyond the strict scope of the provided `locale-data`.

The RFC makes it clear that private use subtags and private region subtags aren't primary choice mechanisms, and should only be used where no other approach suffices.

Python's own `gettext` handles locale modifiers just like you'd expect, and while it seems that the `Locale` class would need to be modified to represent modifiers, there appears no reason why "we should not care" about these.

Please do not discard them. I am happy to have a shot at providing a PR if replies here are encouraging.

akx:
Hi, thanks for chiming in!

One issue I see here is that Babel does currently not have support for custom locales (#454), and `de_DE@formal` sounds like one (and so do private use tags), i.e. any `Locale.parse()` (which is an unfortunate name for a function that actually loads a locale, but that's a future Babel 3 issue) for such will fail with an `UnknownLocaleError`.

For the time being, to allow this (very valid!) use, I think

* `parse_locale` should be extended to keep the modifier
* locale loading should be taught to fail if a parsed locale has a modifier
* the translations mechanism should be extended/loosened to support well-formed locale identifiers that might not have locale data available in Babel.

If you're willing to give a PR a shot, please do!

madduck:
Thanks for your kind reply, @akx.

[...]

PR #947, opened 2023-01-18
Keep @modifiers when parsing locales

madduck:
Locale modifiers ("@variants") are described in the GNU gettext documentation like this:

> The ‘@variant’ can denote any kind of characteristics that is not
> already implied by the language ll and the country CC. […] It can also
> denote a dialect of the language, …

Wherein Babel previously would discard these, this patch stores the modifier information in the `Locale` objects, handling string representation accordingly.

Not implemented is the lookup of a meaningful description of modifiers, but instead — for now — an identity mapping is provided.

All tests pass. Furthermore, integration with `flask-babel` has been verified to work.

Resolves: #946
Signed-off-by: martin f. krafft <madduck@madduck.net>

### DIFF
--- a/babel/core.py
+++ b/babel/core.py
@@ -1245,6 +1245,9 @@ def get_locale_identifier(tup: tuple[str, str | None, str | None, str | None, st
 
     >>> get_locale_identifier(('de', 'DE', None, '1999', 'custom'))
     'de_DE_1999@custom'
+    >>> get_locale_identifier(('fi', None, None, None, 'custom'))
+    'fi@custom'
+
 
     .. versionadded:: 1.0
 
--- a/babel/core.py
+++ b/babel/core.py
@@ -182,7 +182,7 @@ class Locale:
         :param territory: the territory (country or region) code
         :param script: the script code
         :param variant: the variant code
-        :param modifier: a modifier ('@variant')
+        :param modifier: a modifier (following the '@' symbol, sometimes called '@variant')
         :raise `UnknownLocaleError`: if no locale data is available for the
                                      requested locale
         """
--- a/babel/core.py
+++ b/babel/core.py
@@ -457,7 +457,7 @@ class Locale:
             if self.variant:
                 details.append(locale.variants.get(self.variant))
             if self.modifier:
-                details.append(locale.modifiers.get(self.modifier))
+                details.append(self.modifier)
             details = filter(None, details)
             if details:
                 retval += f" ({', '.join(details)})"
@@ -584,24 +584,6 @@ class Locale:
         u'Alte deutsche Rechtschreibung'
         """
         return self._data['variants']
-
-    @property
-    def modifiers(self) -> localedata.LocaleDataDict:
-        """Identity mapping of modifiers with "@" prefixed (Temporary implementation)
-
-        TODO: This is not yet implemented, as it would need modification of the
-        locale_data files, so instead it just returns the key for now, with "@"
-        prefixed.
-
-        >>> Locale('de', 'DE').modifiers['euro']
-        u'@euro'
-        """
-
-        class IdentityDict(localedata.LocaleDataDict):
-            def __getitem__(self, key: str) -> Any:
-                return f'@{key!s}'
-
-        return IdentityDict(self._data)
 
     # { Number Formatting
 
--- a/babel/core.py
+++ b/babel/core.py
@@ -1174,6 +1174,8 @@ def parse_locale(
     ('it', 'IT', None, None, 'euro')
     >>> parse_locale('it_IT@custom')
     ('it', 'IT', None, None, 'custom')
+    >>> parse_locale('it_IT@')
+    ('it', 'IT', None, None, None)
 
     The default component separator is "_", but a different separator can be
     specified using the `sep` parameter. Note that an optional modifier is
@@ -1206,10 +1208,7 @@ def parse_locale(
     :raise `ValueError`: if the string does not appear to be a valid locale
                          identifier
     """
-    modifier = None
-    if '@' in identifier:
-        identifier, modifier = identifier.split('@', 1)
-
+    identifier, _, modifier = identifier.partition('@')
     if '.' in identifier:
         # this is probably the charset/encoding, which we don't care about
         identifier = identifier.split('.', 1)[0]
@@ -1238,7 +1237,7 @@ def parse_locale(
     if parts:
         raise ValueError(f"{identifier!r} is not a valid locale identifier")
 
-    return lang, territory, script, variant, modifier
+    return lang, territory, script, variant, (modifier or None)
 
 
 def get_locale_identifier(tup: tuple[str, str | None, str | None, str | None, str | None], sep: str = '_') -> str:
--- a/babel/core.py
+++ b/babel/core.py
@@ -358,7 +358,11 @@ class Locale:
         # implement ICU like fuzzy locale objects and provide a way to
         # maximize and minimize locale tags.
 
-        language, territory, script, variant, modifier = parts
+        if len(parts) == 5:
+            language, territory, script, variant, modifier = parts
+        else:
+            language, territory, script, variant = parts
+            modifier = None
         language = get_global('language_aliases').get(language, language)
         territory = get_global('territory_aliases').get(territory, (territory,))[0]
         script = get_global('script_aliases').get(script, script)
@@ -383,7 +387,12 @@ class Locale:
         # simplified identifier that is just the language
         likely_subtag = get_global('likely_subtags').get(language)
         if likely_subtag is not None:
-            language2, _, script2, variant2, modifier2 = parse_locale(likely_subtag)
+            parts2 = parse_locale(likely_subtag)
+            if len(parts2) == 5:
+                language2, _, script2, variant2, modifier2 = parse_locale(likely_subtag)
+            else:
+                language2, _, script2, variant2 = parse_locale(likely_subtag)
+                modifier2 = None
             locale = _try_load_reducing((language2, territory, script2, variant2, modifier2))
             if locale is not None:
                 return locale
@@ -440,7 +449,7 @@ class Locale:
         Modifiers are currently passed through verbatim:
 
         >>> Locale('it', 'IT', modifier='euro').get_display_name('en')
-        u'Italian (Italy, @euro)'
+        u'Italian (Italy, euro)'
 
         :param locale: the locale to use
         """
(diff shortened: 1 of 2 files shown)


babel/core.py
```diff
@@ -191,10 +193,13 @@ def __init__(
         self.script = script
         #: the variant code
         self.variant = variant
+        #: the modifier
+        self.modifier = modifier
         self.__data = None
 
         identifier = str(self)
-        if not localedata.exists(identifier):
+        withoutmodifier = identifier.split('@', 1)[0]
```
akx, line 201: Can you move this to a property, e.g. `identifier_without_modifier`? I think we'd need to use this in the `_data` property too.

When you do that, can you also make it use `partition` instead:

```suggestion
        without_modifier = identifier.partition('@')[0]
```

babel/core.py
```diff
@@ -289,6 +294,11 @@ def parse(
         >>> Locale.parse('und_AT')
         Locale('de', territory='AT')
 
+        Modifiers are optional, and always at the end, separated by "@":
```
akx, line 297: Might be clearer as
```suggestion
        An optional modifier can be added at the end, separated by "@".
```

babel/core.py
```diff
@@ -562,6 +582,24 @@ def variants(self) -> localedata.LocaleDataDict:
...
+        >>> Locale('de', 'DE').modifiers['euro']
+        u'@euro'
+        """
+
+        class IdentityDict(localedata.LocaleDataDict):
+            def __getitem__(self, key: str) -> Any:
+                return f'@{key!s}'
+
+        return IdentityDict(self._data)
+
```
akx, line 602: I'm not sure this needs to be implemented at all yet?

babel/core.py
```diff
@@ -1112,26 +1150,32 @@ def negotiate_locale(preferred: Iterable[str], available: Iterable[str], sep: st
     return None
 
 
-def parse_locale(identifier: str, sep: str = '_') -> tuple[str, str | None, str | None, str | None]:
+def parse_locale(identifier: str, sep: str = '_') \
+        -> tuple[str, str | None, str | None, str | None, str | None]:
```
akx, line 1154:
```suggestion
def parse_locale(
    identifier: str,
    sep: str = '_',
) -> tuple[str, str | None, str | None, str | None, str | None]:
```

babel/core.py
```diff
@@ -1112,26 +1150,32 @@ def negotiate_locale(preferred: Iterable[str], available: Iterable[str], sep: st
...
     >>> parse_locale('en_150')
-    ('en', '150', None, None)
+    ('en', '150', None, None, None)
     >>> parse_locale('en_us_posix')
-    ('en', 'US', None, 'POSIX')
+    ('en', 'US', None, 'POSIX', None)
+    >>> parse_locale('it_IT@euro')
+    ('it', 'IT', None, None, 'euro')
+    >>> parse_locale('it_IT@custom')
+    ('it', 'IT', None, None, 'custom')
```
akx, line 1171: I'd like to be sure this parses an empty modifier as None:
```suggestion
    >>> parse_locale('it_IT@custom')
    ('it', 'IT', None, None, 'custom')
    >>> parse_locale('it_IT@')
    ('it', 'IT', None, None, None)
```

babel/core.py
```diff
@@ -1112,26 +1150,32 @@ def negotiate_locale(preferred: Iterable[str], available: Iterable[str], sep: st
...
 
     The default component separator is "_", but a different separator can be
-    specified using the `sep` parameter:
+    specified using the `sep` parameter. Note that an optional modifier is
+    always appended and separated with "@":
```
akx, line 1175:
```suggestion
    specified using the `sep` parameter.
    
    The optional modifier component is always separated with "@":
```

babel/core.py
```diff
@@ -1112,26 +1150,32 @@ def negotiate_locale(preferred: Iterable[str], available: Iterable[str], sep: st
...
 
     >>> parse_locale('zh-CN', sep='-')
-    ('zh', 'CN', None, None)
+    ('zh', 'CN', None, None, None)
```
akx, line 1178:
```suggestion
    ('zh', 'CN', None, None, None)
    >>> parse_locale('zh-CN@custom', sep='-')
    ('zh', 'CN', None, None, 'custom')
```

babel/core.py
```diff
@@ -1158,13 +1201,13 @@ def parse_locale(identifier: str, sep: str = '_') -> tuple[str, str | None, str
     :raise `ValueError`: if the string does not appear to be a valid locale
                          identifier
     """
+    modifier = None
+    if '@' in identifier:
+        identifier, modifier = identifier.split('@', 1)
```
akx, line 1206: You can use `partition` here too.
```suggestion
    identifier, _, modifier = identifier.partition('@')
```

babel/core.py
```diff
@@ -1190,22 +1233,23 @@ def parse_locale(identifier: str, sep: str = '_') -> tuple[str, str | None, str
     if parts:
         raise ValueError(f"{identifier!r} is not a valid locale identifier")
 
-    return lang, territory, script, variant
+    return lang, territory, script, variant, modifier
```
akx, line 1236: If you use `partition`:
```suggestion
    return lang, territory, script, variant, (modifier or None)
```

babel/core.py
```diff
@@ -1190,22 +1233,23 @@ def parse_locale(identifier: str, sep: str = '_') -> tuple[str, str | None, str
...
 
 
-def get_locale_identifier(tup: tuple[str, str | None, str | None, str | None], sep: str = '_') -> str:
+def get_locale_identifier(tup: tuple[str, str | None, str | None, str | None, str | None], sep: str = '_') -> str:
```
akx, line 1239: This function should also support 4-tuples. In fact, it would be best if it supported any tuple size from 1 to 5. (I think the code already does, but the signature doesn't reflect that.)

babel/core.py
```diff
@@ -1190,22 +1233,23 @@ def parse_locale(identifier: str, sep: str = '_') -> tuple[str, str | None, str
...
     .. versionadded:: 1.0
 
     :param tup: the tuple as returned by :func:`parse_locale`.
     :param sep: the separator for the identifier.
     """
-    tup = tuple(tup[:4])
-    lang, territory, script, variant = tup + (None,) * (4 - len(tup))
-    return sep.join(filter(None, (lang, script, territory, variant)))
+    tup = tuple(tup[:5])
+    lang, territory, script, variant, modifier = tup + (None,) * (4 - len(tup))
```
akx, line 1253: I think the `4` here is off by one?

```suggestion
    lang, territory, script, variant, modifier = tup + (None,) * (5 - len(tup))
```

That said, that's quite the horrible expression...

babel/core.py
```diff
@@ -180,6 +181,7 @@ def __init__(
         :param territory: the territory (country or region) code
         :param script: the script code
         :param variant: the variant code
+        :param modifier: a modifier ('@variant')
```
akx, line 184: Does this imply the modifier has the `@` intact here? It shouldn't, IMO..?

akx (changes requested):
Thank you for the contribution!
Some initial thoughts and comments within; you'll probably also need to rebase since #949 just landed...

babel/core.py
madduck: I can make this a property, sure, but I didn't, because `identifier` also isn't one.

I am unsure about `_data`. I didn't really dive this far on purpose because I didn't invest time trying to figure out where all the `*.dat` files are being created from.

As pertaining to the use of `partition`: I learnt something new. In this case, however, I don't think it matters much, but I'll happily change it.

babel/core.py
madduck: Yeah, neither. I figure it doesn't hurt and provides a bit of an anchor to keep working on this, but the truth is that we probably will need to either remove it, or tap into the creation of `*.dat` files.

babel/core.py
madduck: No discussion. I was running `black` on the code, and it would have done more or less the same, but it also changed 350 other lines, so I hit 'undo' ;)

babel/core.py
madduck: Makes sense! Thanks.

babel/core.py
madduck: Sweet! I learnt something.

babel/core.py
madduck: Yeah, related to the change above. Makes tests pass.

babel/core.py
madduck: Done.

babel/core.py
madduck: Cool.

babel/core.py
madduck: Hm, I am not aware of a way to represent this in the function signature. Yes, I think the code does support it already, but I am unsure how else you want the signature to look. Any further hints?

babel/core.py
madduck: Good catch!

I cannot really think of a better way, other than to replace the tuple with e.g. a `collections.namedtuple` or otherwise class of its own.

babel/core.py
madduck: I agree. I was studying the docs, and often saw '@variant', which I thought would be read as "at-variant", and so I included it. But it should probably go. Done.

madduck:
@akx Thank you for your time. I addressed most of your comments, and left an open question in the above. Happy to proceed any way, just let me know what you prefer.

babel/core.py
akx: Let's go with removing it for now, please? That way we can come up with a proper API later and aren't stuck freezing this as a public API :)

babel/core.py
akx: I think it'll be a bit unwieldy, but...

```suggestion
def get_locale_identifier(
    tup: tuple[str]
    | tuple[str, str | None]
    | tuple[str, str | None, str | None]
    | tuple[str, str | None, str | None, str | None]
    | tuple[str, str | None, str | None, str | None, str | None],
    sep: str = "_",
) -> str:
```

babel/core.py
```diff
@@ -1193,22 +1240,23 @@ def parse_locale(identifier: str, sep: str = '_') -> tuple[str, str | None, str
...
+def get_locale_identifier(tup: tuple[str, str | None, str | None, str | None, str | None], sep: str = '_') -> str:
     """The reverse of :func:`parse_locale`.  It creates a locale identifier out
-    of a ``(language, territory, script, variant)`` tuple.  Items can be set to
+    of a ``(language, territory, script, variant, modifier)`` tuple.  Items can be set to
     ``None`` and trailing ``None``\\s can also be left out of the tuple.
 
-    >>> get_locale_identifier(('de', 'DE', None, '1999'))
-    'de_DE_1999'
+    >>> get_locale_identifier(('de', 'DE', None, '1999', 'custom'))
+    'de_DE_1999@custom'
```
akx, line 1252: Can you make sure this works with elided bits in the middle too?
```suggestion
    >>> get_locale_identifier(('de', 'DE', None, '1999', 'custom'))
    'de_DE_1999@custom'
    >>> get_locale_identifier(('fi', None, None, None, 'custom'))
    'fi@custom'
```

babel/core.py
```diff
@@ -1193,22 +1240,23 @@ def parse_locale(identifier: str, sep: str = '_') -> tuple[str, str | None, str
     if parts:
         raise ValueError(f"{identifier!r} is not a valid locale identifier")
 
-    return lang, territory, script, variant
+    return lang, territory, script, variant, (modifier or None)
```
akx, line 1243: Actually, can you change this function to only return a 5-tuple if there is a modifier, and then add a `# TODO(3.0): always return a 5-tuple` sort of comment? We're trying pretty hard to not change public APIs.

babel/core.py
```diff
@@ -1193,22 +1222,37 @@ def parse_locale(identifier: str, sep: str = '_') -> tuple[str, str | None, str
...
 
-    return lang, territory, script, variant
-
-
-def get_locale_identifier(tup: tuple[str, str | None, str | None, str | None], sep: str = '_') -> str:
+    # TODO(3.0): always return a 5-tuple
+    if modifier:
+        return lang, territory, script, variant, modifier
+    else:
+        return lang, territory, script, variant
```
akx, line 1229: No need for the `else` :)
```suggestion
    # TODO(3.0): always return a 5-tuple
    if modifier:
        return lang, territory, script, variant, modifier
    return lang, territory, script, variant
```

akx (changes requested):
Tests are failing, can you take a look? Thanks.

babel/core.py
madduck: True, but I personally find it clearer that way. But happy to go without if you insist.

madduck:
Sorry for the failing tests, was due to the 4/5 tuple return depending on `modifier==None`. I am not actually very sure it's a good idea, but I see your point. Fixed now.

akx (approved):
Thanks, looks good!

PR #953, opened 2023-01-20
Update test dependencies (tox v4)

DenverCoder1:
Since tox-gh-actions has a stable release supporting tox v4, both dependencies can be updated.

To test with tox v4, `whitelist_externals` needs to be replaced with `allowlist_externals`.

Also updates codecov/codecov-action from v2 to v3.

* https://tox.wiki/en/latest/upgrading.html
* https://github.com/codecov/codecov-action#migration-from-v1-to-v3

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -40,11 +40,11 @@ jobs:
     - name: Install dependencies
       run: |
         python -m pip install --upgrade pip setuptools wheel
-        python -m pip install 'tox<4.0.0' 'tox-gh-actions==2.12.0'
+        python -m pip install 'tox~=4.0' 'tox-gh-actions~=3.0'
     - name: Run test via Tox
       run: tox --skip-missing-interpreters
       env:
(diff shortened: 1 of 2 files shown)


.github/workflows/test.yml
```diff
@@ -40,11 +40,11 @@ jobs:
     - name: Install dependencies
       run: |
         python -m pip install --upgrade pip setuptools wheel
-        python -m pip install 'tox<4.0.0' 'tox-gh-actions==2.12.0'
+        python -m pip install 'tox~=4.0' 'tox-gh-actions~=3.0'
```
akx, line 43: Can you actually switch us to the first-party [tox-dev/tox-gh](https://github.com/tox-dev/tox-gh) instead?

Just requires changing `[gh-actions]` to `[gh]` in the ini file.

akx (changes requested):

akx (approved):
Excellent, thanks!

PR #955, opened 2023-01-20
Support for formatting NaN, Infinity

### LINKED ISSUE #132: format functions fail on infinity
``` python
>>> format_decimal(Decimal('Infinity'))
  File "babel/numbers.py", line 165, in format_decimal
    return pattern.apply(number, locale)
  File "babel/numbers.py", line 563, in apply
    self.frac_prec[1]))
  File "babel/numbers.py", line 380, in bankersround
    a, b = split_number(value)
  File "babel/numbers.py", line 330, in split_number
    for i in range(-exp):
TypeError: bad operand type for unary -: 'str'
```

DenverCoder1:
This came up while I was looking for ways to improve type-safety - specifically that the type of `decimal_tuple.exponent` in `get_decimal_precision()` is `int | Literal['n', 'N', 'F']` and can't be compared with 0 or taken the absolute value of if it is not an int.

This fixes that type issue but I'm making this a separate PR since it also fixes NaN and Infinity cases in other areas, so it won't raise exceptions when formatting.

Fixes #132 

```py
>>> format_decimal(decimal.Decimal('Infinity'))
'∞'
>>> format_decimal(decimal.Decimal('-Infinity'))
'-∞'
>>> format_decimal(decimal.Decimal('NaN'))
'NaN'
```

### DIFF
--- a/babel/numbers.py
+++ b/babel/numbers.py
@@ -126,7 +126,10 @@ def get_currency_name(
     """
     loc = Locale.parse(locale)
     if count is not None:
-        plural_form = loc.plural_form(count)
+        try:
+            plural_form = loc.plural_form(count)
+        except (OverflowError, ValueError):
+            plural_form = 'other'
         plural_names = loc._data['currency_names_plural']
         if currency in plural_names:
             currency_plural_names = plural_names[currency]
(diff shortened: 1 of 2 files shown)


akx (approved):
Looks good! I wonder if there are some parameter combinations that might not handle infinities or NaNs, but this is way better than doing nothing with them 😄

PR #959, opened 2023-01-25
Fix unbound `exc` in babel.dates

akx:
See https://stackoverflow.com/a/24271786/51685

My bad! I had remembered the behavior to be like with Python 2...

### DIFF
--- a/babel/dates.py
+++ b/babel/dates.py
@@ -243,18 +243,17 @@ def get_timezone(zone: str | datetime.tzinfo | None = None) -> datetime.tzinfo:
     if not isinstance(zone, str):
         return zone
 
-    exc = None
     if pytz:
         try:
             return pytz.timezone(zone)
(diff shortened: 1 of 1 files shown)


babel/dates.py
```diff
@@ -243,18 +243,17 @@ def get_timezone(zone: str | datetime.tzinfo | None = None) -> datetime.tzinfo:
...
         return zone
 
-    exc = None
     if pytz:
         try:
             return pytz.timezone(zone)
-        except pytz.UnknownTimeZoneError as exc:  # noqa: F841
-            pass
+        except pytz.UnknownTimeZoneError as e:
+            exc = e
```
DenverCoder1, line 250: If the variable is not used, can't it be removed?
```suggestion
        except pytz.UnknownTimeZoneError:
            pass
```

PR #961, opened 2023-01-25
Make LocaleDataDict Generic and fix core and numbers types

DenverCoder1:
Type annotation improvements:

1. Made `LocaleDataDict` use `TypeVar`s so that the key and value types can be annotated for each instance
2. Changed `Locale.parse` to always return a `Locale` (this fixes type issues in multiple places where the return type is assumed to not be None)
3. Fixed Pyright type errors and added `type: ignore` in certain places in core.py and numbers.py

**Note:** Changing `Locale.parse` to always return a `Locale` simplifies the logic, but it also can potentially be a breaking change if people are relying on passing `None` not raising an exception. It is a bit of a stretch though, since passing `None` has no functional effect in the use of the lib.


akx:
> 2\. Changed `Locale.parse` to always return a `Locale` (this fixes type issues in multiple places where the return type is assumed to not be None)

That's a great change – I don't think it messes up any valid downstream workflows. However seeing as that retval was added here https://github.com/python-babel/babel/commit/8b7c5e1585b52ea110a5e884151d4eaa2175985d in a commit that talks about likely subtag resolving, could you make sure that it doesn't affect that functionality? I think we might need to do `if identifier is None: raise IdentifierNone()` and catch it where there currently are other `if locale is not None`s...

DenverCoder1:
> > 2. Changed `Locale.parse` to always return a `Locale` (this fixes type issues in multiple places where the return type is assumed to not be None)
> 
> That's a great change – I don't think it messes up any valid downstream workflows. However seeing as that retval was added here [8b7c5e1](https://github.com/python-babel/babel/commit/8b7c5e1585b52ea110a5e884151d4eaa2175985d) in a commit that talks about likely subtag resolving, could you make sure that it doesn't affect that functionality? I think we might need to do `if identifier is None: raise IdentifierNone()` and catch it where there currently are other `if locale is not None`s...

As far as I can tell, there aren't any locations in the lib that expect the return type of `Locale.parse` to be anything other than `Locale`. I don't see any places where `parse` is used and then the type is checked to not be `None`.

babel/localedata.py
```diff
@@ -207,33 +209,32 @@ def resolve(self, data: Mapping[str | int | None, Any]) -> Mapping[str | int | N
...
     """
 
-    def __init__(self, data: MutableMapping[str | int | None, Any], base: Mapping[str | int | None, Any] | None = None):
-        self._data = data
+    def __init__(self, data: MutableMapping[_Key, _Value], base: Mapping[_Key, _Value] | None = None):
+        self._data = dict(data)  # Storing as a dict allows copy() to work
         if base is None:
             base = data
-        self.base = base
+        self.base = dict(base)
```
akx, line 229: I have the feeling these `dict()` casts aren't a good idea... won't this imply that the `base` and `data` lose any specific dict type they are (e.g. `LocaleDataDict`) _and_ that there's an additional copy made..?

Should we maybe just do `cast(dict, ...)`?

akx (commented):
I'm a little wary about the LocaleDataDict changes... The numbers changes look alright though, can you break them out into a separate PR?

babel/localedata.py
DenverCoder1: Internally they are probably dicts already.

An alternative for better type safety may be to make an ABC implementing Mapping with a copy method and use that as the annotation, or as a simpler approach, make the parameter annotations use dict

PR #966, opened 2023-02-03
Numbers and core type fixes

DenverCoder1:
Type annotation improvements:

1. Changed `Locale.parse` to always return a `Locale` (this fixes type issues in multiple places where the return type is assumed to not be None)
2. Fixed Pyright type errors and added `type: ignore` in certain places in core.py and numbers.py

**Note:** Changing `Locale.parse` to always return a `Locale` simplifies the logic, but it also can potentially be a breaking change if people are relying on passing `None` not raising an exception. It is a bit of a stretch though, since passing `None` has no functional effect in the use of the lib.

These changes have been separated from #961

### DIFF
--- a/babel/core.py
+++ b/babel/core.py
@@ -13,7 +13,7 @@ from __future__ import annotations
 import os
 import pickle
 from collections.abc import Iterable, Mapping
-from typing import TYPE_CHECKING, Any, overload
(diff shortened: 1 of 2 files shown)


akx (approved):
Looks good, thanks!

PR #968, opened 2023-02-07
Improve extract performance via ignoring directories early during os.walk

akx:
Closes #694 (rebases and reimplements it)

### DIFF
--- a/babel/messages/extract.py
+++ b/babel/messages/extract.py
@@ -23,6 +23,7 @@ import io
 import os
 import sys
 import tokenize
+import warnings
 from collections.abc import (
     Callable,
     Collection,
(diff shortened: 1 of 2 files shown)


akx:
@DenverCoder1 If you have a moment, could you review this? :)

DenverCoder1:
Looks good to me (I haven't tested)

I noticed a similar method in frontend.py - https://github.com/python-babel/babel/blob/373a52f0425d0af1e578e8562a8db93fc3a02488/babel/messages/frontend.py#L277-L288 - it seems to not need to create relative paths there, but maybe it's different. Possibly some logic could be combined or renamed to be consistent, but besides that, it seems like this should work.

babel/messages/extract.py
```diff
@@ -111,10 +111,25 @@ def _strip(line: str):
     comments[:] = map(_strip, comments)
 
 
-def default_directory_filter(dirpath: str | os.PathLike[str]) -> bool:
```
tomasr8, line 114: Would this function be considered part of the public API? If so, shouldn't we deprecate it first?

babel/messages/extract.py
```diff
@@ -111,10 +111,25 @@ def _strip(line: str):
...
+        subdir = os.path.basename(dirpath)
+        # Legacy default behavior: ignore dot and underscore directories
+        if subdir.startswith('.') or subdir.startswith('_'):
+            return False
+
+        dir_rel = os.path.relpath(dirpath, root_dir).replace(os.sep, '/')
+
+        for pattern, method in method_map:
+            if method == "ignore" and pathmatch(pattern, dir_rel):
+                return False
```
tomasr8, line 128: This line's missing test coverage ;)

PR #970, opened 2023-02-12
Fix two issues with fuzzy matching

### LINKED ISSUE #969: Fuzzy matching sometimes misses obvious candidates
## Overview Description

I encountered this while updating catalogs for a documentation project in Sphinx with `sphinx-intl`, which uses Babel. A long msgid with a very small modification had its translation reset (the old one moved to obsolete), instead of being marked fuzzy.

## Steps to Reproduce

Here is a reproducible example:

```python
from io import BytesIO

from babel.messages.catalog import Catalog
from babel.messages.pofile import write_po

msgid = 'As will be later explained in [](parents-bounds), grobs come in several flavors, most importantly and spanners. The grob type normally mandates use as item or spanner. However, it happens that a grob can be used both ways. This is mainly the case for so-called "sticky grobs", which attach to another arbitrary grob, such as footnotes, balloons and parentheses. While many grobs attach to other grobs (e.g., articulations attach to note heads), sticky grobs are special because the grob they attach to, called their "host", is arbitrary and can therefore be either an item or a spanner. In turn, this necessitates creating the sticky grob either as item or spanner depending on the flavor of its host. The following function supports this common case:'

msgstr = "foobar"

catalog = Catalog()
catalog.add(msgid, msgstr)

new_msgid = "As explained in [](grob-flavors)" + msgid.removeprefix("As will be later explained in [](parents-bounds)")

template = Catalog()
template.add(new_msgid)

catalog.update(template)

print(list(catalog))
```

## Actual Results

```
[<Message '' (flags: ['fuzzy'])>, <Message 'As explained in [](grob-flavors), grobs come in several flavors, most importantly and spanners. The grob type normally mandates use as item or spanner. However, it happens that a grob can be used both ways. This is mainly the case for so-called "sticky grobs", which attach to another arbitrary grob, such as footnotes, balloons and parentheses. While many grobs attach to other grobs (e.g., articulations attach to note heads), sticky grobs are special because the grob they attach to, called their "host", is arbitrary and can therefore be either an item or a spanner. In turn, this necessitates creating the sticky grob either as item or spanner depending on the flavor of its host. The following function supports this common case:' (flags: [])>]
```

## Expected Results

The merged catalog has a fuzzy entry rather than an untranslated entry.

## Additional Information

This appears to be caused by https://github.com/python/cpython/issues/90825.

jeanas:
- One issue with case,
- One issue with difflib's “autojunk” heuristic.

Fixes #969

### DIFF
--- a/babel/messages/catalog.py
+++ b/babel/messages/catalog.py
@@ -14,8 +14,9 @@ import re
 from collections import OrderedDict
 from collections.abc import Iterable, Iterator
 from copy import copy
-from difflib import get_close_matches
+from difflib import SequenceMatcher
 from email import message_from_string
+from heapq import nlargest
 from typing import TYPE_CHECKING
 
 from babel import __version__ as VERSION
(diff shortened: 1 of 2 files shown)


akx:
> Is the 0.01% coverage decrease something I should worry about?

Not at all :)

akx (approved):
Thanks, looks good! Sorry for the delayed review.

PR #934, opened 2022-12-04
Added inline type annotations to Babel

DenverCoder1:
This PR adds inline type annotations to the public methods and classes in Babel.

This PR tries to focus on only _adding_ correct type annotations. There are still some Pyright warnings due to non-type-safe implementations which can be fixed at a later time.

Most of the annotations have already been reviewed and approved by [typeshed](https://github.com/python/typeshed).

Tested on Python 3.7 - 3.11

PR #975, opened 2023-02-28
Include `py.typed` file in wheels

### LINKED ISSUE #795: This library should have type annotations

AlexWaygood:
#934 by @DenverCoder1 added inline type annotations and a `py.typed` file, which is great! Unfortunately, however, the `py.typed` file hasn't been included in wheels published to PyPI for v2.12.0. This means that if I `pip install` babel locally, mypy still doesn't recognise `babel` as a typed library. Running mypy on a file with just `import babel` in it results in this output from mypy (unless the typeshed `types-babel` package is installed):

```
test.py:1: error: Library stubs not installed for "babel"  [import]
test.py:1: note: Hint: "python3 -m pip install types-babel"
test.py:1: note: (or run "mypy --install-types" to install all missing stub packages)
test.py:1: note: See https://mypy.readthedocs.io/en/stable/running_mypy.html#missing-imports
Found 1 error in 1 file (checked 1 source file)
```

The fix is simple: just a one-line addition to `setup.py`.

Closes #795 (which should maybe have been closed after #934).

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -58,6 +58,7 @@ setup(
     ],
     python_requires='>=3.7',
     packages=['babel', 'babel.messages', 'babel.localtime'],
+    package_data={"babel": ["py.typed"]},
     include_package_data=True,
     install_requires=[
         # This version identifier is currently necessary as


akx (approved):

AlexWaygood:
Thanks!

> Is `package_data` merged with what `MANIFEST.in` brings in? If not, we'll just want include `babel/py.typed` in the manifest...

I... hope so. If I'm honest, I'm young enough that I've never maintained a project that uses a `MANIFEST.in` file or uses `setuptools` as a build backend (I generally just use `hatchling`).

akx:
@AlexWaygood Thanks again. Released in Babel 2.12.1, now on PyPI.

AlexWaygood:
Confirmed that mypy recognises the library as having inline type annotations with the latest release 🎉

Issue #977, opened 2023-02-28
`Locale.parse()` no longer accepts None as the locale identifier


neuroid:
## Overview Description

`Locale.parse()` no longer accepts None as the locale identifier, [contrary to the type annotation](https://github.com/python-babel/babel/blob/56071c90116e6b9ebbb62ac072bcf032fc6987cb/babel/core.py#L266).

## Steps to Reproduce

```python
>>> from babel import Locale
>>> Locale.parse(None) is None
```

## Actual Results

```python
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/env/lib/python3.8/site-packages/babel/core.py", line 317, in parse
    raise TypeError(f"Unexpected value for identifier: {identifier!r}")
TypeError: Unexpected value for identifier: None
```

## Expected Results

```python
True
```

## Reproducibility

👍 

## Additional Information

Arguably, the previous behaviour was useful. However, if this breaking change is intentional, the type annotation should be adjusted accordingly.

akx:
This change was intentional (but I forgot to distil it into the changelog).

Why would the behaviour be useful?

PR #978, opened 2023-03-01
Note that `Locale.parse()` does not accept `None`

akx:
Refs #977

### DIFF
--- a/babel/core.py
+++ b/babel/core.py
@@ -279,6 +279,15 @@ class Locale:
         >>> Locale.parse(l)
         Locale('de', territory='DE')
 
+        If the `identifier` parameter is neither of these, such as `None`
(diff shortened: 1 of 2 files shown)


babel/core.py
```diff
@@ -263,7 +263,7 @@ def negotiate(
     @classmethod
     def parse(
         cls,
-        identifier: str | Locale | None,
+        identifier: str | Locale,
```
DenverCoder1, line 266: It's a bit complicated here since all the LC_... constants could be None if there is a problem determining the default and they are potentially passed to this method, meaning for type checkers, it sees this function as accepting None, but always raising the exception here.

akx: Great points, thanks! Updated the approach here to docs-only.

DenverCoder1 (approved):

PR #980, opened 2023-03-01
babel.messages.catalog: deduplicate _to_fuzzy_match_key logic

akx:
Follows up on #970.


akx:
@jeanas, care to take a look?

PR #983, opened 2023-03-02
Improve type annotation for `babel.support.Translations.load`

### LINKED ISSUE #982: Type annotation inconsistency
## Overview Description

Babel has inconsistent type annotations (for mypy) in the `babel.support.Translations.load` method.

## Steps to Reproduce

1. The documentation mentions that `babel.support.Translations.load` accepts strings but also Locale objects for its locales argument:
`locales – the list of locales in order of preference (items in this list can be either Locale objects or locale strings)`
2. However the code type annotations do not accept Locale objects:
`classmethod load(..., locales: ...str...) → ...`
3. The library behavior indicates that it is the documentation that is correct - Locale objects are accepted by the code

## Reproducibility

Tested on babel 2.12.1 with python 3.10.9 through mamba 1.3.1 on MacOS 13.2.1

## Additional Information

Documentation available here:
https://babel.pocoo.org/en/latest/api/support.html

akx:
Fixes #982

### DIFF
--- a/babel/support.py
+++ b/babel/support.py
@@ -684,8 +684,11 @@ class Translations(NullTranslations, gettext.GNUTranslations):
 def _locales_to_names(
     locales: Iterable[str | Locale] | str | Locale | None,
 ) -> list[str] | None:
-    """
-    Normalize a `locales` argument to a list of locale names.
+    """Normalize a `locales` argument to a list of locale names.
+
+    :param locales: the list of locales in order of preference (items in
+                    this list can be either `Locale` objects or locale
+                    strings)
     """
     if locales is None:
         return None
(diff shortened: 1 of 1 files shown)


babel/support.py
```diff
@@ -683,3 +679,18 @@ def merge(self, translations: Translations):
...
 
         return self
+
+
+def _locales_to_names(
+    locales: Iterable[str | Locale] | str | Locale | None,
+) -> list[str] | None:
+    """
+    Normalize a `locales` argument to a list of locale names.
+    """
```
DenverCoder1, line 689:
```suggestion
    """Normalize a `locales` argument to a list of locale names.

    :param locales: the list of locales in order of preference (items in
                    this list can be either `Locale` objects or locale
                    strings)
    """
```

DenverCoder1 (approved):

PR #984, opened 2023-03-02
Makefile: use python3 instead of python

bfontaine:
Ubuntu doesn’t have a `python` executable; it’s either `python2` or `python3`.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,5 +1,5 @@
 test: import-cldr
-	python ${PYTHON_TEST_FLAGS} -m pytest ${PYTEST_FLAGS}
+	python3 ${PYTHON_TEST_FLAGS} -m pytest ${PYTEST_FLAGS}
 
 test-env:
 	virtualenv test-env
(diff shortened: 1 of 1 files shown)


bfontaine:
The failure seems to be unrelated:
```
C:\hostedtoolcache\windows\Python\3.10.10\x64\python3.exe: No module named pytest
  mingw[32](https://github.com/python-babel/babel/actions/runs/4315580116/jobs/7530168374#step:6:33)-make: *** [makefile:2: test] Error 1
  py310: exit 2 (9.42 seconds) D:\a\babel\babel> make clean-cldr test pid=6116
```

akx:
> The failure seems to be unrelated:
> 
> ```
> C:\hostedtoolcache\windows\Python\3.10.10\x64\python3.exe: No module named pytest
> ```

I don't think it is unrelated; that's not the tox-created virtualenv's `python`.

bfontaine:
You’re right.

According to https://github.com/tox-dev/tox/issues/2801 on Windows you have to use `python` because `python3` doesn’t exist, while on Debian (and so Ubuntu) you have to use `python3` because `python` doesn’t exist. However, if you create a virtual environment with `virtualenv` or similar, you should have a `python` executable anyway, so this PR is not relevant.

PR #991, opened 2023-03-15
Add pydantic 1.x.x native integration

07pepa:
- __get_validators__ added to locale and DateTimePattern
- added missing validation of incorrect type to parse_pattern
- added pydantic to test dependencies in tox.ini
- wrote test to validate functionality

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -1,6 +1,13 @@
 Babel Changelog
 ===============
 
+
+Unreleased
+----------
+New features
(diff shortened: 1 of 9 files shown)


07pepa:
probably pydantic maintainer should give also some blessing apart from babel mainainer ...  if you can rope someone in it would be awsome  if not, i will try mysef

if they give blessing it may be listed as compatible library somewhere in documentation https://docs.pydantic.dev
or simply as courtesy

PR #994, opened 2023-03-28
Use SPDX license identifier

vargenau:
Use SPDX license identifier: BSD-3-Clause
This will help tools to produce valid SPDX.


akx (approved):
Thanks!

Issue #988, opened 2023-03-12
Tests fail on days when EST/EDT swap happens


mweinelt:
## Overview Description

Timezone related tests fail around when eastern daylight and easter standard time swap.

## Steps to Reproduce

1. Run the tests on the 2nd sunday of march or the 1st sunday in november

## Actual Results

```
______________________ [doctest] babel.dates.format_time _______________________
770     what a given time would translate to in a different timezone without
771     information about whether daylight savings time is in effect or not. This
772     means that time values are left as-is, and the value of the `tzinfo`
773     parameter is only used to display the timezone name if needed:
774 
775     >>> t = time(15, 30)
776     >>> format_time(t, format='full', tzinfo=get_timezone('Europe/Paris'),
777     ...             locale='fr_FR')
778     u'15:30:00 heure normale d\u2019Europe centrale'
779     >>> format_time(t, format='full', tzinfo=get_timezone('US/Eastern'),
Expected:
    u'3:30:00\u202fPM Eastern Standard Time'
Got:
    '3:30:00\u202fPM Eastern Daylight Time'

/build/Babel-2.12.1/babel/dates.py:779: DocTestFailure
___________________ [doctest] babel.dates.get_timezone_name ____________________
536 Return the localized display name for the given timezone. The timezone
537     may be specified using a ``datetime`` or `tzinfo` object.
538 
539     >>> from datetime import time
540     >>> dt = time(15, 30, tzinfo=get_timezone('America/Los_Angeles'))
541     >>> get_timezone_name(dt, locale='en_US')
Expected:
    u'Pacific Standard Time'
Got:
    'Pacific Daylight Time'

/build/Babel-2.12.1/babel/dates.py:541: DocTestFailure
_____________________ test_format_time[zoneinfo.ZoneInfo] ______________________

timezone_getter = <class 'zoneinfo.ZoneInfo'>

    def test_format_time(timezone_getter):
        t = time(15, 30)
        assert dates.format_time(t, locale='en_US') == '3:30:00\u202fPM'
        assert dates.format_time(t, format='short', locale='de_DE') == '15:30'
    
        assert (dates.format_time(t, "hh 'o''clock' a", locale='en') ==
                "03 o'clock PM")
    
        paris = timezone_getter('Europe/Paris')
        eastern = timezone_getter('US/Eastern')
    
        t = _localize(paris, datetime(2007, 4, 1, 15, 30))
[...]

PR #998, opened 2023-04-07
Freeze format_time() tests to a specific date to fix test failures

mgorny:
Freeze the date when performing the tests for format_time() with a timezone specified.  Since the time object does not specify a date, the formatter uses the format string specific to the current date. As a result, if the current DST state is different than when the test was last updated, it failed.

This fix covers only regular tests.  I have no idea how to do the same for doctests.

Issue #988


mgorny:
Ping.

akx (approved):
LGTM, sorry for the delay.

akx:
> I have no idea how to do the same for doctests.

TBH I ended up just `# doctest: +SKIP`ing them in #1017 :)

Issue #725, opened 2020-08-06
Option to skip POT-Creation-Date


Dreamsorcerer:
Could we get an option to skip adding/updating the POT-Creation-Date header when running pybabel extract?

Use case is that we want a pre-commit hook that updates the translations, to ensure our translations are always in sync with the code base. But, because it changes this header every time, it results in a merge conflict with every single branch.

cburmeister:
I would love this!

We've got a GitHub Action step that runs `pybabel extract` on every push to `master` that opens a Pull Request to update the message catalog. This option would allow us to skip creating Pull Requests that only contain changes to the `POT-Creation-Date` (or any other header changes like `Generated-By`, etc).

cburmeister:
@Dreamsorcerer would you mind sharing some of that if possible? I started writing some bash to do this but it got unwieldily fairly quickly 😅

Dreamsorcerer:
Slightly modified, but this should work (if you change `project_path` to the correct directory.
https://gist.github.com/Dreamsorcerer/9355fd2a308121589d7c2fa5ffaa3e8a

We ended up using the extract function in a pre-commit hook to ensure that translations were always in-sync with the code.

cburmeister:
@Dreamsorcerer thank you!

emibcn:
This feature would be very useful. Is there any updates on it?

emibcn:
This feature can be to keep the creation date when updating, change it only on creation.

akx:
PRs are very much welcome :)

joeportela:
@akx Any interest in my PR? https://github.com/python-babel/babel/pull/999

emibcn:
@akx 
When will this be published into a new release?

joeportela:
> @akx When will this be published into a new release?

This just got released in 2.13.0!

emibcn:
@joeportela I have just tested the PO `update` functionality using the new option `--ignore-pot-creation-date` and it looks good (no creation date differences):

![image](https://github.com/python-babel/babel/assets/693791/4403b99a-589f-4ebd-bb07-0b116f299a9b)


What I am missing now is something similar for the POT `extract` functionality:

![image](https://github.com/python-babel/babel/assets/693791/90ea3851-cc21-4c44-8b5c-7b413bc606c0)

StanFromIreland:
There is nothing left to do here and this should not be added for extract, its extracting, not updating.

This should be closed @akx

emibcn:
[...]

PR #999, opened 2023-04-11
Add flag to ignore POT-Creation-Date for updates

joeportela:
Adding a flag to ignore changes to POT-Creation-Date when performing updates or checks. 
I use a CI check that ensures my .po files are updated before merging, but there is always a diff on POT-Creation-Date.

Related to https://github.com/python-babel/babel/issues/725 - but solving (my) problem via update instead of extract.

### DIFF
--- a/babel/messages/catalog.py
+++ b/babel/messages/catalog.py
@@ -769,7 +769,7 @@ class Catalog:
         no_fuzzy_matching: bool = False,
         update_header_comment: bool = False,
         keep_user_comments: bool = True,
-        ignore_pot_creation_date: bool = False,
+        update_creation_date: bool = True,
     ) -> None:
         """Update the catalog based on the given template catalog.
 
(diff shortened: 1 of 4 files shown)


joeportela:
Ping @akx 
Any interest in this PR?

babel/messages/catalog.py
```diff
@@ -769,6 +769,7 @@ def update(
         no_fuzzy_matching: bool = False,
         update_header_comment: bool = False,
         keep_user_comments: bool = True,
+        ignore_pot_creation_date: bool = False,
```
akx, line 772: I think

```suggestion
        update_creation_date: bool = True,
```

would be more intuitive? 

The command line parameter could be left `--ignore-pot-creation-date` though.

akx (changes requested):
Thank you for the contribution! As mentioned inside, I think it could be more intuitive for the kwarg to be positive (and default to True).

joeportela:
> Thank you for the contribution! As mentioned inside, I think it could be more intuitive for the kwarg to be positive (and default to True).

My pleasure, thanks for taking a look! Just updated per your suggestion.

akx (approved):
LGTM, sorry for the delay!

Issue #637, opened 2019-03-19
parse_decimal is not works correctly when space symbol as grouping


llybin:
```
from babel.numbers import format_decimal

format_decimal(12345.123, locale='ru')

'12\xa0345,123'
```
```
from babel.numbers import parse_decimal

parse_decimal('12 345,123', locale='ru')

babel/numbers.py", line 605, in parse_decimal
    .replace(get_decimal_symbol(locale), '.'))
decimal.InvalidOperation: [<class 'decimal.ConversionSyntax'>]
babel.numbers.NumberFormatError
```

There is a problem because ' '.replace('\xa0', '.') is not replace space to dot.

akx:
This is technically by design, but it does make sense to allow a plain space in U+00A0's stead when it is the group separator.

PR #1007, opened 2023-06-08
Allow alternative space characters as group separator when parsing numbers

ronnix:
## Context

The French group separator is `"\u202f"` (narrow non-breaking space), but when parsing numbers in the real world, you will most often encounter either a regular space character (`" "`) or a non-breaking space character (`"\xa0"`).

The issue was partially adressed earlier in https://github.com/python-babel/babel/issues/637, but only to allow regular spaces instead of non-breaking spaces `"\xa0"` in `parse_decimal`.
 
## Contents

This PR goes further by changing both `parse_number` and `parse_decimal` to allow ~~any other space character (using the `\s` character class of regular expressions)~~ any of those 3 space characters when the expected group symbol is itself one such space character, but is not present in the string to parse.

Unit tests are included.

### DIFF
--- a/babel/numbers.py
+++ b/babel/numbers.py
@@ -998,6 +998,15 @@ class NumberFormatError(ValueError):
         self.suggestions = suggestions
 
 
+SPACE_CHARS = {
+    ' ',  # space
+    '\xa0',  # no-break space
+    '\u202f',  # narrow no-break space
+}
+
+SPACE_CHARS_RE = re.compile('|'.join(SPACE_CHARS))
+
+
 def parse_number(
     string: str,
     locale: Locale | str | None = LC_NUMERIC,
@@ -1029,12 +1038,12 @@ def parse_number(
     group_symbol = get_group_symbol(locale, numbering_system=numbering_system)
 
     if (
-        re.match(r'\s', group_symbol) and  # if the grouping symbol is a kind of space,
+        group_symbol in SPACE_CHARS and  # if the grouping symbol is a kind of space,
         group_symbol not in string and  # and the string to be parsed does not contain it,
-        re.search(r'\s', string)  # but it does contain any other kind of space instead,
+        SPACE_CHARS_RE.search(string)  # but it does contain any other kind of space instead,
     ):
         # ... it's reasonable to assume it is taking the place of the grouping symbol.
-        string = re.sub(r'\s', group_symbol, string)
+        string = SPACE_CHARS_RE.sub(group_symbol, string)
 
     try:
         return int(string.replace(group_symbol, ''))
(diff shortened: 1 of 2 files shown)


babel/numbers.py
```diff
@@ -882,8 +882,18 @@ def parse_number(string: str, locale: Locale | str | None = LC_NUMERIC) -> int:
...
     """
+    group_symbol = get_group_symbol(locale)
+
+    if (
+        re.match(r"\s", group_symbol) and  # if the grouping symbol is a kind of space,
+        group_symbol not in string and  # and the string to be parsed does not contain it,
+        re.search(r"\s", string)  # but it does contain any other kind of space instead,
+    ):
+        # ... it's reasonable to assume it is taking the place of the grouping symbol.
+        string = re.sub(r"\s", group_symbol, string)
```
akx, line 893: I'm a little worried about the possible performance issues here. Could 

* this be reordered so the cheapest check is first (`group_symbol not in string and ...`)
* `re.match(r"\s", group_symbol)` moved to an `lru_cache`d helper function – or if `group_symbol.isspace()` is enough, then let's do that instead?

babel/numbers.py
```diff
@@ -930,12 +940,12 @@ def parse_decimal(string: str, locale: Locale | str | None = LC_NUMERIC, strict:
     decimal_symbol = get_decimal_symbol(locale)
 
     if not strict and (
-        group_symbol == '\xa0' and  # if the grouping symbol is U+00A0 NO-BREAK SPACE,
+        re.match(r"\s", group_symbol) and  # if the grouping symbol is a kind of space,
         group_symbol not in string and  # and the string to be parsed does not contain it,
-        ' ' in string  # but it does contain a space instead,
+        re.search(r"\s", string)  # but it does contain any other kind of space instead,
```
akx, line 945: Same perf stuff here. If you end up using a cached helper function, it can well be shared between these two functions.

akx (changes requested):
Thanks for the contribution! Some perf-related comments within.

babel/numbers.py
ronnix: Sorry for the delay, I kind of forgot about this PR.

It looks like`group_symbol.isspace()` should work the same as `re.match(r"\s", group_symbol)`, but I'm thinking they both might be too wide and match things like new lines. Maybe it’s better to be more specific and only match `" "`, `"\xa0"` and `"\u202f"`.

akx: I think using `.isspace()` might be better than hard-coding the set of spaces. I think we're already missing some space-like characters from that set anyway and I don't think `group_symbol` will ever be set to a newline...

akx:
@ronnix Sorry for dropping the ball on my part here... Could you rebase this? :)

ronnix:
@akx done!

ronnix:
@akx I fixed the linter error

akx (approved):
Thanks, LGTM! 😃

PR #1010, opened 2023-06-11
Allow precision specification.

### LINKED ISSUE #893: format_percent doesn't return trailing decimal zeros with precision
## Overview Description
Even when `decimal_quantization=False`, format_percent doesn't return decimal zeros with precision. Is it possible to include another flag to specify the number of precision values? We are using this for localizing interest rates. 
## Steps to Reproduce

1. Call `format_percent` with any number that would yield a whole number. 

## Actual Results
format_percent(0.23, locale=en_US, decimal_quantization=False) -> 23%
format_percent(0.234, locale=en_US, decimal_quantization=False) -> 23.4%

## Expected Results
format_percent(0.23, locale=en_US, decimal_quantization=False) -> 23.00%
format_percent(0.234, locale=en_US, decimal_quantization=False) -> 23.40%

Olunusib:
When bypassing decimal quantization, we should be able to specify precision. Fixes #893.

### DIFF
--- a/babel/numbers.py
+++ b/babel/numbers.py
@@ -1363,4 +1363,3 @@ class NumberPattern:
         while len(value) > min and value[-1] == '0':
             value = value[:-1]
         return get_decimal_symbol(locale) + value
-
--- a/babel/numbers.py
+++ b/babel/numbers.py
@@ -485,7 +485,7 @@ def format_decimal(
         format = locale.decimal_formats[format]
     pattern = parse_pattern(format)
     return pattern.apply(
-        number, locale, decimal_quantization=decimal_quantization,  precision=precision, group_separator=group_separator)
+        number, locale, decimal_quantization=decimal_quantization, precision=precision, group_separator=group_separator)
 
 
 def format_compact_decimal(
@@ -1234,6 +1234,9 @@ class NumberPattern:
             frac_prec = (get_currency_precision(currency), ) * 2
         else:
             frac_prec = self.frac_prec
+
+        if decimal_quantization and precision is not None:
+            raise ValueError("To specify precision, decimal_quantization should be set to False.")
 
         # Bump decimal precision to the natural precision of the number if it
         # exceeds the one we're about to use. This adaptative precision is only
--- a/babel/numbers.py
+++ b/babel/numbers.py
@@ -429,6 +429,7 @@ def format_decimal(
     format: str | NumberPattern | None = None,
     locale: Locale | str | None = LC_NUMERIC,
     decimal_quantization: bool = True,
+    precision: int | None = None,
     group_separator: bool = True,
 ) -> str:
     """Return the given decimal number formatted for a specific locale.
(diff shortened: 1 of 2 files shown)


babel/numbers.py
```diff
@@ -476,7 +485,7 @@ def format_decimal(
         format = locale.decimal_formats[format]
     pattern = parse_pattern(format)
     return pattern.apply(
-        number, locale, decimal_quantization=decimal_quantization, group_separator=group_separator)
+        number, locale, decimal_quantization=decimal_quantization,  precision=precision, group_separator=group_separator)
```
akx, line 488: Spurious space :)
```suggestion
        number, locale, decimal_quantization=decimal_quantization, precision=precision, group_separator=group_separator)
```

akx (changes requested):
This seems to overlap with the previously (#494, #577) soft-deprecated `frac_prec` override... Would it be enough to set `pattern.frac_prec = (precision, precision)` for the `pattern` instead of adding the new parameter here in the bowels of `Pattern`?

Further, I think an exception should be raised when trying to use the mutually exclusive arguments?

Olunusib:
> This seems to overlap with the previously (#494, #577) soft-deprecated `frac_prec` override... Would it be enough to set `pattern.frac_prec = (precision, precision)` for the `pattern` instead of adding the new parameter here in the bowels of `Pattern`?
> 
> Further, I think an exception should be raised when trying to use the mutually exclusive arguments?

@akx,

- No, that won't be enough. If `precision` is set outside the `NumberPattern` class, it will be overwritten by [this](https://github.com/python-babel/babel/blob/1747d2230a0294a07c99563cf50fb95a37b6ac0f/babel/numbers.py#L1210C8-L1211C90) since decimal_quantization is False.
- Indeed, raising an exception is ideal though the arguments, in this case, aren't mutually exclusive.

babel/numbers.py
Olunusib: Good catch

gdiscry:
I was looking in a similar issue. In my case, I also wanted to use the implicit precision coming with a `str` or `Decimal` value, even if the least significant digits are 0.

I came up with the following function to compute the value of the `force_frac` override for `NumberPattern.apply`:

```python
def _get_frac_prec(
    pattern: babel.numbers.NumberPattern,
    value: float | decimal.Decimal | str,
    currency: str | None = None,
    currency_digits: bool = True,
    decimal_quantization: bool = True,
    decimal_normalization: bool = True,
    precision: int | None = None,
) -> tuple[int, int]:
    if precision is not None:
        min_prec = max_prec = precision
    elif currency and currency_digits:
        min_prec = max_prec = babel.numbers.get_currency_precision(currency)
    else:
        min_prec, max_prec = pattern.frac_prec

    if decimal_quantization and (
        pattern.exp_prec is None
        or min_prec != 0
        or max_prec != 0
    ):
        return min_prec, max_prec

    # Duplicates the beginning of NumberPattern.apply
    if not isinstance(value, decimal.Decimal):
        value = decimal.Decimal(str(value))
    value = value.scaleb(pattern.scale)

    num_prec = get_decimal_precision(value, normalized=decimal_normalization)
    return max(num_prec, min_prec), max(num_prec, max_prec)
```

Which requires to add a new parameter to `get_decimal_precision` to disable normalization:

```python
def get_decimal_precision(number: decimal.Decimal, normalized: bool = True) -> int:
    assert isinstance(number, decimal.Decimal)
    if normalized:
        number = number.normalize()
    exponent = number.as_tuple().exponent
    # Note: DecimalTuple.exponent can be 'n' (qNaN), 'N' (sNaN), or 'F' (Infinity)
    if not isinstance(exponent, int) or exponent >= 0:
        return 0
    return abs(exponent)
```

When `decimal_quantization` is enabled, `precision` sets the exact number of digits after the decimal point. When disabled, `precision` sets the minimal number of digits, but there can be more digits (with `decimal_normalization` removing or keeping the non-significant 0).

FelixSchwarz:
@Olunusib Did you close this on purpose or just resignated? I found this PR useful and it was some functionality that would be really useful to have.

Olunusib:
@FelixSchwarz  It was inadvertently closed. I'll get it back up very soon.

clj:
Hey @Olunusib, I'm just wondering about getting this back up? Seems like some generally useful functionality and it  would help me get rid of the warnings I am getting because I'm currently using `force_frac` to achieve the same as this PR would allow me to do. :)

PR #1015, opened 2023-06-16
Support 't' specifier in keywords

### LINKED ISSUE #1014: Feature request: Support gettext-like "t" specifier in keywords
The full description of xgettext's `--keyword` option is [here](https://www.gnu.org/software/gettext/manual/html_node/xgettext-Invocation.html#Language-specific-options).

`pybabel extract` supports a similar syntax, but it does not support the trailing "*n*t" to indicate that this keyword only applies when there are *n* arguments. This can be useful in some Python projects ([example](https://github.com/frescobaldi/frescobaldi)) where the `_` function is variadic for convenience. (In this particular project, it behaves like `gettext` with 1 arg, `pgettext` with 2 args, `ngettext` with 3 args and `npgettext` with 4 args, i.e., `_(message)`, `_(context, message)`, `_(singular, plural, number)` or `_(context, singular, plural, number)`.)

jeanas:
Fixes #1014

### DIFF
--- a/babel/messages/extract.py
+++ b/babel/messages/extract.py
@@ -453,13 +453,13 @@ def extract(
                 )
                 continue
 
-            messages = tuple(msgs)
-            if len(messages) == 1:
-                messages = messages[0]
+            msgs = tuple(msgs)
+            if len(msgs) == 1:
+                msgs = msgs[0]
 
             if strip_comment_tags:
                 _strip_comment_tags(comments, comment_tags)
-            yield lineno, messages, comments, context
+            yield lineno, msgs, comments, context
 
 
 def extract_nothing(
--- a/babel/messages/extract.py
+++ b/babel/messages/extract.py
@@ -317,7 +317,7 @@ def extract_from_file(
 
 
 def _match_messages_against_spec(lineno: int, messages: list[str|None], comments: list[str],
-                                 fileobj: _FileObj, spec: tuple[int|tuple, ...]):
+                                 fileobj: _FileObj, spec: tuple[int|tuple[int, str], ...]):
     translatable = []
     context = None
 
--- a/babel/messages/extract.py
+++ b/babel/messages/extract.py
@@ -316,6 +316,47 @@ def extract_from_file(
                             options, strip_comment_tags))
 
 
+def _match_messages_against_spec(lineno, messages, comments, fileobj, spec):
+    translatable = []
+    context = None
+
+    # last_index is 1 based like the keyword spec
+    last_index = len(messages)
+    for index in spec:
+        if isinstance(index, tuple): # (n, 'c')
+            context = messages[index[0] - 1]
+            continue
+        if last_index < index:
+            # Not enough arguments
+            return
+        message = messages[index - 1]
+        if message is None:
+            return
+        translatable.append(message)
+
+    # An empty string msgid isn't valid, emit a warning
+
+    # keyword spec indexes are 1 based, therefore '-1'
+    if isinstance(spec[0], tuple):
+        # context-aware *gettext method
+        first_msg_index = spec[1] - 1
+    else:
+        first_msg_index = spec[0] - 1
+    if not messages[first_msg_index]:
+        filename = (getattr(fileobj, "name", None) or "(unknown)")
+        sys.stderr.write(
+            f"{filename}:{lineno}: warning: Empty msgid.  It is reserved by GNU gettext: gettext(\"\") "
+            f"returns the header entry with meta information, not the empty string.\n"
+        )
+        return
+
+    translatable = tuple(translatable)
+    if len(translatable) == 1:
+        translatable = translatable[0]
+
+    yield lineno, translatable, comments, context
+
+
 def extract(
     method: _ExtractionMethod,
     fileobj: _FileObj,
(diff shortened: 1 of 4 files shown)


akx:
> There seems to be a mixture of pytest and unittest, setuptools integration and not.

For new tests, prefer plain `pytest` style. Don't worry about the setuptools integration, it's... a bit of a mess.

> Or should I just write a test that exercises the inner function directly? Help is welcome.

You can write a test that uses exercises the extraction function with a bespoke `keywords` directly, that's fine!

babel/messages/frontend.py
```diff
@@ -1111,30 +1111,46 @@ def parse_mapping(fileobj, filename=None):
...
+            if number is None:
+                # For best backwards compatibility, collapse {None: x} into x.
+                keywords[funcname] = inds
+            else:
+                keywords[funcname] = {number: inds}
+        else:
+            if isinstance(keywords[funcname], tuple) or keywords[funcname] is None:
+                keywords[funcname] = {None: keywords[funcname]}
+            else:
+                assert isinstance(keywords[funcname], dict)
```
akx, line 1152: Is this just a type-checking assertion?

babel/messages/extract.py
```diff
@@ -400,56 +401,65 @@ def extract(
                    options=options or {})
 
     for lineno, funcname, messages, comments in results:
-        spec = keywords[funcname] or (1,) if funcname else (1,)
+        specs = keywords[funcname] or (1,) if funcname else (1,)
```
akx, line 404: Might as well simplify this to
```suggestion
        specs = keywords.get(funcname) or (1,)
```
(assuming `keywords[x]` is never supposed to be falsy)?

babel/messages/extract.py
```diff
@@ -400,56 +401,65 @@ def extract(
...
-        for index in spec:
-            if isinstance(index, tuple):
-                context = messages[index[0] - 1]
+
+        # None matches all arities.
+        for arity in (None, len(messages)):
+            try:
+                spec = specs[arity]
+            except KeyError:
+                continue
```
akx, line 419: Instead of this big huge for loop over the 2 arities, could the loop body here be broken out to a separate helper function?

akx (changes requested):
A couple of comments – sorry for the delay...

babel/messages/extract.py
akx: Sure thing. :)

akx:
@jeanas Sorry, I was on holiday and all that 😊

babel/messages/extract.py
```diff
@@ -315,6 +316,47 @@ def extract_from_file(
...
+            continue
+        if last_index < index:
+            # Not enough arguments
+            return
+        message = messages[index - 1]
+        if message is None:
+            return
+        translatable.append(message)
+
+    # An empty string msgid isn't valid, emit a warning
```
akx, line 337: I think this comment belongs down next to `if not messages...`, not here?

babel/messages/extract.py
```diff
@@ -315,6 +316,47 @@ def extract_from_file(
                             options, strip_comment_tags))
 
 
+def _match_messages_against_spec(lineno, messages, comments, fileobj, spec):
```
akx, line 319: Would it be possible to add some type annotations here?

babel/messages/extract.py
```diff
@@ -315,6 +316,47 @@ def extract_from_file(
...
+            f"{filename}:{lineno}: warning: Empty msgid.  It is reserved by GNU gettext: gettext(\"\") "
+            f"returns the header entry with meta information, not the empty string.\n"
+        )
+        return
+
+    translatable = tuple(translatable)
+    if len(translatable) == 1:
+        translatable = translatable[0]
+
+    yield lineno, translatable, comments, context
```
akx, line 357: This should probably be a plain `return` – it's a bit idiosyncratic to only ever yield a single value. That would probably also be more performant as the runtime didn't need to set up a generator state and all.

babel/messages/extract.py
```diff
@@ -400,56 +442,28 @@ def extract(
...
+
+        # None matches all arities.
+        for arity in (None, len(messages)):
+            try:
+                spec = specs[arity]
+            except KeyError:
+                continue
+            if spec is None:
+                spec = (1,)
+            yield from _match_messages_against_spec(lineno, messages, comments, fileobj, spec)
```
akx, line 466: (... and if `_match_messages_against_spec` is changed to only ever return None or a message 5ple,

```suggestion
            message = _match_messages_against_spec(lineno, messages, comments, fileobj, spec)
            if message is not None:
                yield message
```
here of course.

babel/messages/frontend.py
```diff
@@ -1111,34 +1111,63 @@ def parse_mapping(fileobj, filename=None):
 
     return method_map, options_map
 
+def _parse_spec(s):
```
akx, line 1114:
```suggestion
def _parse_spec(s: str) -> tuple[int | None, tuple[int, ...]]:
    """Parse a keyword spec string.
    
    See the documentation for `parse_keywords`
    for documentation on the format.
    """
```
or similar might be nice here?

babel/messages/frontend.py
```diff
@@ -1111,34 +1111,63 @@ def parse_mapping(fileobj, filename=None):
...
-                indices = tuple(inds)
-            keywords[funcname] = indices
+            funcname = string
+            number = None
+            spec = None
+        keywords.setdefault(funcname, {})[number] = spec
+
+    # For best backwards compatibility, collapse {None: x} into x.
+    for k, v in keywords.items():
+        if set(v.keys()) == {None}:
```
akx, line 1168:
```suggestion
        if set(v) == {None}:
```

akx (changes requested):
A couple requests remaining. Thanks for your patience 😅

akx (approved):
Thank you, and sorry for the long review delays!

Issue #1005, opened 2023-05-25
Test failures with Python 3.12.0b1


mgorny:
## Overview Description

The test suite fails when run with Python 3.12.0b1:

```
FAILED tests/messages/test_extract.py::ExtractPythonTestCase::test_utf8_message_with_utf8_bom -   File "<string>", line 1
FAILED tests/messages/test_extract.py::ExtractPythonTestCase::test_utf8_message_with_utf8_bom_and_magic_comment -   File "<string>", line 1
FAILED tests/messages/test_extract.py::ExtractPythonTestCase::test_utf8_raw_strings_match_unicode_strings -   File "<string>", line 1
FAILED tests/messages/test_extract.py::ExtractTestCase::test_f_strings - AssertionError: assert 3 == 4
FAILED tests/messages/test_extract.py::ExtractTestCase::test_f_strings_non_utf8 - assert 0 == 1
```

Furthermore, `tox -e py312` fails by default because of missing `distutils` module (installing `setuptools` can workaround that but distutils use should be removed altogether).

## Steps to Reproduce

1. `tox -e py312`

## Actual Results

```pytb
________________________________________ ExtractPythonTestCase.test_utf8_message_with_utf8_bom ________________________________________

self = <tests.messages.test_extract.ExtractPythonTestCase testMethod=test_utf8_message_with_utf8_bom>

        def test_utf8_message_with_utf8_bom(self):
            buf = BytesIO(codecs.BOM_UTF8 + """
    # NOTE: hello
    msg = _('Bonjour à tous')
    """.encode('utf-8'))
>           messages = list(extract.extract_python(buf, ('_',), ['NOTE:'], {}))

tests/messages/test_extract.py:367: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _
babel/messages/extract.py:500: in extract_python
    for tok, value, (lineno, _), _, _ in tokens:
/usr/lib/python3.12/tokenize.py:451: in _tokenize
    for token in _generate_tokens_from_c_tokenizer(source, extra_tokens=True):
[...]

PR #1027, opened 2023-09-21
Add f-string parsing for Python 3.12 (PEP 701)

encukou:
Since Python 3.12, f-strings are tokenized and parsed like the rest of Python's grammar, using the new tokens `FSTRING_START`, `FSTRING_MIDDLE` and `FSTRING_END`, as described in [pep 701 #new-tokens](https://peps.python.org/pep-0701/#new-tokens).

Make the babel message extractor concatenate these three if they're adjacent to each other. If they're not, that means there are dynamic substitutions, so the f-string is ignored.

### DIFF
--- a/babel/messages/extract.py
+++ b/babel/messages/extract.py
@@ -21,6 +21,7 @@ import ast
 import io
 import os
 import sys
+import tokenize
 from collections.abc import (
     Callable,
     Collection,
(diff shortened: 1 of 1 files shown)


encukou:
Not sure if y'all want to add 3.12 to the CI yet?
For that to pass [setuptools will need to be added to tox.ini](https://github.com/python-babel/babel/issues/1005#issuecomment-1728105742) as it's needed in the data setup.

mgorny:
Thanks. I can confirm that with this patch, the tests pass both on 3.11 and 3.12.

akx (approved):
LGTM! I tested this locally on 3.12 and seems to work great. I took the liberty of fixing the lint fail too.

akx:
> Not sure if y'all want to add 3.12 to the CI yet? For that to pass [setuptools will need to be added to tox.ini](https://github.com/python-babel/babel/issues/1005#issuecomment-1728105742) as it's needed in the data setup.

Working on that in #1028 :)

BrentHuang:
how to use?  In babel 2.14.0, the code can't be extracted to .pot file:

```
foo= 'Hardy'
print(_(f"Hello {foo}"))
```

BrentHuang:
> how to use? In babel 2.14.0, the code can't be extracted to .pot file:
> 
> ```
> foo= 'Hardy'
> print(_(f"Hello {foo}"))
> ```

```
foo= 'Hardy'
print(_(f"Hello {foo}"))
```

in the code above, extract ok, but not translated by gettext.

PR #1033, opened 2023-10-07
Depend on setuptools since Python 3.12

### LINKED ISSUE #1031: distutils no longer part of Python 3.12 - error due to missing dependency
## Overview Description

babel has some code that tries to use "setuptools" or fall back to "distutils".
https://github.com/python-babel/babel/blob/76677eacd10cc35bea0a11ec1d2f0566cadd4066/babel/messages/frontend.py#L54

"distutils" used to be shipped with Python. Additionally, "setuptools" used to be ensured by virtualenvs. But both of these are gone with Python 3.12 so now babel ends up using a dependency that it never declared.
https://github.com/python-babel/babel/blob/76677eacd10cc35bea0a11ec1d2f0566cadd4066/setup.py#L64

From my assessment, it seems it would be correct to add a dependency on setuptools directly.
Either only for python_version>=3.12 or perhaps always add this dependency and drop the distutils fallback code.

**Workaround:** `pip install setuptools`

## Steps to Reproduce

Install babel not in virtualenv and run it under Python 3.12

## Actual Results

https://github.com/mkdocs/mkdocs/actions/runs/6382007433/job/17355070257

```
        File "/tmp/pip-build-env-z7xd48v4/normal/lib/python3.12/site-packages/babel/messages/frontend.py", line 45, in <module>
          from setuptools import Command as _Command
      ModuleNotFoundError: No module named 'setuptools'
      
      During handling of the above exception, another exception occurred:
      
        File "/home/runner/.local/share/hatch/env/virtual/mkdocs/FPjLbWME/test.py3.12-default/lib/python3.12/site-packages/pip/_vendor/pyproject_hooks/_in_process/_in_process.py", line 353, in <module>
          main()
        File "/home/runner/.local/share/hatch/env/virtual/mkdocs/FPjLbWME/test.py3.12-default/lib/python3.12/site-packages/pip/_vendor/pyproject_hooks/_in_process/_in_process.py", line 335, in main
          json_out['return_val'] = hook(**hook_input['kwargs'])
                                   ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
        File "/home/runner/.local/share/hatch/env/virtual/mkdocs/FPjLbWME/test.py3.12-default/lib/python3.12/site-packages/pip/_vendor/pyproject_hooks/_in_process/_in_process.py", line 176, in prepare_metadata_for_build_editable
          whl_basename = build_hook(metadata_directory, config_settings)
                         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
        File "/tmp/pip-build-env-z7xd48v4/overlay/lib/python3.12/site-packages/hatchling/build.py", line 78, in build_editable
          return os.path.basename(next(builder.build(wheel_directory, ['editable'])))
                                  ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
        File "/tmp/pip-build-env-z7xd48v4/overlay/lib/python3.12/site-packages/hatchling/builders/plugin/interface.py", line 150, in build
          build_hook.initialize(version, build_data)
        File "/home/runner/work/mkdocs/mkdocs/hatch_build.py", line 8, in initialize
          from babel.messages.frontend import compile_catalog
        File "/tmp/pip-build-env-z7xd48v4/normal/lib/python3.12/site-packages/babel/messages/frontend.py", line 54, in <module>
          from distutils import log as distutils_log
      ModuleNotFoundError: No module named 'distutils'
      [end of output]
```
## Expected Results

No error

## Background

https://docs.python.org/3.12/whatsnew/3.12.html#distutils

> Remove the distutils package. It was deprecated in Python 3.10 by PEP 632 “Deprecate distutils module”. For projects still using distutils and cannot be updated to something else, the setuptools project can be installed: it still provides distutils. 

> easy_install, pkg_resources, setuptools and distutils are no longer provided by default in environments created with venv or bootstrapped with ensurepip, since they are part of the setuptools package. For projects relying on these at runtime, the setuptools project should be declared as a dependency and installed separately (typically, using pip).

oprypin:
The dependency was already added specifically for CI, but that is not correct, because it's the package itself that actually depends on setuptools

* Fixes #1031

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -67,6 +67,9 @@ setup(
         # higher.
         # Python 3.9 and later include zoneinfo which replaces pytz
         'pytz>=2015.7; python_version<"3.9"',
+        # https://github.com/python/cpython/issues/95299
+        # https://github.com/python-babel/babel/issues/1031
+        'setuptools; python_version>="3.12"',
     ],
     extras_require={
         'dev': [
--- a/tox.ini
+++ b/tox.ini
@@ -8,10 +8,6 @@ envlist =
 extras =
     dev
 deps =
-    # including setuptools here for CI;
-    #   see https://github.com/python/cpython/issues/95299
-    #   see https://github.com/python-babel/babel/issues/1005#issuecomment-1728105742
-    setuptools;python_version>="3.12"
     backports.zoneinfo;python_version<"3.9"
     tzdata;sys_platform == 'win32'
     pytz: pytz


vstinner (approved):
LGTM.

You might explain in a comment that the intent of installing setuptools is to get distutils ;-)

akx (approved):
Makes sense. I only wish I knew how many people still depend on the `distutils`/`setuptools` stuff in the first place.

PR #1035, opened 2023-10-12
Fix for #990 (updated)

### LINKED ISSUE #990: Crash with double-slashes in /etc/localtime on Python 3.9
## Overview Description

Importing `babel.localtime` crashes on Python 3.9 if `/etc/localtime` contains double-slashes.

## Steps to Reproduce

```none
ls -lh /etc/localtime 
lrwxrwxrwx 1 root root 24 Mar  2 03:25 /etc/localtime -> /usr/share/zoneinfo//UTC
```

Note the double-slash, `//UTC`.

```none
# python
Python 3.9.16 (main, Mar 12 2023, 19:18:41) 
[GCC 9.4.0] on linux
>>> import babel.localtime
  File "/usr/local/lib/python3.9/site-packages/babel/localtime/__init__.py", line 41, in <module>
    LOCALTZ = get_localzone()
  File "/usr/local/lib/python3.9/site-packages/babel/localtime/__init__.py", line 37, in get_localzone
    return _get_localzone()
  File "/usr/local/lib/python3.9/site-packages/babel/localtime/_unix.py", line 49, in _get_localzone
    tzinfo = _get_tzinfo(zone_name)
  File "/usr/local/lib/python3.9/site-packages/babel/localtime/_helpers.py", line 21, in _get_tzinfo
  File "/usr/local/lib/python3.9/zoneinfo/_tzpath.py", line 67, in find_tzfile
    _validate_tzfile_path(key)
  File "/usr/local/lib/python3.9/zoneinfo/_tzpath.py", line 81, in _validate_tzfile_path
    raise ValueError(
ValueError: ZoneInfo keys may not be absolute paths, got: /UTC
>>> import babel
>>> babel.__version__
'2.12.1'
```

## Actual Results

Crash.

## Expected Results

No crash.

## Reproducibility

Crashes for me on Python 3.9 with `babel` `2.12.1`. Works on Python 3.8.

The problem is that `babel.localtime._unix._get_localzone` [has naive path handling](https://github.com/python-babel/babel/blob/134c792df3b6aa84de3e74e81a875493c95dcc2e/babel/localtime/_unix.py#L48).

## Additional Information

I'm not sure whether `babel` is to blame here, but since this is working on Python 3.8 it would be nice to make `babel` more robust here. As far as I can tell, [double slashes are valid in POSIX and should be treated as single slashes](https://unix.stackexchange.com/a/1919/30497).

mdklatt:
**This is an update of https://github.com/python-babel/babel/pull/1006 that incorporates upstream fixes to previously-broken tests. It should pass all CI checks now.**

This PR:

- Adds some basic tests for the `localtime` module (developed on macOS, test stub added for Windows)
- Fixes #990

### DIFF
--- a/babel/localtime/_unix.py
+++ b/babel/localtime/_unix.py
@@ -39,7 +39,10 @@ def _get_localzone(_root: str = '/') -> datetime.tzinfo:
     # zone on operating systems like OS X.  On OS X especially this is the
     # only one that actually works.
     try:
-        link_dst = os.readlink('/etc/localtime')
+        # Make sure the path is normalized so that it will be parsed correctly
+        # below, e.g. 'Etc//UTC' is a valid path but not a valid zoneinfo key.
+        # <https://github.com/python-babel/babel/issues/990>
+        link_dst = os.path.realpath('/etc/localtime')
     except OSError:
         pass
     else:
--- a/setup.py
+++ b/setup.py
@@ -72,6 +72,7 @@ setup(
         'dev': [
             'pytest>=6.0',
             'pytest-cov',
+            'pyfakefs~=5.0',
             'freezegun~=1.0',
         ],
     },
--- a/tests/test_localtime.py
+++ b/tests/test_localtime.py
@@ -5,13 +5,13 @@ executed.
 
 """
 import datetime
-import pytest
-import sys
 import os.path
+import sys
 from pathlib import Path
 
-from babel.localtime import get_localzone, LOCALTZ
+import pytest
 
+from babel.localtime import LOCALTZ, get_localzone
 
 _timezones = {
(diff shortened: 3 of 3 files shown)


akx (commented):
Would it be possible to formulate the tests without adding the new `pyfakefs` dependency?

mdklatt:
> Would it be possible to formulate the tests without adding the new `pyfakefs` dependency?

The Unix tests are the only ones that use `pyfakefs`, so I can probably replace that with `os.chroot`.

akx:
> The Unix tests are the only ones that use `pyfakefs`, so I can probably replace that with `os.chroot`.

That sounds a bit heavy-handed too... What if you just `mock.patch` some of the relevant calls?

Issue #446, opened 2016-09-20
Adjust Persian locale for numbers ('fa_IR')


5j9:
Currently:

```python
>>> from babel import numbers
>>> numbers.format_number(1234567890.123456790, 'fa_IR')
'1,234,567,890.123'
```

What I was expecting:

```python
>>> from babel import numbers
>>> numbers.format_number(1234567890.123456790, 'fa_IR')
'۱٬۲۳۴٬۵۶۷٬۸۹۰٫۱۲۳'
```

Also the current currency symbol is "﷼" and the "%" sign is "٪".

I'm using Python 3.5.2 64 bit on a Windows 10 machine.

Thanks!

jtwang:
There are a bunch of problems in this ticket.

1. Converting Arabic numerals to Persian numerals
2. Incorrect currency symbol
3. Incorrect number symbols/separators.

**Regarding 3. Incorrect number symbols/separators**

Ahh, the [code that parses number symbols|https://github.com/python-babel/babel/blob/master/scripts/import_cldr.py#L694] doesn't account for different number systems. 

{code}
def parse_number_symbols(data, tree):
    number_symbols = data.setdefault('number_symbols', {})
    for elem in tree.findall('.//numbers/symbols/*'):
        if _should_skip_elem(elem):
            continue
        number_symbols[elem.tag] = text_type(elem.text)
{code}

In the CLDR, fa_IR.xml doesn't define number symbols, so it inherits from fa.xml, [which specifies the latn number system *last*|http://www.unicode.org/repos/cldr/tags/release-28/common/main/fa.xml]. Which means these elements will be processed last, which means they're the ones that 'stick'.

`import_cldr.py` should respect the `defaultNumberingSystem` tag.

jtwang:
**Regarding 1. Converting Arabic numerals to Persian numerals**

It looks like [data for this conversation is supported in the CLDR](http://www.unicode.org/repos/cldr/trunk/common/transforms/Persian-Latin-BGN.xml), but it's not currently read into Babel.

IMO this is a feature request.

jtwang:
**Regarding 2. Incorrect currency symbol**

```
>>> print format_currency(123.45, 'IRR', locale='fa_IR')
‎ریال
```

Which is different from @5j9 's requested symbol '﷼'

[CLDR 28 defines the currency symbol for the Iranian Rial](http://www.unicode.org/repos/cldr/tags/release-28/common/main/fa.xml) as 

```
<currency type="IRR">
[...]

PR #1036, opened 2023-10-23
Add support of local numbering systems for number symbols

kajte:
Changes:
- Load number symbols for multiple number systems from cldr data
- Add numbering_systems and default_numbering_system properties for Locale
- Use default numbering system of the locale for formatting number symbols

Example:
Before:
```
>>> numbers.format_decimal(123456789.01234, locale="fa_IR")
'123,456,789.012'
```
Now:
```
numbers.format_decimal(123456789.01234, locale="fa_IR")
'123٬456٬789٫012'
```

Fixes partially issue:https://github.com/python-babel/babel/issues/446
I took some inspiration from the closed PR: https://github.com/python-babel/babel/pull/470

Couple questions:
- How much do we have to care about backward compatibility? Is it okay that this changes the numbering system used for formatting symbols from `latn` to the default numbering system for locales which default system isn't `latn`? Eg. for before the change `fa_IR` used `latn` but now it starts using `arabext`.
- Do we necessarily need a way to force usage of other numbering system than a default one for a locale? I'm not sure how common need it is to use other than the default numbering system of a locale but I feel it's not that useful.

I would eventually like to fix other `# TODO: Support other number systems` TODOs from the code and take default numbering system to use in percentage, currency etc. formatting and add a function for formatting digits to the local numbering system but I decided to create a pull request at this point to check the direction I'm taking is ok for maintainers. I feel even features added in this PR are an enhancement as such.

### DIFF
--- a/babel/core.py
+++ b/babel/core.py
@@ -149,7 +149,7 @@ class Locale:
     `Locale` objects provide access to a collection of locale data, such as
     territory and language names, number and date format patterns, and more:
 
-    >>> locale.number_symbols['decimal']
+    >>> locale.number_symbols['latn']['decimal']
     u'.'
 
     If a locale is requested for which no locale data is available, an
@@ -625,15 +625,41 @@ class Locale:
 
     @property
     def number_symbols(self) -> localedata.LocaleDataDict:
-        """Symbols used in number formatting.
+        """Symbols used in number formatting by number system.
 
         .. note:: The format of the value returned may change between
                   Babel versions.
 
-        >>> Locale('fr', 'FR').number_symbols['decimal']
+        >>> Locale('fr', 'FR').number_symbols["latn"]['decimal']
         u','
+        >>> Locale('fa', 'IR').number_symbols["arabext"]['decimal']
+        u'٫'
+        >>> Locale('fa', 'IR').number_symbols["latn"]['decimal']
+        u'.'
         """
         return self._data['number_symbols']
+
+    @property
+    def other_numbering_systems(self) -> localedata.LocaleDataDict:
+        """
+        Mapping of other numbering systems available for the locale.
+        See: https://www.unicode.org/reports/tr35/tr35-numbers.html#otherNumberingSystems
+
+        >>> Locale('el', 'GR').other_numbering_systems['traditional']
+        u'grek'
+
+        .. note:: The format of the value returned may change between
+                  Babel versions.
+        """
+        return self._data['numbering_systems']
+
+    @property
+    def default_numbering_system(self) -> str:
+        """The default numbering system used by the locale.
+        >>> Locale('el', 'GR').default_numbering_system
+        u'latn'
+        """
+        return self._data['default_numbering_system']
 
     @property
     def decimal_formats(self) -> localedata.LocaleDataDict:
--- a/babel/numbers.py
+++ b/babel/numbers.py
@@ -317,70 +317,161 @@ def get_territory_currencies(
     return result
 
 
-def get_decimal_symbol(locale: Locale | str | None = LC_NUMERIC) -> str:
+def _get_numbering_system(locale: Locale, numbering_system: Literal["default"] | str = "latn") -> str:
+    if numbering_system == "default":
+        return locale.default_numbering_system
+    else:
+        return numbering_system
+
+
+def _get_number_symbols(
+    locale: Locale | str,
+    *,
+    numbering_system: Literal["default"] | str = "latn"
+) -> LocaleDataDict:
+    parsed_locale = Locale.parse(locale)
+    numbering_system = _get_numbering_system(parsed_locale, numbering_system)
+    try:
+        return parsed_locale.number_symbols[numbering_system]
+    except KeyError as error:
+        raise UnsupportedNumberingSystemError(f"Unknown numbering system {numbering_system} for Locale {parsed_locale}.") from error
+
+
+class UnsupportedNumberingSystemError(Exception):
+    """Exception thrown when an unsupported numbering system is requested for the given Locale."""
+    pass
+
+
+def get_decimal_symbol(
+    locale: Locale | str | None = LC_NUMERIC,
+    *,
+    numbering_system: Literal["default"] | str = "latn",
+) -> str:
     """Return the symbol used by the locale to separate decimal fractions.
 
     >>> get_decimal_symbol('en_US')
     u'.'
+    >>> get_decimal_symbol('ar_EG', numbering_system='default')
+    u'٫'
+    >>> get_decimal_symbol('ar_EG', numbering_system='latn')
+    u'.'
 
     :param locale: the `Locale` object or locale identifier
+    :param numbering_system: The numbering system used for fetching the symbol. Defaults to "latn".
+                             The special value "default" will use the default numbering system of the locale.
+    :raise `UnsupportedNumberingSystemError`: If the numbering system is not supported by the locale.
     """
-    return Locale.parse(locale).number_symbols.get('decimal', '.')
+    return _get_number_symbols(locale, numbering_system=numbering_system).get('decimal', '.')
 
 
-def get_plus_sign_symbol(locale: Locale | str | None = LC_NUMERIC) -> str:
+def get_plus_sign_symbol(
+    locale: Locale | str | None = LC_NUMERIC,
+    *,
+    numbering_system: Literal["default"] | str = "latn",
+) -> str:
     """Return the plus sign symbol used by the current locale.
 
     >>> get_plus_sign_symbol('en_US')
     u'+'
+    >>> get_plus_sign_symbol('ar_EG', numbering_system='default')
+    u'\u061c+'
+    >>> get_plus_sign_symbol('ar_EG', numbering_system='latn')
+    u'\u200e+'
 
     :param locale: the `Locale` object or locale identifier
+    :param numbering_system: The numbering system used for fetching the symbol. Defaults to "latn".
+                             The special value "default" will use the default numbering system of the locale.
+    :raise `UnsupportedNumberingSystemError`: if the numbering system is not supported by the locale.
     """
-    return Locale.parse(locale).number_symbols.get('plusSign', '+')
+    return _get_number_symbols(locale, numbering_system=numbering_system).get('plusSign', '+')
 
 
-def get_minus_sign_symbol(locale: Locale | str | None = LC_NUMERIC) -> str:
+def get_minus_sign_symbol(
+    locale: Locale | str | None = LC_NUMERIC,
+    *,
+    numbering_system: Literal["default"] | str = "latn",
+) -> str:
     """Return the plus sign symbol used by the current locale.
 
     >>> get_minus_sign_symbol('en_US')
     u'-'
+    >>> get_minus_sign_symbol('ar_EG', numbering_system='default')
+    u'\u061c-'
+    >>> get_minus_sign_symbol('ar_EG', numbering_system='latn')
+    u'\u200e-'
 
     :param locale: the `Locale` object or locale identifier
+    :param numbering_system: The numbering system used for fetching the symbol. Defaults to "latn".
+                             The special value "default" will use the default numbering system of the locale.
+    :raise `UnsupportedNumberingSystemError`: if the numbering system is not supported by the locale.
     """
-    return Locale.parse(locale).number_symbols.get('minusSign', '-')
+    return _get_number_symbols(locale, numbering_system=numbering_system).get('minusSign', '-')
 
 
-def get_exponential_symbol(locale: Locale | str | None = LC_NUMERIC) -> str:
+def get_exponential_symbol(
+    locale: Locale | str | None = LC_NUMERIC,
+    *,
+    numbering_system: Literal["default"] | str = "latn",
+) -> str:
     """Return the symbol used by the locale to separate mantissa and exponent.
 
     >>> get_exponential_symbol('en_US')
     u'E'
+    >>> get_exponential_symbol('ar_EG', numbering_system='default')
+    u'اس'
+    >>> get_exponential_symbol('ar_EG', numbering_system='latn')
+    u'E'
 
     :param locale: the `Locale` object or locale identifier
+    :param numbering_system: The numbering system used for fetching the symbol. Defaults to "latn".
+                             The special value "default" will use the default numbering system of the locale.
+    :raise `UnsupportedNumberingSystemError`: if the numbering system is not supported by the locale.
     """
-    return Locale.parse(locale).number_symbols.get('exponential', 'E')
+    return _get_number_symbols(locale, numbering_system=numbering_system).get('exponential', 'E')
 
 
-def get_group_symbol(locale: Locale | str | None = LC_NUMERIC) -> str:
+def get_group_symbol(
+    locale: Locale | str | None = LC_NUMERIC,
+    *,
+    numbering_system: Literal["default"] | str = "latn",
+) -> str:
     """Return the symbol used by the locale to separate groups of thousands.
 
     >>> get_group_symbol('en_US')
     u','
+    >>> get_group_symbol('ar_EG', numbering_system='default')
+    u'٬'
+    >>> get_group_symbol('ar_EG', numbering_system='latn')
+    u','
 
     :param locale: the `Locale` object or locale identifier
+    :param numbering_system: The numbering system used for fetching the symbol. Defaults to "latn".
+                             The special value "default" will use the default numbering system of the locale.
+    :raise `UnsupportedNumberingSystemError`: if the numbering system is not supported by the locale.
     """
-    return Locale.parse(locale).number_symbols.get('group', ',')
+    return _get_number_symbols(locale, numbering_system=numbering_system).get('group', ',')
 
 
-def get_infinity_symbol(locale: Locale | str | None = LC_NUMERIC) -> str:
+def get_infinity_symbol(
+    locale: Locale | str | None = LC_NUMERIC,
+    *,
+    numbering_system: Literal["default"] | str = "latn",
+) -> str:
     """Return the symbol used by the locale to represent infinity.
 
     >>> get_infinity_symbol('en_US')
     u'∞'
+    >>> get_infinity_symbol('ar_EG', numbering_system='default')
+    u'∞'
+    >>> get_infinity_symbol('ar_EG', numbering_system='latn')
+    u'∞'
 
     :param locale: the `Locale` object or locale identifier
+    :param numbering_system: The numbering system used for fetching the symbol. Defaults to "latn".
+                             The special value "default" will use the default numbering system of the locale.
+    :raise `UnsupportedNumberingSystemError`: if the numbering system is not supported by the locale.
     """
-    return Locale.parse(locale).number_symbols.get('infinity', '∞')
+    return _get_number_symbols(locale, numbering_system=numbering_system).get('infinity', '∞')
 
 
 def format_number(number: float | decimal.Decimal | str, locale: Locale | str | None = LC_NUMERIC) -> str:
@@ -430,6 +521,8 @@ def format_decimal(
     locale: Locale | str | None = LC_NUMERIC,
     decimal_quantization: bool = True,
     group_separator: bool = True,
+    *,
+    numbering_system: Literal["default"] | str = "latn",
 ) -> str:
     """Return the given decimal number formatted for a specific locale.
 
@@ -443,6 +536,10 @@ def format_decimal(
     u'1,234'
     >>> format_decimal(1.2345, locale='de')
     u'1,234'
+    >>> format_decimal(1.2345, locale='ar_EG', numbering_system='default')
+    u'1٫234'
+    >>> format_decimal(1.2345, locale='ar_EG', numbering_system='latn')
+    u'1.234'
 
     The appropriate thousands grouping and the decimal separator are used for
     each locale:
@@ -470,13 +567,16 @@ def format_decimal(
                                  the format pattern. Defaults to `True`.
     :param group_separator: Boolean to switch group separator on/off in a locale's
                             number format.
+    :param numbering_system: The numbering system used for formatting number symbols. Defaults to "latn".
+                             The special value "default" will use the default numbering system of the locale.
+    :raise `UnsupportedNumberingSystemError`: If the numbering system is not supported by the locale.
     """
     locale = Locale.parse(locale)
     if format is None:
         format = locale.decimal_formats[format]
     pattern = parse_pattern(format)
     return pattern.apply(
-        number, locale, decimal_quantization=decimal_quantization, group_separator=group_separator)
+        number, locale, decimal_quantization=decimal_quantization, group_separator=group_separator, numbering_system=numbering_system)
 
 
 def format_compact_decimal(
@@ -485,6 +585,7 @@ def format_compact_decimal(
     format_type: Literal["short", "long"] = "short",
     locale: Locale | str | None = LC_NUMERIC,
     fraction_digits: int = 0,
+    numbering_system: Literal["default"] | str = "latn",
 ) -> str:
     """Return the given decimal number formatted for a specific locale in compact form.
 
@@ -500,11 +601,16 @@ def format_compact_decimal(
     u'2 милиони'
     >>> format_compact_decimal(21000000, format_type="long", locale="mk")
     u'21 милион'
+    >>> format_compact_decimal(12345, format_type="short", locale='ar_EG', fraction_digits=2, numbering_system='default')
+    u'12٫34\xa0ألف'
 
     :param number: the number to format
     :param format_type: Compact format to use ("short" or "long")
     :param locale: the `Locale` object or locale identifier
     :param fraction_digits: Number of digits after the decimal point to use. Defaults to `0`.
+    :param numbering_system: The numbering system used for formatting number symbols. Defaults to "latn".
+                             The special value "default" will use the default numbering system of the locale.
+    :raise `UnsupportedNumberingSystemError`: If the numbering system is not supported by the locale.
     """
     locale = Locale.parse(locale)
     compact_format = locale.compact_decimal_formats[format_type]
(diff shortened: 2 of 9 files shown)


babel/core.py
```diff
@@ -625,16 +625,38 @@ def currency_symbols(self) -> localedata.LocaleDataDict:
...
         return self._data['number_symbols']
 
+    @property
+    def numbering_systems(self) -> localedata.LocaleDataDict:
+        """Mapping of numbering systems.
+        >>> Locale('el', 'GR').numbering_systems['default']
+        u'latn'
+        >>> Locale('el', 'GR').numbering_systems['traditional']
+        [u'grek']
+        """
```
akx, line 649:
```suggestion

        .. note:: The format of the value returned may change between
                  Babel versions.
        """
```

babel/numbers.py
```diff
@@ -325,7 +325,8 @@ def get_decimal_symbol(locale: Locale | str | None = LC_NUMERIC) -> str:
 
     :param locale: the `Locale` object or locale identifier
```
akx, line 326: I think we will want a `numbering_system: str | None = None` parameter for this function (and the other similar functions). 

It should probably support both explicit numbering system names as well as the aliases (`'default'`, `'traditional'`), and if a specific numbering system is explicitly requested and it is not available in the locale, we should raise an exception instead of returning some default.

As you rightly did point out,

> Is it okay that this changes the numbering system used for formatting symbols from `latn` to the default numbering system for locales which default system isn't `latn`?

is a bit of a question. Looking at the data, 
```
>>> import collections
>>> collections.Counter(x.values())
Counter({'latn': 733, 'arab': 29, 'arabext': 19, 'deva': 13, 'adlm': 13, 'beng': 8, 'olck': 3, 'cakm': 3, 'tibt': 2, 'mymr': 2})
```
this would affect about 90 locales (21 macrolocales); I'd _love_ some input from `{'sd', 'dz', 'my', 'lrc', 'bn', 'ar', 'sa', 'raj', 'ccp', 'bgc', 'ks', 'bho', 'ps', 'sat', 'as', 'ne', 'mr', 'ckb', 'mzn', 'mni', 'fa'}` speakers for whether it would be weird to use non-latin symbols in conjuction with latin numerals. (On that note, could we have a test that proves this would now happen for e.g. `ar` locales?)

scripts/import_cldr.py
```diff
@@ -739,14 +740,28 @@ def parse_calendar_datetime_skeletons(data, calendar):
...
+
+            number_symbols.setdefault(number_system, {})[symbol_element.tag] = str(symbol_element.text)
+
+
+def parse_numbering_systems(data, tree):
+    numbering_systems = data.setdefault('numbering_systems', {})
+    default_number_system_node = tree.find('.//numbers/defaultNumberingSystem')
+
+    if default_number_system_node is not None:
+        numbering_systems['default'] = str(default_number_system_node.text)
```
akx, line 760: I think it might be better to just have `default_numbering_system` in the `data` by itself, because otherwise `numbering_systems`'s type will be a silly `dict[str, str | list[str]]`.

akx (changes requested):
~Kiitos~ Thank you for the PR! 💪 

Some comments, requests, and discussion within. Re the defaults, if you happen to know speakers of the affected locales (see comment), please tag them in too :)

babel/numbers.py
kajte: > I'd love some input from {'sd', 'dz', 'my', 'lrc', 'bn', 'ar', 'sa', 'raj', 'ccp', 'bgc', 'ks', 'bho', 'ps', 'sat', 'as', 'ne', 'mr', 'ckb', 'mzn', 'mni', 'fa'} speakers for whether it would be weird to use non-latin symbols in conjuction with latin numerals.

Good question. Let's see if can get some native person to comment on that. I'll try to ask around in my circles.

I would like to later add a function like `format_digits(to_numbering_system, from_numbering_system="latn")` to convert digits between different locales. Do you think that it should be eventually integrated to other formatting numbers (like `format_decimal`)? If yes, then this question will become pointless as both numerals and symbols would be using same numbering system.

kajte: > On that note, could we have a test that proves this would now happen for e.g. ar locales?

I'm not sure if I understand what kind of test you would like to have. But yup it's now mixing non-latin symbols and latin numerals.

kajte (commented):
Thanks for the quick response and good comments!

How do you see the end goal how function like `format_decimal` should work? Should it also convert the digits like this `numbers.format_decimal(1.2, locale="ar", numbering_system='default) == '٢٫٣'`?

babel/core.py
kajte: Done.

babel/numbers.py
akx: > It will lead to changing quite many functions when adding same parameter also to all functions that calls `get_decimal_symbol` (eg. `format_decimal` calls `NumberPattern.apply` which calls `get_decimal_symbol`) and other similar functions edited here. But it's probably the right thing to do.

Yeah, I think that's how it'll need to be. Oh, now that we're on this, I think the new parameter should definitely be kwarg-only! 

An additional thought: we could add a new API `get_numeric_symbols` which would return a dict (or preferably a dataclass) that'd contain all of the symbols for a given locale/system configuration, so we wouldn't need to call the small functions over and over again.

> It could be due to `jpan` being algorithmic system. Maybe not a problem but it's good to notice even with aliases it could raise an error. https://github.com/unicode-org/cldr/blob/main/common/supplemental/numberingSystems.xml#L45

Yeah, I think it's okay to let the KeyError bubble up if there are no digits or symbols given. We might want to document that in a docstring.

> On the other note language `hi` defines default and native numbering systems but there's symbols for over 40 numbering systems: https://github.com/unicode-org/cldr/blob/main/common/main/hi.xml#L5696-L6404

Most of those are `draft` status though, so we don't support them. Phew! ;)

> I would like to later add a function like `format_digits(to_numbering_system, from_numbering_system="latn")` to convert digits between different locales.

I have the feeling there might be subtleties there even with a non-algorithmic digit system. I haven't (yet...) read the CLDR spec about digit systems closely enough to say whether a simple translation can ever be enough; again, I think we would really benefit of a subject matter expert in e.g. `ar` or other systems 🤔 

> I'm not sure if I understand what kind of test you would like to have. But yup it's now mixing non-latin symbols and latin numerals.

Yeah, basically currently just a test that would show that you (could) get the non-Latin symbol with Latin numerals when this is merged.

scripts/import_cldr.py
akx: I think it might be worth to have a separate `default_numbering_system` since I foresee that to be a pretty hot path, and saving one dict lookup might compound :)

akx:
> Should it also convert the digits like this `numbers.format_decimal(1.2, locale="ar", numbering_system='default') == '٢٫٣'`?

I mean, if that's a sensible formatting, I think so, yeah. I do wonder if we should do one version where the high-level formatting functions still use `"latn"` and emit e.g. a FutureWarning when the default for the locale (e.g. `"ar"`) would not be Latin, like

```
FutureWarning: the default numbering system for the `"ar"` locale is ...; to keep Latin formatting, pass `numbering_system='latn'`
```
? I'm just kind of worried about putting people through the trouble of their formatting suddenly changing with a "routine" Babel upgrade.

babel/numbers.py
kajte: > An additional thought: we could add a new API get_numeric_symbols which would return a dict (or preferably a dataclass) that'd contain all of the symbols for a given locale/system configuration, so we wouldn't need to call the small functions over and over again.

Hmm to which level functions would the dataclass be passed? I'm not sure how nice the api would have to pass the dataclass to eg. `format_decimal`.

kajte:
**Stats about numbering systems regarding symbols**
I did some statistics regarding availability of number symbols (I'll paste code used for calculations to the end of message):
- All 825 supported locales have number symbols data for `latn`
- All locales have number symbols data for their default numbering system
- All locales have native numbering system and symbols for it
- 41/825 locales have traditional numbering system. 0/41 traditional numbering system have symbols.
- There's 69 locales where if native isn't latn then native isnt' same as default.

Conclusion: There's no need to have traditional alias. Also I feel `native` alias isn't that useful and we could skip it if not having it makes code simpler but it's your call.
_______________________________________________________________________________________
**Next steps on this PR**

Could we preliminarily agree on sensible signature for functions so I can't start implementing it without the need to do more changes later?

One option could be to add something like `numbering_system_config: Literal["default"] | str = "latn"` to all relevant functions defaulting to latn. On later versions the `FutureWarning` can be added as you suggested. 

It would be nicer if it don't have to call `_determine_numbering_system` on every stage and for lower level functions the argument could be only `numbering_system: str = "latn"`. However, as far as I understand the library structure, most of functions in `numbers` module (and all on the example below) are public and documented so someone might be using them directly and then the ability to pass "default" is probably useful. Correct me if I'm wrong on this. 

What do you think? Do you have more suitable suggestions?
```
# new function
def _determine_numbering_system(locale: Locale, numbering_system: Literal["default"] | str = "latn") -> str:
    if numbering_system == "default":
        return locale.default_numbering_system
    else:
        return numbering_system
 
#existing but numbering_system_config argument added
def format_decimal(
    number: float | decimal.Decimal | str,
    format: str | NumberPattern | None = None,
    locale: Locale | str | None = LC_NUMERIC,
    decimal_quantization: bool = True,
    group_separator: bool = True,
    *,
    numbering_system_config: Literal["default"] | str = "latn"
) -> str:
    locale = Locale.parse(locale)
    numbering_system = _determine_numbering_system(locale, numbering_system_config)
    ...
    NumberPattern.apply(..., numbering_system_config=numbering_system)
    
    
NumberPattern():
#existing but numbering_system_config argument added
     def apply(
        self,
        value: float | decimal.Decimal | str,
        locale: Locale | str | None,
        currency: str | None = None,
        currency_digits: bool = True,
        decimal_quantization: bool = True,
        force_frac: tuple[int, int] | None = None,
        group_separator: bool = True,
        *,
        numbering_system_config: Literal["default"] | str = "latn"
    ):
         numbering_system = _determine_numbering_system(locale, numbering_system_config)
         ...
          symbol = get_decimal_symbol(locale, numbering_system_config=numbering_system)
          ....

#existing but numbering_system_config argument and modified to use number system for getting correct symbol
def get_decimal_symbol(
        locale: Locale | str | None = LC_NUMERIC,
        *,
        numbering_system_config: Literal["default"] | str = "latn"
) -> str:
    parsed_locale = Locale.parse(locale)
    numbering_system = _determine_numbering_system(locale, numbering_system_config)
    return parsed_locale.number_symbols[numbering_system].get('decimal', '.')
```

___________________________________________
**Other thoughts**
I was chatting with my friend who is native speaker of `bn`. For that language it seems eg. decimal delimiters are same for bo latn and beng numering systems so it didn't help for the question whether mixing local symbols and latn numbers looks weird. However, I learned from him that converting digits from latn to beng is just doing 1-to-1 mapping from 0123456789 to ০১২৩৪৫৬৭৮৯ (https://github.com/unicode-org/cldr/blob/main/common/supplemental/numberingSystems.xml#L18). Also, he said they use quite often also latn numbers.

I think it would be quite easy to validate digit formatting against eg. the java/kotlin number formatter from the standard library to get more confidence (or to findout  there's more nuances) that converting digits is as simple as 1-to-1 mapping for all number systems having 10 digits. 

____________________________________
Here's code that I used to get stats on the beginning of post. Basically calculating number of failures or successes depending on the case. Just for the reference.

```@pytest.mark.all_locales
def test_latn_available(locale):
    assert Locale.parse(locale).number_symbols["latn"]


@pytest.mark.all_locales
def test_default_available(locale):
    locale = Locale.parse(locale)
    assert locale.number_symbols[locale.default_numbering_system]


@pytest.mark.all_locales
def test_locale_have_native(locale):
    assert Locale.parse(locale).other_numbering_systems["native"]


@pytest.mark.all_locales
def test_native_locale_have_symbols(locale):
    locale = Locale.parse(locale)
    assert locale.number_symbols[locale.other_numbering_systems["native"]]


@pytest.mark.all_locales
def test_locale_have_traditional(locale):
    assert Locale.parse(locale).other_numbering_systems["traditional"]


@pytest.mark.all_locales
def test_traditional_locale_have_symbols(locale):
    locale = Locale.parse(locale)
    assert locale.number_symbols[locale.other_numbering_systems["traditional"]]


@pytest.mark.all_locales
def test_native_and_default_same_if_native_isnt_latin(locale):
    locale = Locale.parse(locale)
    native = locale.other_numbering_systems["native"]
    default = locale.default_numbering_system

    if native != "latn":
        assert native == default, f"{default} {native} {locale}"

```

akx:
Sorry for the delay, had a bunch of other things to attend to...

> One option could be to add something like `numbering_system_config: Literal["default"] | str = "latn"` to all relevant functions defaulting to latn. On later versions the `FutureWarning` can be added as you suggested.

Sounds good, but let's call the argument `numbering_system` instead, since it's kwarg-only so `numbering_system_config=` would be littering user code with an extra `_config` word :)

> It would be nicer if it don't have to call `_determine_numbering_system` on every stage and for lower level functions the argument could be only `numbering_system: str = "latn"`. However, as far as I understand the library structure, most of functions in `numbers` module (and all on the example below) are public and documented so someone might be using them directly and then the ability to pass "default" is probably useful. 

That is the case, yep. That's somewhat analogous to how you can pass in a locale identifier to almost anything, and we call `Locale.parse` on it.

The API surface looks good to me at first glance aside from that naming change :)

akx:
@kajte I'm planning on trying to get a new Babel out maybe this week or early next week, and I think we could at least include the numbering-system loading code from this PR, even if there'd be no surface changes that'd allow to use them just yet, if you have the time to crop this PR to only that? IOW, pretty much just change the `get_*_symbol()` code to always use `"latn"` for the time being.

kajte:
> @kajte I'm planning on trying to get a new Babel out maybe this week or early next week, and I think we could at least include the numbering-system loading code from this PR, even if there'd be no surface changes that'd allow to use them just yet, if you have the time to crop this PR to only that? IOW, pretty much just change the `get_*_symbol()` code to always use `"latn"` for the time being.

I wrote some code last week that I didn't have time to complete yet. I can check today or tomorrow whether it's close to done. If not, your suggestions sounds good.

kajte:
Okay, I got something done now.

1. Loads number symbols for available numbering systems from cldr data
2. All number formatting functions using number symbol data should now take argument `numbering_system: Literal["default"] | str = "latn"` which is used to select from which numbering system symbols are used. 
I hope I didn't miss any function or forgot to test any function where I added the numbering_system argument

How does it look? If needed I can also change it what you suggested on Wednesday so that we only want to include that part to the next release.

babel/core.py
```diff
@@ -625,16 +625,39 @@ def currency_symbols(self) -> localedata.LocaleDataDict:
...
+        >>> Locale('fa', 'IR').number_symbols["arabext"]['decimal']
+        u'٫'
+        >>> Locale('fa', 'IR').number_symbols["latn"]['decimal']
+        u'.'
         """
         return self._data['number_symbols']
 
+    @property
+    def other_numbering_systems(self) -> localedata.LocaleDataDict:
+        """Mapping of othern numbering systems.
```
akx, line 644:
```suggestion
        """Mapping of other numbering systems.
```

What does "other" mean here? Could we elucidate that some?

babel/core.py
```diff
@@ -625,16 +625,39 @@ def currency_symbols(self) -> localedata.LocaleDataDict:
...
+        u'grek'
+
+        .. note:: The format of the value returned may change between
+                  Babel versions.
+        """
+        return self._data['numbering_systems']
+
+    @property
+    def default_numbering_system(self) -> str:
+        """The default numbering systems used by the locale.
```
akx, line 655:
```suggestion
        """The default numbering system used by the locale.
```

babel/numbers.py
```diff
@@ -317,70 +317,163 @@ def _is_active(start, end):
...
-def get_decimal_symbol(locale: Locale | str | None = LC_NUMERIC) -> str:
+def _get_numbering_system(locale: Locale, numbering_system: Literal["default"] | str = "latn") -> str:
+    if numbering_system == "default":
+        return locale.default_numbering_system
+    else:
+        return numbering_system
+
+
+def _get_number_symbols(
+    locale:  Locale | str | None = LC_NUMERIC,
```
akx, line 328: I don't think this internal function should have a default here.
```suggestion
    locale: Locale | str,
```

babel/numbers.py
```diff
@@ -317,70 +317,163 @@ def _is_active(start, end):
...
+        return parsed_locale.number_symbols[numbering_system]
+    except KeyError:
+        raise UnsupportedNumberingSystemError(numbering_system, parsed_locale)
+
+
+class UnsupportedNumberingSystemError(Exception):
+    """Exception thrown when an unsupported numbering system is requested for the given Locale."""
+
+    def __init__(self, numbering_system: str, locale: Locale | str | None) -> None:
+        super().__init__(f"Unknown numbering system {numbering_system} for Locale {locale}.")
```
akx, line 344: Just format this error message at the call site since you're not putting the `numbering_system` and `locale` into fields in the exception anyway.

Alternately, put them in there :)

babel/numbers.py
```diff
@@ -317,70 +317,163 @@ def _is_active(start, end):
...
+    try:
+        return parsed_locale.number_symbols[numbering_system]
+    except KeyError:
+        raise UnsupportedNumberingSystemError(numbering_system, parsed_locale)
+
+
+class UnsupportedNumberingSystemError(Exception):
+    """Exception thrown when an unsupported numbering system is requested for the given Locale."""
+
+    def __init__(self, numbering_system: str, locale: Locale | str | None) -> None:
```
akx, line 343: This does need to be (and currently is) constructed with an actual Locale.
```suggestion
    def __init__(self, numbering_system: str, locale: Locale) -> None:
```

babel/numbers.py
```diff
@@ -317,70 +317,163 @@ def _is_active(start, end):
...
     >>> get_decimal_symbol('en_US')
     u'.'
+    >>> get_decimal_symbol('ar_EG', numbering_system='default')
+    u'٫'
+    >>> get_decimal_symbol('ar_EG', numbering_system='latn')
+    u'.'
 
     :param locale: the `Locale` object or locale identifier
+    :param numbering_system: The numbering system used for fetching the symbol. Defaults to "latn", special
+                             value "default" will use the default numbering system of the locale.
```
akx, line 363:
```suggestion
    :param numbering_system: The numbering system used for fetching the symbol. Defaults to "latn".
                             The special value "default" will use the default numbering system of the locale.
```

(Consider this repeated for all `param numbering_system`s :) )

tests/test_numbers.py
```diff
@@ -176,15 +176,26 @@ def test_compact(self):
         assert numbers.format_compact_decimal(1234, locale='it', format_type='long') == '1 mila'
         assert numbers.format_compact_decimal(1000, locale='fr', format_type='long') == 'mille'
         assert numbers.format_compact_decimal(1234, locale='fr', format_type='long') == '1 millier'
+        assert numbers.format_compact_decimal(1234, locale='fr', format_type='long') == '1 millier'
```
akx, line 179: Accidental copy-paste of the one above?

```suggestion
```

tests/test_support.py
```diff
@@ -296,16 +297,55 @@ def raise_attribute_error():
...
+class TestFormat:
+    def test_format_datetime(self, timezone_getter):
+        when = datetime.datetime(2007, 4, 1, 15, 30)
+        fmt = support.Format('en_US', tzinfo=timezone_getter('US/Eastern'))
+        assert fmt.datetime(when) == 'Apr 1, 2007, 11:30:00\u202fAM'
+
+    def test_format_time(self, timezone_getter):
+        when = datetime.datetime(2007, 4, 1, 15, 30)
+        fmt = support.Format('en_US', tzinfo=timezone_getter('US/Eastern'))
+        assert fmt.time(when) == '11:30:00\u202fAM'
```
akx, line 309: Could you revert this (and unwrap the new tests from the class?), unless there's some reason for the class I'm not quite seeing here? 😁

akx (changes requested):
This is looking pretty good!

tests/test_support.py
kajte: For me it seemed logical and also felt like following the convention in this test module.

`class LazyProxy` is tested in `class LazyProxyTestCase`
`class NullTranslations` is tested in `class NullTranslationsTestCase`
`class Translations` is tested in `class TranslationsTestCase`

So it would be logical to have tests for `class Format` also grouped into once class. Though following the convention it should have been named to `FormatTestCase`.

Should I unwrap or change the class name?

babel/numbers.py
kajte: I formatted it at the call

babel/numbers.py
kajte: Whole `__init__` function based on the comment above

babel/numbers.py
kajte: Fixed

tests/test_support.py
kajte: Though changing the name isn't probably good idea as pytest doesn't recognize `FormatTestCase` unless it inherits `unittest.TestCase` which on the other hand doesn't find `timezone_getter` fixture in `test_format_datetime` test function. 

So either unwrapping or keeping as it is now.

babel/core.py
kajte: Changed it now to:
```
Mapping of other numbering systems available for the locale.
See: https://www.unicode.org/reports/tr35/tr35-numbers.html#otherNumberingSystems

**Comment by kajte** [comment]
> @kajte Thanks! Can you rebase this on `master` and take care of the lint errors? Beyond the ones currently flagged by Ruff, there will be new ones since #1045 was just merged. I'll give the rest a review too now.

Rebased and at least pre-commit hook succeed locally now.

File: tests/test_support.py (hunk 11 above)
Comment (akx) [reply tests/test_support.py:309]: It's okay. I'm just trying to get rid of `unittest.TestCase`s where they don't make sense (where they're just there for legacy reasons).

File: babel/numbers.py [hunk 12]
```diff
@@ -317,70 +317,161 @@ def _is_active(start, end):
...
-def get_decimal_symbol(locale: Locale | str | None = LC_NUMERIC) -> str:
+def _get_numbering_system(locale: Locale, numbering_system: Literal["default"] | str = "latn") -> str:
+    if numbering_system == "default":
+        return locale.default_numbering_system
+    else:
+        return numbering_system
+
+
+def _get_number_symbols(
+    locale: Locale | str,
```
akx, line 328:
```suggestion
    locale: Locale | str | None,
```

babel/numbers.py
```diff
@@ -1111,7 +1271,12 @@ def compute_scale(self) -> Literal[0, 2, 3]:
             scale = 3
         return scale
 
-    def scientific_notation_elements(self, value: decimal.Decimal, locale: Locale | str | None) -> tuple[decimal.Decimal, int, str]:
+    def scientific_notation_elements(
+        self,
+        value: decimal.Decimal,
+        locale: Locale | str | None,
+        numbering_system: str,
+    ) -> tuple[decimal.Decimal, int, str]:
```
akx, line 1279:
```suggestion
        *,
        numbering_system: str,
    ) -> tuple[decimal.Decimal, int, str]:
```

babel/units.py
```diff
@@ -230,6 +239,9 @@ def format_compound_unit(
     >>> format_compound_unit(32.5, "ton", 15, denominator_unit="hour", locale="en")
     '32.5 tons per 15 hours'
 
+    >>> format_compound_unit(1234.5, "ton", 15, denominator_unit="hour", locale="ar_EG", numbering_system="default")
```
akx, line 242: Probably best to showcase explicitly setting the numbering system too:

```suggestion
    >>> format_compound_unit(1234.5, "ton", 15, denominator_unit="hour", locale="ar_EG", numbering_system="arab")
```

tests/test_numbers.py
```diff
@@ -564,6 +633,8 @@ def test_format_percent():
             == '2\xa0512\xa0%')
     assert (numbers.format_percent(25.1234, '#,##0\u2030', locale='en_US')
             == '25,123\u2030')
+    assert numbers.format_percent(134.5, locale='ar_EG', numbering_system="default") == '13٬450%'
+
```
akx, line 637:
```suggestion
    assert numbers.format_percent(134.5, locale='ar_EG', numbering_system="default") == '13٬450%'
```

tests/test_numbers.py
```diff
@@ -356,6 +405,15 @@ def test_format_decimal():
...
 
+    assert numbers.format_decimal(12345.5, locale='ar_EG') == '12,345.5'
+    assert numbers.format_decimal(12345.5, locale='ar_EG', numbering_system="default") == '12٬345٫5'
+    assert numbers.format_decimal(12345.5, locale='ar_EG', numbering_system="arab") == '12٬345٫5'
+
+    with pytest.raises(numbers.UnsupportedNumberingSystemError):
+        numbers.format_decimal(12345.5, locale='en_US', numbering_system="unknown")
+
+
+
```
akx, line 416:
```suggestion
```

tests/test_numbers.py
```diff
@@ -660,11 +733,15 @@ def test_format_scientific_quantization():
...
     assert numbers.parse_number('1.099', locale='de_DE') == 1099
+    assert numbers.parse_number('1٬099', locale='ar_EG', numbering_system="default") == 1099
 
     with pytest.raises(numbers.NumberFormatError) as excinfo:
         numbers.parse_number('1.099,98', locale='de')
     assert excinfo.value.args[0] == "'1.099,98' is not a valid number"
 
+    with pytest.raises(numbers.UnsupportedNumberingSystemError):
+        numbers.parse_number('1.099,98', locale='en', numbering_system="unsupported")
+
```
akx, line 744:
```suggestion
```

akx (changes requested):
(Just applying some minor formatting fixes via review.)

akx:
@kajte Thanks! 🎉

kajte:
How should I run linters locally to catch these? I installed the pre-commit hook and it passed for my commit.

kajte:
> @kajte Thanks! 🎉

Thanks for your support! I have couple busy weeks ahead but I'll try continue with formatting actual numerals after that.

Issue #1037, opened 2023-10-24
Unable to load .mo files due to incorrect conversion of Locale to a List of language names


DI84:
## Description
The method "_locales_to_names()" from the "Translations" class (babel/support.py) does not correctly convert a given Locale to a List of language names. 

Instead of calling `str()` on the parameter variable `locales` (note the -s), it is called on the module object `locale` (without -s) if the parameter variable `locales` is an instance of `Locale` (see code below) .

```python
def _locales_to_names(
    locales: Iterable[str | Locale] | str | Locale | None,
) -> list[str] | None:
    """Normalize a `locales` argument to a list of locale names.

    :param locales: the list of locales in order of preference (items in
                    this list can be either `Locale` objects or locale
                    strings)
    """
    if locales is None:
        return None
    if isinstance(locales, Locale):
        return [str(locale)]
    if isinstance(locales, str):
        return [locales]
    return [str(locale) for locale in locales]
```
The 4th lowest line should read `return [str(locales)]` instead of `return [str(locale)]` (as in `locale` the module object).

As a result, the `gettext.find()` call in the `load` method of the `Translations` class does not return a valid .mo file path and therefore the translations cannot be loaded.

inabout:
@akx: This is an important bug for us in the current version 2.13.0. Would it be possible to release a 2.13.1 as soon as a fix for this is available?

akx:
Oops, great catch. Will release a patch version soon...

akx:
Fixed in https://github.com/python-babel/babel/releases/tag/v2.13.1. Sorry for the inconvenience and thanks for spotting and fixing the issue.

PR #1038, opened 2023-10-24
Fix _locales_to_names (Fix for #1037)

DI84:
Call str() on the parameter variable instead of the module object.


akx (approved):
Great catch.

PR #1041, opened 2023-10-30
Decouple `pybabel` frontend from distutils/setuptools; remove dependency

### LINKED ISSUE #1040: Make setuptools an optional dependency?
> It's a bit rough that this forces setuptools on everyone that has babel in their dependency chain - is there a plan to remove it again eventually? Could this be made an optional feature?

_Originally posted by @sigma67 in https://github.com/python-babel/babel/issues/1031#issuecomment-1783024880_

akx:
Fixes #1040

### DIFF
--- a/babel/messages/frontend.py
+++ b/babel/messages/frontend.py
@@ -1109,19 +1109,16 @@ def parse_keywords(strings: Iterable[str] = ()):
     return keywords
 
 
-try:
+def __getattr__(name: str):
     # Re-exports for backwards compatibility;
     # `setuptools_frontend` is the canonical import location.
-    from babel.messages.setuptools_frontend import (
-        check_message_extractors,  # noqa: F401
-        compile_catalog,  # noqa: F401
-        extract_messages,  # noqa: F401
-        init_catalog,  # noqa: F401
-        update_catalog,  # noqa: F401
-    )
-except ImportError:
-    # We expect this to mean that neither setuptools or distutils are installed.
-    pass
+    if name in {'check_message_extractors', 'compile_catalog', 'extract_messages', 'init_catalog', 'update_catalog'}:
+        from babel.messages import setuptools_frontend
+
+        return getattr(setuptools_frontend, name)
+
+    raise AttributeError(f"module {__name__!r} has no attribute {name!r}")
+
 
 if __name__ == '__main__':
     main()
--- a/babel/messages/frontend.py
+++ b/babel/messages/frontend.py
@@ -40,22 +40,17 @@ from babel.util import LOCALTZ
 
 log = logging.getLogger('babel')
 
-try:
-    # See: https://setuptools.pypa.io/en/latest/deprecated/distutils-legacy.html
-    from setuptools import Command as _Command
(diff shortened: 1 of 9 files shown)


akx:
Cc @sigma67 @oprypin if you want to take a review-y look! :)

setup.py
```diff
@@ -67,9 +67,6 @@ def run(self):
         # higher.
         # Python 3.9 and later include zoneinfo which replaces pytz
         'pytz>=2015.7; python_version<"3.9"',
-        # https://github.com/python/cpython/issues/95299
-        # https://github.com/python-babel/babel/issues/1031
-        'setuptools; python_version>="3.12"',
```
sigma67, line 72: I'm not sure but shouldn't this become an optional dependency? Removing it entirely means the user needs to provision it separately. Unless that is what is intended, but then it should be in the docs somewhere

sigma67 (commented):
Thanks for the fast response. Seems it will be a breaking change for those using the setuptools integration?

setup.py
akx: People who are using `setup.py` (who would thus use the `setuptools_frontend` bits) will necessarily be depending on `setuptools`, so I don't think that's absolutely necessary :)

akx:
@oprypin Could you review this too? :)

babel/messages/frontend.py
```diff
@@ -1173,5 +1109,19 @@ def parse_keywords(strings: Iterable[str] = ()):
...
+    from babel.messages.setuptools_frontend import (
+        check_message_extractors,  # noqa: F401
+        compile_catalog,  # noqa: F401
+        extract_messages,  # noqa: F401
+        init_catalog,  # noqa: F401
+        update_catalog,  # noqa: F401
+    )
+except ImportError:
+    # We expect this to mean that neither setuptools or distutils are installed.
+    pass
```
oprypin, line 1124: There's a great way (new in Python 3.7) to provide a deprecated member only on the fly!

If you want, you can even add a deprecation warning in the body. I skipped doing it for now because why not have a transition period without any deprecation message.

```suggestion
def __getattr__(name: str):
    # Re-exports for backwards compatibility;
    # `setuptools_frontend` is the canonical import location.
    if name in {'check_message_extractors', 'compile_catalog', 'extract_messages', 'init_catalog', 'update_catalog'}:
        from babel.messages import setuptools_frontend

        return getattr(setuptools_frontend, name)

    raise AttributeError(f"module {__name__!r} has no attribute {name!r}")

```

This is kind of a big deal also because setuptools is a relatively very slow import, great to be able to skip it if not necessary!

oprypin (commented):
Nice one!
I have only one big comment about the legacy re-exports

oprypin:
Also checked that MkDocs' setup.py integration (before it was removed) works with this change (also with my suggested edit)

babel/messages/frontend.py
akx: Yeah, this is probably the better approach. I keep forgetting about module-level `__getattr__` 😁

PR #1042, opened 2023-11-11
fix(numbers): correctly determine malformed decimals

### LINKED ISSUE #928: parse_decimal raises error for decimals with thousand separator and trailing zeroes
# Bug report

`parse_decimal` raises a surprising NumberFormatError when parsing numbers with thousand separator and non-zero decimals but trailing zeros, when strict mode is set:
```
In [1]: parse_decimal('3,400.60',locale='en',strict=True)`
---------------------------------------------------------------------------`
NumberFormatError                         Traceback (most recent call last)
<ipython-input-6-ccc56249e0c2> in <module>
----> 1 parse_decimal('3,400.60', locale='en', strict=True)

~/iit_repos/hrreport/.tox/test/lib/python3.10/site-packages/babel/numbers.py in parse_decimal(string, locale, strict)
    765                     ), suggestions=[proper])
    766                 else:
--> 767                     raise NumberFormatError((
    768                         "%r is not a properly formatted decimal number. Did you mean %r? Or maybe %r?" %
    769                         (string, proper, proper_alt)

NumberFormatError: '3,400.60' is not a properly formatted decimal number. Did you mean '3,400.6'? Or maybe '3.4006'?`
```
Working variants:
```
In [2]: parse_decimal('3400.60', locale='en', strict=True)
Out[2]: Decimal('3400.60')

In [3]: parse_decimal('3,400.6', locale='en', strict=True)
Out[3]: Decimal('3400.6')

In [4]: parse_decimal('3,400.00', locale='en', strict=True)
Out[4]: Decimal('3400.00')
```
It is surprising and inconsistent that the presence or absence of the thousand separator should effect the result so.

# Your environment

- CPython versions tested on: 3.11.0, 3.10.6
- Babel versions tested on: 2.10.3, 2.11.0
- locales tested: 'en', 'de'
- Operating system and architecture: `5.15.0-52-generic #58-Ubuntu x86_64 GNU/Linux` (up-to-date Ubuntu 22.04 LTS)

# Additional Information
Thanks for the great library!

Olunusib:
Fixes #928

Problem:

The current code was primarily designed for numbers like `3,400.00`, where the r-stripped value (`3,400.`) matches the expected `proper` (`3,400`) + `.` (`3,400.`) format. It doesn't work for numbers like `3,400.60` since `3,400.6` (the value after r-stripping) is incorrectly perceived as not equal to the expected format, which is the `proper` (`3,400.6`) + `.` (`3,400.6.`), thereby leading to its erroneous detection as improperly formatted. 

Solution:

Added a more robust method for stripping trailing zeros, specifically targeting only those zeros that appear after the last non-zero digit following the decimal symbol.

### DIFF
--- a/babel/numbers.py
+++ b/babel/numbers.py
@@ -1099,7 +1099,7 @@ def parse_decimal(
         raise NumberFormatError(f"{string!r} is not a valid decimal number") from exc
     if strict and group_symbol in string:
         proper = format_decimal(parsed, locale=locale, decimal_quantization=False, numbering_system=numbering_system)
-        if string != proper and string.rstrip('0') != (proper + decimal_symbol):
+        if string != proper and proper != _remove_trailing_zeros_after_decimal(string, decimal_symbol):
             try:
                 parsed_alt = decimal.Decimal(string.replace(decimal_symbol, '')
                                                    .replace(group_symbol, '.'))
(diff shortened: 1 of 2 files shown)


babel/util.py
```diff
@@ -134,6 +134,16 @@ def parse_future_flags(fp: IO[bytes], encoding: str = 'latin-1') -> int:
     return flags
 
 
+def remove_trailing_zeros_after_decimal(string: str, decimal_symbol: str) -> str:
+    parts = string.split(decimal_symbol)
```
akx, line 138: This looks a bit suspicious, if `string` has more than one `decimal_symbol`. Should this use `partition` or `rpartition` instead, à la

```suggestion
    integer_part, _, decimal_part = string.partition(decimal_symbol)
```

babel/util.py
```diff
@@ -134,6 +134,16 @@ def parse_future_flags(fp: IO[bytes], encoding: str = 'latin-1') -> int:
     return flags
 
 
+def remove_trailing_zeros_after_decimal(string: str, decimal_symbol: str) -> str:
```
akx, line 137: I don't think this should be a public-smelling API – maybe move it to `babel/numbers.py` and rename it to `_remove_trailing_zeros_after_decimal`?

Also, a docstring would be great in any case :)

babel/util.py
Olunusib: Thanks for the suggestion. I've used `partition` to ensure robustness. While there are already checks that should prevent multiple `decimal_symbol` at that stage of the function, I agree that using `partition` adds a layer of safety and better communicates what the function is trying to do.

babel/util.py
Olunusib: That's true. The function has been moved to `babel/numbers.py`, renamed to `_remove_trailing_zeros_after_decimal`, and now includes a docstring.

akx (approved):
Thank you! I took the liberty of rebasing this, cleaning up the logic a bit and adding a few more test cases for good measure.

PR #1044, opened 2023-11-24
Allow last_translator to be passed as an option to extract_message

AivGitHub:
Commit allows to pass last translator to the CLI.

For example,
`pybabel extract --last-translator=test@test.test`

### DIFF
--- a/babel/messages/frontend.py
+++ b/babel/messages/frontend.py
@@ -339,6 +339,8 @@ class ExtractMessages(CommandMixin):
          'Separate multiple patterns with spaces (default ".* ._")'),
         ('header-comment=', None,
          'header comment for the catalog'),
+        ('last-translator=', None,
+         'set the name and email of the last translator in output'),
     ]
     boolean_options = [
         'no-default-keywords', 'no-location', 'omit-header', 'no-wrap',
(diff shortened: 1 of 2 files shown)


tests/messages/test_frontend.py
```diff
@@ -1535,6 +1535,12 @@ def test_extract_cli_knows_dash_s():
     assert cmdinst.strip_comments
 
 
+def test_extract_cli_knows_dash_dash_last_dash_translator():
+    cmdinst = configure_cli_command("extract --last-translator e-mail -o foo babel")
+    assert isinstance(cmdinst, ExtractMessages)
+    assert cmdinst.last_translator
```
akx, line 1541: This should presumably check that it was assigned to `e-mail`?

babel/messages/frontend.py
```diff
@@ -339,6 +339,8 @@ class ExtractMessages(CommandMixin):
          'Separate multiple patterns with spaces (default ".* ._")'),
         ('header-comment=', None,
          'header comment for the catalog'),
+        ('last-translator=', 'T',
```
akx, line 342: Maybe skip the shortname?
```suggestion
        ('last-translator=', None,
```

akx (changes requested):

tests/messages/test_frontend.py
AivGitHub: Sure you are right. Actually it can be not only an e-mail, fixed.

babel/messages/frontend.py
AivGitHub: I'm good with that. Changed.

PR #1051, opened 2023-12-21
Fallback to LC_MONETARY when formatting currency

afh:
POSIX defines the [`LC_MONETARY` category](https://pubs.opengroup.org/onlinepubs/9699919799/basedefs/V1_chap07.html) to be used for monetary formatting, hence babel could be more standards-compliant in this regard.

### DIFF
--- a/babel/numbers.py
+++ b/babel/numbers.py
@@ -7,7 +7,7 @@
     The default locale for the functions in this module is determined by the
     following environment variables, in that order:
 
-     * ``LC_NUMERIC``,
+     * ``LC_NUMERIC`` or ``LC_MONETARY`` for currency related functions,
      * ``LC_ALL``, and
      * ``LANG``
 
@@ -31,6 +31,7 @@ from babel.localedata import LocaleDataDict
 if TYPE_CHECKING:
     from typing_extensions import Literal
 
+LC_MONETARY = default_locale('LC_MONETARY')
 LC_NUMERIC = default_locale('LC_NUMERIC')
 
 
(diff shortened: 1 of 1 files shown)


afh:
Friendly ping to @akx and @oprypin to get some eyes on this. Not sure if you are the right folks to ping, yet looking at other PRs it seems you would at least know who else to get involved 🙂

babel/numbers.py
```diff
@@ -7,7 +7,7 @@
     The default locale for the functions in this module is determined by the
     following environment variables, in that order:
 
-     * ``LC_NUMERIC``,
+     * ``LC_NUMERIC`` or ``LC_MONETARY`` for currency related functions,
```
akx, line 10: This documentation doesn't seem to match the new behavior; it's not an "or" if the envvar is set. This should probably say 

```suggestion
     * ``LC_MONETARY`` for currency related functions,
     * ``LC_NUMERIC`` for other numbers,
```
or something along those lines.

akx (changes requested):
This is technically okay, but we really should fall back to `LC_NUMERIC` if `LC_MONETARY` isn't set so as to match the old behavior for people who don't have `LC_MONETARY` set; right now `default_locale` would then fall back to `LC_ALL`, `LANG`, etc.

You'd need to change `default_locale()` to accept multiple "primary options" before those fallbacks.

Furthermore, previously, `LC_MONETARY` was ignored altogether, but now it will be used if set for money formatting, which could be a surprising change for people who have it set. This change will need to be thoroughly documented in the changelog.

afh:
Thanks for the review and feedback, @akx, much appreciated. I'll look into addressing your comments and updating this PR accordingly. Not sure when I can get to it, but I'll try to make time soon.

akx:
@afh Do you have time to maybe look at this PR? Could get it merged for 2.16 :)

afh:
Thanks for reaching out, @akx 🙂 
Unfortunately I cannot make the time in the foreseeable future to address your suggested improvements, as much as I wish I could 😕
I might be able to test, if that would be helpful and someone else would be willing to take on the implementation.

akx:
@afh No worries, thanks for replying! I can take this over, no problem.

afh:
Very much appreciated, @akx! 🙏 Do ping me if there's something that can be tested.

Issue #1061, opened 2024-01-25
Unexpected behaviour: parse_numbers() doesn't handle spaces as a grouping symbol, however, parse_decimal() does


donnillo:
## Overview Description
Function ```parse_decimal()``` from ```babel.numbers``` module is wise enough to handle spaces as a grouping symbol where non-breakable space ```\xa0``` is expected. However, ```parse_number()``` does not reproduce the same logic.

## Steps to Reproduce

```python
from babel.numbers import (
    parse_number,
    parse_decimal
)

parse_decimal('1 099', locale='ru', numbering_system='default')  # OK
```
```shell
Decimal('1099')
```
```python
parse_number('1 099', locale='ru', numbering_system='default')  # FAILS UNEXPECTEDLY
```
```shell
Traceback (most recent call last):
  File "/home/myuser/.pyenv/versions/3.11.2/lib/python3.11/site-packages/babel/numbers.py", line 1030, in parse_number
    return int(string.replace(get_group_symbol(locale, numbering_system=numbering_system), ''))
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
ValueError: invalid literal for int() with base 10: '1 099'

The above exception was the direct cause of the following exception:

Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/home/myuser/.pyenv/versions/3.11.2/lib/python3.11/site-packages/babel/numbers.py", line 1032, in parse_number
    raise NumberFormatError(f"{string!r} is not a valid number") from ve
babel.numbers.NumberFormatError: '1 099' is not a valid number
```

## Actual Results
```parse_number()``` unreasonably fails while trying to parse number with spaces in locales where non-breakable space is used as a grouping symbol.

## Expected Results
Space and non-breakable space should be considered the same while parsing numbers. Logic used in ```parse_decimal()``` function should be used in ```parse_number()``` as well.
```python
parse_number('1 099', locale='ru', numbering_system='default')
```
```shell
1099
```

## Reproducibility
Python 3.11.2
WSL\Ubuntu
Babel 2.14.0

## Additional Information
I guess, ```parse_number()``` should also introduce ```strict``` argument as well as ```parse_decimal()``` does.

PR #1062, opened 2024-01-31
Fix: Unexpected behaviour: parse_numbers() doesn't handle space as a grouping symbol, however, parse_decimal() does #1061

sebas-inf:
Cleaned the string to ensure that it still works if there is a non-breakable space, \xa0, in the number and separed the logic so that it is easeir to understand.

### DIFF
--- a/babel/messages/pofile.py
+++ b/babel/messages/pofile.py
@@ -80,6 +80,50 @@ def denormalize(string: str) -> str:
         return unescape(string)
 
 
+def _extract_locations(line: str) -> list[str]:
+    """Extract locations from location comments.
+
+    Locations are extracted while properly handling First Strong
+    Isolate (U+2068) and Pop Directional Isolate (U+2069), used by
+    gettext to enclose filenames with spaces and tabs in their names.
+    """
+    if "\u2068" not in line and "\u2069" not in line:
+        return line.lstrip().split()
+
+    locations = []
+    location = ""
+    in_filename = False
+    for c in line:
+        if c == "\u2068":
+            if in_filename:
+                raise ValueError("location comment contains more First Strong Isolate "
+                                 "characters, than Pop Directional Isolate characters")
+            in_filename = True
+            continue
+        elif c == "\u2069":
+            if not in_filename:
+                raise ValueError("location comment contains more Pop Directional Isolate "
+                                 "characters, than First Strong Isolate characters")
+            in_filename = False
+            continue
+        elif c == " ":
+            if in_filename:
+                location += c
+            elif location:
+                locations.append(location)
+                location = ""
+        else:
+            location += c
+    else:
+        if location:
+            if in_filename:
+                raise ValueError("location comment contains more First Strong Isolate "
+                                 "characters, than Pop Directional Isolate characters")
+            locations.append(location)
+
+    return locations
+
+
 class PoFileError(Exception):
     """Exception thrown by PoParser when an invalid po file is encountered."""
 
@@ -269,7 +313,7 @@ class PoFileParser:
         self._finish_current_message()
 
         if line[1:].startswith(':'):
-            for location in line[2:].lstrip().split():
+            for location in _extract_locations(line[2:]):
                 pos = location.rfind(':')
                 if pos >= 0:
                     try:
@@ -307,7 +351,10 @@ class PoFileParser:
                 if line[1:].startswith('~'):
                     self._process_message_line(lineno, line[2:].lstrip(), obsolete=True)
                 else:
-                    self._process_comment(line)
+                    try:
+                        self._process_comment(line)
+                    except ValueError as exc:
+                        self._invalid_pofile(line, lineno, str(exc))
             else:
                 self._process_message_line(lineno, line)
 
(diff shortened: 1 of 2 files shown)


akx (changes requested):
You've committed and pushed your entire `.venv` – please clean that up first.

sebas-inf:
> You've committed and pushed your entire `.venv` – please clean that up first.

Should I just push the file I changed or everything except the .venv?

akx:
> Should I just push the file I changed or everything except the .venv?

Your commit(s) should only have relevant changes.

sebas-inf:
> > Should I just push the file I changed or everything except the .venv?
> 
> Your commit(s) should only have relevant changes.

Ok, I will push the file with the fix only then.

sebas-inf:
> You still have 755 files in this pull request. You'll probably need to `git reset 40e60a1f6cf178d9f57fcc14f157ea1b2ab77361` (the current `master`), then carefully `git add` only the changes you want, `git commit` them, and `git push --force-with-lease` them to your repository.

should i do all of this within the terminal in vscode where i was working?

akx:
> should i do all of this within the terminal in vscode where i was working?

Maybe. I don't use Visual Studio Code much. Visual Studio Code might have other tools for backing out commits.

sebas-inf:
> > should i do all of this within the terminal in vscode where i was working?
> 
> Maybe. I don't use Visual Studio Code much. Visual Studio Code might have other tools for backing out commits.

It worked, I made it so i only pushed the numbers.py file, how do i change that in this pr now?

babel/numbers.py
```diff
@@ -1027,11 +1029,21 @@ def parse_number(
     :raise `UnsupportedNumberingSystemError`: if the numbering system is not supported by the locale.
     """
     try:
-        return int(string.replace(get_group_symbol(locale, numbering_system=numbering_system), ''))
+        # Get the group symbol from the locale
+        group_symbol = get_group_symbol(locale, numbering_system=numbering_system)
+
+        # Replace non-breakable spaces with the group symbol
+        cleaned_string = string.replace('\xa0', group_symbol)
```
akx, line 1036: Could you add a test case (either a doctest or a real test in `tests/`) that uses an `\xa0` character?

akx (changes requested):
I would like to err on the side of caution here, if there is the remotest chance this could cause mis-parsing of numbers (e.g. something that couldn't be parsed before being now accidentally parsed as something it's not intended to be parsed as).

To that end, maybe this function should also have a `strict=True` parameter like `parse_decimal` does, and then maybe the grouping character logic could be shared between the two in non-strict mode?

babel/numbers.py
sebas-inf: would this be a valid test case for that: >>> parse_number('1 099', locale='ru')

sebas-inf: made some changes applying some of the logic from parse_decimal and it says its passing the test cases on my end

babel/numbers.py
```diff
@@ -1001,6 +1001,7 @@ def __init__(self, message: str, suggestions: list[str] | None = None) -> None:
 def parse_number(
     string: str,
     locale: Locale | str | None = LC_NUMERIC,
+    strict: bool = False,
```
akx, line 1004: Please add the new argument into the docstring. `strict` mode should probably be the default here, and the new behavior (which allows more loose strings to be parsed) should be opt-in.

babel/numbers.py
```diff
@@ -1010,6 +1011,8 @@ def parse_number(
     1099
     >>> parse_number('1.099', locale='de_DE')
     1099
+    >>> parse_number('1 099', locale='ru')
```
akx, line 1014: If the intent is to test a string that has a non-breaking space character, make it obvious:
```suggestion
    >>> parse_number('1\xa0099', locale='ru')
```

babel/numbers.py
```diff
@@ -1027,11 +1030,67 @@ def parse_number(
...
+                        )
+                    else:
+                        raise NumberFormatError(
+                            f"{string!r} is not a properly formatted decimal number. "
+                            f"Did you mean {proper!r}? Or maybe {proper_alt!r}?",
+                            suggestions=[proper, proper_alt],
+                        )
+
+        # Remove other spaces and replace the group symbol with an empty string
+        cleaned_string = string.replace(' ', '').replace(group_symbol, '')
```
akx, line 1085: `format_number` is not meant to parse decimal numbers, only ever integers, anything decimal-related shouldn't be here at all.

If anything, we should check for the existence of a decimal separator in the input in strict mode, and fail if it is there.

babel/numbers.py
```diff
@@ -1027,11 +1030,67 @@ def parse_number(
...
+        # Get the group symbol from the locale
+        group_symbol = get_group_symbol(locale, numbering_system=numbering_system)
+
+        # Replace non-breakable spaces with the group symbol in non-strict mode
+        if not strict and group_symbol == '\xa0' and '\xa0' not in string and ' ' in string:
+            string = string.replace(' ', group_symbol)
+
+        # If strict mode is enabled, handle irregular formatting
+        if strict and group_symbol in string:
+            # Your additional logic for strict mode here
```
akx, line 1042: What a strange comment... 😁

akx (changes requested):

babel/numbers.py
sebas-inf: made the changes in the latest commit

akx:
@sebas-inf If you're still around, please see the comments and rebase this to resolve conflicts. Thanks :)

Issue #728, opened 2020-09-20
Multiple units in `format_timedelta()`


jack1142:
Would it be possible to return more than just one unit for passed `timedelta` object?
For example:
```py
>>> humanize_timedelta(timedelta(days=2, hours=5, minutes=2, seconds=5), granularity=3)
'2 days, 5 hours, and 2 minutes'
```

samayer12:
I'm looking at this in my free time. 
Looks like the solution *could* be an extra optional parameter in `format_timedelta()` within `dates.py`.

samayer12:
I think there's probably a good recursive solution to this. [Here's my fork](https://github.com/samayer12/babel/tree/728_timedelta_granularity)

samayer12:
[My first crack](https://github.com/samayer12/babel/commit/dd732b416631a04f79d025354f266c04c7f067fa) at the problem can produce some of your desired output.
Check out `test_format_four_units()` in `test_dates.py`.

However, it doesn't handle instances where the granularity is before the base case of `second`.

samayer12:
Alright, [this commit](https://github.com/samayer12/babel/commit/ae6ac8e0ff1fd46c9e3333aa659cb646197657cb) does everything but add a trailing `s` if the value is greater than 1.

samayer12:
^^^ No it definitely does not lol. Still working on my fork. Handling the case where you have granularity of seconds but say, just one week.

luckydonald:
I'm not sure what the original intention is, but for me it would be to simply have it actually spell out all the information we have and be exact with that.

That means, if I have a year, one week and 2 minutes, I don't wanna ignore the 2 minutes.
Like a `exact=True` parameter would be what I'm looking for.

jack1142:
> That means, if I have a year, one week and 2 minutes, I don't wanna ignore the 2 minutes.

Right now, `format_timedelta()` in such case would return just `1 year`, not `1 year, 1 week` or `1 year, 1 week, 2 minutes`. So with the use of the `granularity` kwarg I proposed, it would for example allow you to specify `2` to show at max two units (i.e. `1 year, 1 week`).

With what you said in mind, I guess it would also make sense to allow passing some value (`-1` perhaps?) which would cause it to show all units without skipping any.

xmo-odoo:
I'd really like this as well, it's often annoying to have bits at or above granularity cut off in "format_timedelta".

[...]

PR #1066, opened 2024-02-21
Support multiple units in format_timedelta()

hartmutobendorf:
Issue #728 describes returning measures for several units when using format_timedelta().

E.g. instead of 

```
10840 seconds
3 hours
```
return for depth=="fullest"

`3 hours 0 minutes 40 seconds`

and for depth=="full"

`3 hours 40 seconds`

Tests are passing, but I haven't tried contributing to babel before, and am not aware of any additional requirements beyond the few lines of source code in this PR. Feedback appreciated.

### DIFF
--- a/babel/dates.py
+++ b/babel/dates.py
@@ -967,10 +967,10 @@ def format_timedelta(
             if pattern is None:
                 return ''
             if (depth=='shallow'):
-                formatted_string = ' '.join([formatted_string, pattern.replace('{0}', str(value))])
+                formatted_string = ' '.join(filter(None, [formatted_string, pattern.replace('{0}', str(value))]))
                 break
             elif ((depth=='full' and value > 0) or depth == 'fullest'):
-                formatted_string = ' '.join([formatted_string, pattern.replace('{0}', str(value))])
+                formatted_string = ' '.join(filter(None, [formatted_string, pattern.replace('{0}', str(value))]))
                 seconds = seconds - value * secs_per_unit
 
     return formatted_string


babel/dates.py
```diff
@@ -862,6 +863,7 @@ def format_timedelta(
     threshold: float = .85,
     add_direction: bool = False,
     format: Literal['narrow', 'short', 'medium', 'long'] = 'long',
+    depth: Literal['shallow', 'full', 'fullest'] = 'shallow',
```
akx, line 866: I'm not sure these names are the best or most descriptive. No matter what they end up being, they do need to be documented, though.

babel/dates.py
```diff
@@ -960,9 +966,14 @@ def _iter_patterns(a_unit):
             # This really should not happen
             if pattern is None:
                 return ''
-            return pattern.replace('{0}', str(value))
-
-    return ''
+            if (depth=='shallow'):
```
akx, line 969:
```suggestion
            if depth == 'shallow':
```

babel/dates.py
```diff
@@ -960,9 +966,14 @@ def _iter_patterns(a_unit):
...
+                formatted_string = ' '.join(filter(None, [formatted_string, pattern.replace('{0}', str(value))]))
+                break
+            elif ((depth=='full' and value > 0) or depth == 'fullest'):
```
akx, line 972:
```suggestion
            elif (depth == 'full' and value > 0) or depth == 'fullest':
```

babel/dates.py
```diff
@@ -960,9 +966,14 @@ def _iter_patterns(a_unit):
             # This really should not happen
             if pattern is None:
                 return ''
-            return pattern.replace('{0}', str(value))
-
-    return ''
+            if (depth=='shallow'):
+                formatted_string = ' '.join(filter(None, [formatted_string, pattern.replace('{0}', str(value))]))
```
akx, line 970: I'm not sure I understand what's going on in this magic incantation 😄 

It's repeated two lines underneath this, so maybe it should be DRYed out too?

akx (changes requested):
Thanks! 

The depth names don't sound great to me, and this would really require some tests, in various locales too.

Issue #1070, opened 2024-04-11
Upgrade to CLDR 44


tomasr8:
CLDR v44 has been released a while ago. Are there any plans to include it in babel?

ThiefMaster:
v45 is [scheduled](https://docs.google.com/spreadsheets/d/1N6inI5R84UoYlRwuCNPBOAP7ri4q2CmJmh8DC5g-S6c/edit#gid=1680747936) to be released in less than a week. Maybe best to aim for that straight away?

akx:
#1071 bumps us up to CLDR 44. Cc @nileshpatra via https://github.com/python-babel/babel/issues/1059#issuecomment-1979589995 – would it be useful for e.g. Debian that there'd be a Babel version that uses CLDR 44, even if CLDR 45 is released soon?

nileshpatra:
On Mon, Apr 15, 2024 at 07:12:43AM -0700, Aarni Koskela wrote:
> #1071 bumps us up to CLDR 44. Cc @nileshpatra via https://github.com/python-babel/babel/issues/1059#issuecomment-1979589995 – would it be useful for e.g. Debian that there'd be a Babel version that uses CLDR 44, even if CLDR 45 is released soon?

I am not sure if I understand that question. Currently CLDR version in Debian is
44 - which is what we need. It probably will be upgraded to 45 at some point.

> That said, the next version of Babel will not officially support Python 3.7 anymore; is that an issue?

No problem.

akx:
> I am not sure if I understand that question. Currently CLDR version in Debian is 44 - which is what we need.

Right, so I think it'll be a good idea to release a version of Babel that basically just bumps up to CLDR 44, just so that version can be easily shipped with the version of Debian that targets CLDR 44.

> It probably will be upgraded to 45 at some point.

As far as I understand Debian's stability/release process, that might not be in the immediate cards as soon as 45 is released, though? :)

nileshpatra:
On 15 April 2024 9:57:06 pm IST, Aarni Koskela ***@***.***> wrote:
>> I am not sure if I understand that question. Currently CLDR version in Debian is 44 - which is what we need.
>
>Right, so I think it'll be a good idea to release a version of Babel that basically just bumps up to CLDR 44, just so that version can be easily shipped with the version of Debian that targets CLDR 44.
>
>> It probably will be upgraded to 45 at some point.
>
>As far as I understand Debian's stability/release process, that might not be in the immediate cards as soon as 45 is released, though? :)

[...]

PR #1071, opened 2024-04-15
Use CLDR 44

akx:
Refs #1070.


ThiefMaster:
Nice! When do you plan to make a release containing the updated CLDR data?

akx:
@ThiefMaster I'd like to get https://github.com/python-babel/babel/pull/1075 into the next version, but  Soon :)

PR #1074, opened 2024-04-19
Include Unicode license in `locale-data` and in documentation

### LINKED ISSUE #1072: Missing license for babel/locale-data
## Overview Description

`babel/locale-data` does not provide any license information, although being based on some third-party code.

## Steps to Reproduce

1. Download a *Babel* distribution.
2. Search for the Unicode license inside the package.

## Actual Results

There is no Unicode-related copyright information inside the distributions.

## Expected Results

The obligations of the Unicode/CLDR license terms are met inside the *Babel* distributions.

## Additional Information

Digging through the code of the repository, it turns out that `babel/locale-data` contains *pickle*-based custom representations of files downloaded from https://github.com/python-babel/babel/blob/fe82fbc90d8044d17bfc4ae1c7a0cb24e85153ef/scripts/download_import_cldr.py#L12 The corresponding ZIP file and repository (https://github.com/unicode-org/cldr) state in their `LICENSE.txt` (bold text/highlighting by me):

> Permission is hereby granted, free of charge, to any person obtaining a copy of data files and any associated documentation (the "Data Files") or software and any associated documentation (the "Software") to deal in the Data Files or Software without restriction, including without limitation the rights to use, copy, modify, merge, publish, distribute, and/or sell copies of the Data Files or Software, and to permit persons to whom the Data Files or Software are furnished to do so, provided that **either (a) this copyright and permission notice appear with all copies of the Data Files or Software, or (b) this copyright and permission notice appear in associated Documentation.**

At the moment, there corresponding notice seems to be missing from *Babel* distributions and from the whole GitHub repository.

akx:
Fixes #1072

### DIFF
--- a/docs/license.rst
+++ b/docs/license.rst
@@ -25,9 +25,13 @@ documentation.
 -   "AUTHORS" hereby refers to all the authors listed in the
     :ref:`authors` section.
 
--   The ":ref:`babel-license`" applies to all the sourcecode shipped as
+-   The ":ref:`babel-license`" applies to all the source code shipped as
     part of Babel (Babel itself as well as the examples and the unit tests)
     as well as documentation.
+
+-   The ":ref:`unicode-license`" applies to the transformed Unicode
+    Common Locale Data Repository (CLDR) data files shipped with Babel,
+    in the directory ``babel/locale-data``.
 
 .. _babel-license:
 
@@ -35,3 +39,10 @@ Babel License
 -------------
 
 .. include:: ../LICENSE
+
+.. _unicode-license:
+
+Unicode License
+---------------
+
+.. include:: ../babel/locale-data/LICENSE.unicode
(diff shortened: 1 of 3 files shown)


docs/license.rst
```diff
@@ -25,13 +25,24 @@ documentation.
...
 -------------
 
 .. include:: ../LICENSE
+
+.. _unicode-license:
+
+Unicode License
+---------------
+
+.. include:: ../babel/localedata/LICENSE.unicode
```
stefan6419846, line 48: I have not tested it, but shouldn't it be `locale-data` here as well?

akx: It absolutely should, great catch. Thanks!

stefan6419846 (approved):
LGTM.

One alternative would be to actually use the license file shipped within the actual CLDR archives to always have the latest/correct license file, but the current approach is also fine for me.

PR #1083, opened 2024-05-12
Replace deprecated `ast.Str` with `ast.Constant`

tomasr8:
Running the tests under python 3.12 I get deprecation warnings related to the usage of `ast.Str` which has been deprecated since 3.8 and will be removed in 3.14.

Since 3.8 is also the lowest version supported by babel, we can safely make the change to use `ast.Constant` everywhere.


akx (approved):
Thanks, LGTM!

PR #1085, opened 2024-05-13
Make pgettext search plurals when translation is not found

tomasr8:
Given this po file:

```po
msgctxt "ctx"
msgid "foo"
msgid_plural "foos"
msgstr[0] "foo translated"
```

`pgettext` fails to find the singular translation i.e. `pgettext("ctx", "foo") == "foo"` when one would expect to get `foo translated`.

The same issue has already been fixed upstream in `gettext`: 
https://github.com/python/cpython/issues/62519
https://github.com/python/cpython/pull/107118

This PR applies the same patch to babel.

### DIFF
--- a/babel/support.py
+++ b/babel/support.py
@@ -466,10 +466,12 @@ class NullTranslations(gettext.NullTranslations):
         missing = object()
         tmsg = self._catalog.get(ctxt_msg_id, missing)
         if tmsg is missing:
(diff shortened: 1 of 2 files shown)


akx (approved):
LGTM, this exactly mirrors https://github.com/python/cpython/pull/107118/files 👍

PR #1093, opened 2024-07-03
Fix for PyBabel Compatibility with Python 3.12 and Third-Party Extractors

podgorniy94:
PyBabel currently fails to work with Python 3.12 when using third-party extractors. Specifically, the extraction method for Jinja2 templates throws a ValueError:

```
[python: **.py]
[jinja2: templates/**.html]
ValueError: Unknown extraction method 'jinja2'
```

This issue does not occur with Python 3.11, where the same setup functions correctly.

Issue: https://github.com/python-babel/babel/issues/861

### DIFF
--- a/babel/messages/checkers.py
+++ b/babel/messages/checkers.py
@@ -157,18 +157,28 @@ def _validate_format(format: str, alternative: str) -> None:
 def _find_checkers() -> list[Callable[[Catalog | None, Message], object]]:
     checkers: list[Callable[[Catalog | None, Message], object]] = []
     try:
+        from pkg_resources import working_set
+    except ImportError:
+        pass
+    else:
+        for entry_point in working_set.iter_entry_points('babel.checkers'):
+            checkers.append(entry_point.load())
+        if checkers:
+            return checkers
+
+    try:
         from importlib.metadata import entry_points
     except ImportError:
         pass
     else:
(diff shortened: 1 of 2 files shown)


akx:
I don't think all of our currently supported platforms have `importlib.metadata`, so this will likely fail CI, but we'll see...

j123b567:
@akx you are right, it should try `pkg_resources` first and if it is not there, then try `importlib.metadata` and if it is not there, then do the fallback to predefined list

akx (changes requested):
As discussed, this needs a fallback for environments where `importlib.metadata.entry_points()` is not available.

podgorniy94:
Another fallback was needed, this time for `importlib.metadata`, specifically for `entry_points()`.

```
Error message:
AttributeError: 'str' object has no attribute 'group'
```

According to the documentation:

> _Compatibility_
> The "selectable" entry points were introduced in importlib_metadata 3.6 and Python 3.10. Prior to those changes, entry_points accepted no parameters and always returned a dictionary of entry points, keyed by group. With importlib_metadata 5.0 and Python 3.12, entry_points always returns an EntryPoints object. See backports.entry_points_selectable for compatibility options.

akx:
@podgorniy94 I wasn't sure you were doing that, so please just see #1102 instead – the commit has you as a co-author :)

Issue #1076, opened 2024-04-24
Add support for other systemic aliases specified in CLDR `root.xml`


akx:
## Overview Description

#892 was caused by us not reading `<alias>`es at all.

#1075 encodes the "hey, you can use `short` if you can't find these time deltas" logic from `root.xml` into code, which fixes that issue. There are other interesting `<alias>`es in `root.xml` we should support:

- [ ] `dates.fields.field`
- [ ] `numbers.decimalFormats.decimalFormatLength`
- [ ] `numbers.scientificFormats`
- [ ] `numbers.percentFormats`
- [ ] `numbers.currencyFormats` (especially that "if accounting isn't specified, use standard")
- [ ] Units (lots of them have name fallback aliases)
- [x] List patterns (#1099)

It looks like (thankfully) there are no `<alias>`es specified in specific locale files, and I don't think we need to support that.

PR #1099, opened 2024-07-11
Support list format fallbacks

### LINKED ISSUE #1098: babel.lists.format_list with style="unit" errors on non-English languages
## Overview Description

babel.lists.format_list does not work with non-English locales, when style is set to "unit".

## Steps to Reproduce

Install latest babel 2.15.0
```py
from babel.lists import format_list
print(format_list([1, 2, 3, 4], style="unit", locale="zh_CN"))
```
## Actual Results

```
  File "<eval command - snippet #20>", line 4, in func
    print(format_list([1, 2, 3, 4], style="unit", locale="zh_CN"))
          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/ubuntu/redenv/lib/python3.11/site-packages/babel/lists.py", line 89, in format_list
    result = patterns['start'].format(lst[0], lst[1])
             ~~~~~~~~^^^^^^^^^
  File "/home/ubuntu/redenv/lib/python3.11/site-packages/babel/localedata.py", line 234, in __getitem__
    orig = val = self._data[key]
                 ~~~~~~~~~~^^^^^
KeyError: 'start'
```
## Expected Results

Something like `1, 2, 3, 4` but locale dependent of course!

## Reproducibility

I can reproduce this in every non-Engilsh language I've tried, es_ES zh_CN fr_FR
Various EN locales en_GB en_US en_CN produce the expected result `1, 2, 3, 4`

## Additional Information

Similar to #781 but this error occurs with [supported style`unit`](https://babel.pocoo.org/en/latest/api/lists.html) and has a wider language scope

akx:
Fixes #1098 by fixing one part of #1076.

As noted in the TODO comment, this isn't the optimal way to implement it, but a better way could be swapped in later.

### DIFF
--- a/babel/lists.py
+++ b/babel/lists.py
@@ -26,9 +26,11 @@ if TYPE_CHECKING:
 DEFAULT_LOCALE = default_locale()
 
 
-def format_list(lst: Sequence[str],
-                style: Literal['standard', 'standard-short', 'or', 'or-short', 'unit', 'unit-short', 'unit-narrow'] = 'standard',
-                locale: Locale | str | None = DEFAULT_LOCALE) -> str:
+def format_list(
+    lst: Sequence[str],
+    style: Literal['standard', 'standard-short', 'or', 'or-short', 'unit', 'unit-short', 'unit-narrow'] = 'standard',
+    locale: Locale | str | None = DEFAULT_LOCALE,
+) -> str:
     """
     Format the items in `lst` as a list.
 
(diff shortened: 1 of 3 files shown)


akx:
Cc @vexed01 for reporting #1098.

Cc @tomasr8 if you feel like taking a quick look!

tomasr8 (approved):
Looks good to me :) Full branch and statement coverage for `babel.lists` as well

As an aside, I was looking for some description for how the list patterns are defined in the XML and I only found this [short explanation](https://cldr.unicode.org/development/development-process/design-proposals/list-formatting). I assume the `<listPattern>` with no type is the _standard_ one?

akx:
> As an aside, I was looking for some description for how the list patterns are defined in the XML and I only found this [short explanation](https://cldr.unicode.org/development/development-process/design-proposals/list-formatting). I assume the `<listPattern>` with no type is the _standard_ one?

It's (vaguely) in the [TR35 doc](https://www.unicode.org/reports/tr35/tr35-72/tr35-general.html#ListPatterns):

> "type attribute value": standard (or no `type`)

PR #1100, opened 2024-07-17
Avoid crashing on importing localtime when TZ is malformed 

### LINKED ISSUE #1092: ValueError from failed import of localtime submodule
## Overview Description

When calling routines like `babel.number.get_decimal_symbol`, an internal import of `babel.localtime` fails under certain circumstances with a ValueError, that may easily slip awareness.

## Steps to Reproduce

1. Have a Unix system
2. set TZ=/UTC
3. in Python 3.9+ without `pytz`, run `babel.numbers.parse_decimal("5.2")`

## Actual Results

```
  File "/usr/local/lib/python3.10/dist-packages/babel/numbers.py", line 1103, in parse_decimal
    group_symbol = get_group_symbol(locale, numbering_system=numbering_system)
  File "/usr/local/lib/python3.10/dist-packages/babel/numbers.py", line 452, in get_group_symbol
    return _get_number_symbols(locale, numbering_system=numbering_system).get('group', ',')
  File "/usr/local/lib/python3.10/dist-packages/babel/numbers.py", line 335, in _get_number_symbols
    return parsed_locale.number_symbols[numbering_system]
  File "/usr/local/lib/python3.10/dist-packages/babel/core.py", line 640, in number_symbols
    return self._data['number_symbols']
  File "/usr/local/lib/python3.10/dist-packages/babel/core.py", line 439, in _data
  File "/usr/local/lib/python3.10/dist-packages/babel/localedata.py", line 137, in load
    data = load(parent).copy()
  File "/usr/local/lib/python3.10/dist-packages/babel/localedata.py", line 137, in load
    data = load(parent).copy()
  File "/usr/local/lib/python3.10/dist-packages/babel/localedata.py", line 137, in load
    data = load(parent).copy()
  File "/usr/local/lib/python3.10/dist-packages/babel/localedata.py", line 143, in load
    data = pickle.load(fileobj)
  File "/usr/local/lib/python3.10/dist-packages/babel/dates.py", line 34, in <module>
    from babel import localtime
  File "/usr/local/lib/python3.10/dist-packages/babel/localtime/__init__.py", line 41, in <module>
    LOCALTZ = get_localzone()
  File "/usr/local/lib/python3.10/dist-packages/babel/localtime/__init__.py", line 37, in get_localzone
    return _get_localzone()
  File "/usr/local/lib/python3.10/dist-packages/babel/localtime/_unix.py", line 36, in _get_localzone
    return _tz_from_env(tzenv)
  File "/usr/local/lib/python3.10/dist-packages/babel/localtime/_unix.py", line 21, in _tz_from_env
    return _get_tzinfo_or_raise(tzenv)
  File "/usr/local/lib/python3.10/dist-packages/babel/localtime/_helpers.py", line 29, in _get_tzinfo_or_raise
    tzinfo = _get_tzinfo(tzenv)
  File "/usr/local/lib/python3.10/dist-packages/babel/localtime/_helpers.py", line 21, in _get_tzinfo
  File "/usr/lib/python3.10/zoneinfo/_tzpath.py", line 67, in find_tzfile
    _validate_tzfile_path(key)
  File "/usr/lib/python3.10/zoneinfo/_tzpath.py", line 81, in _validate_tzfile_path
    raise ValueError(
ValueError: ZoneInfo keys may not be absolute paths, got: /UTC
```

## Expected Results

`Decimal('5.2')`

## Reproducibility

Works repeatedly with many calls, as long as they rely on `localtime`.

## Additional Information

AFAICT the `babel.localtime` submodule is loaded lazily in the course of processing calls like `babel.number.parse_decimal` or `babel.number.get_decimal_symbol`. On systems where `pytz` is not present, `zoneinfo` is loaded in `babel.localtime._unix.py` via  `babel.localtime.__init__.py`, where also `get_localzone` is called. On systems with `TZ` set (and it not being a local file), this triggers a call to `zoneinfo.ZoneInfo` with the value of `TZ`. This may be an invalid value, which raises an uncaught ValueError. This error is raised again on repeated calls, as the import fails.

While the workaround is pretty easy (install `pytz`), the problem here is awareness:

- I don't think many people check the full env list for potential pitfalls. It may even be out of the user's control - especially on cloud services.
- I can't think of a test that catches this accidentally.
- Many of these calls are specifically wrapped to catch `ValueError` under the assumption that this means that the parsing failed due to the input value.

So this is a potential case of "fails only in production silently".

akx:
Fixes #1092.

When `pytz` is _not_ installed, importing `babel.localtime` could fail (repeatedly) when the `TZ` environment variable is malformed enough to be caught by [`_validate_tzfile_path`](https://github.com/python/cpython/blob/ac0745111/Lib/zoneinfo/_tzpath.py#L85), which might throw a certain `ValueError`.

(When `pytz` is installed, it would raise the same `pytz.exceptions.UnknownTimeZoneError` we already catch and ignore for the same input.)

### DIFF
--- a/babel/localtime/_helpers.py
+++ b/babel/localtime/_helpers.py
@@ -2,7 +2,11 @@ try:
     import pytz
 except ModuleNotFoundError:
     pytz = None
+
+try:
     import zoneinfo
+except ModuleNotFoundError:
+    zoneinfo = None
 
 
 def _get_tzinfo(tzenv: str):
@@ -19,6 +23,16 @@ def _get_tzinfo(tzenv: str):
     else:
         try:
             return zoneinfo.ZoneInfo(tzenv)
+        except ValueError as ve:
+            # This is somewhat hacky, but since _validate_tzfile_path() doesn't
+            # raise a specific error type, we'll need to check the message to be
+            # one we know to be from that function.
+            # If so, we pretend it meant that the TZ didn't exist, for the benefit
+            # of `babel.localtime` catching the `LookupError` raised by
+            # `_get_tzinfo_or_raise()`.
+            # See https://github.com/python-babel/babel/issues/1092
+            if str(ve).startswith("ZoneInfo keys "):
+                return None
         except zoneinfo.ZoneInfoNotFoundError:
             pass
 
--- a/tests/test_localtime.py
+++ b/tests/test_localtime.py
@@ -1,0 +1,29 @@
+import sys
+
+import pytest
+
+from babel.localtime import _helpers, get_localzone
+
+
+@pytest.mark.skipif(
+    sys.platform == "win32",
+    reason="Issue 1092 is not applicable on Windows",
+)
+def test_issue_1092_without_pytz(monkeypatch):
+    pytest.importorskip("zoneinfo", reason="zoneinfo is not available")
+    monkeypatch.setenv("TZ", "/UTC")  # Malformed timezone name.
+    # In case pytz _is_ also installed, we want to pretend it's not, so patch it out...
+    monkeypatch.setattr(_helpers, "pytz", None)
+    with pytest.raises(LookupError):
+        get_localzone()
+
+
+@pytest.mark.skipif(
+    sys.platform == "win32",
+    reason="Issue 1092 is not applicable on Windows",
+)
+def test_issue_1092_with_pytz(monkeypatch):
+    pytest.importorskip("pytz", reason="pytz is not installed")
+    monkeypatch.setenv("TZ", "/UTC")  # Malformed timezone name.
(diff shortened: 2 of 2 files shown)


babel/localtime/_helpers.py
```diff
@@ -19,6 +23,16 @@ def _get_tzinfo(tzenv: str):
     else:
```
matthiashuschle, line 23: as `zoneinfo` may be None, should we check for that here?

akx: `pytz` is a required dep when Python < 3.9; `zoneinfo` should be available >= 3.9...

matthiashuschle:
Looks good to me.

PR #1102, opened 2024-07-18
Allow use of importlib.metadata for finding entrypoints

akx:
Closes #1093 (supersedes it, based on it).
Refs #861.

This PR makes Babel attempt to use `importlib.metadata` instead of `pkg_resources` where available.

It also adds a test that proves the Jinja2 extractor also works on Python 3.12; this is ran in CI as well.

### DIFF
--- a/babel/messages/_compat.py
+++ b/babel/messages/_compat.py
@@ -1,3 +1,4 @@
+import sys
 from functools import partial
 
 
@@ -9,19 +10,20 @@ def find_entrypoints(group_name: str):
     Yields tuples of the entrypoint name and a callable function that will
     load the actual entrypoint.
     """
-    try:
-        from importlib.metadata import entry_points
-    except ImportError:
-        pass
-    else:
-        eps = entry_points()
-        if isinstance(eps, dict):  # Old structure before Python 3.10
-            group_eps = eps.get(group_name, [])
-        else:  # New structure in Python 3.10+
-            group_eps = (ep for ep in eps if ep.group == group_name)
-        for entry_point in group_eps:
-            yield (entry_point.name, entry_point.load)
-        return
+    if sys.version_info >= (3, 10):
+        # "Changed in version 3.10: importlib.metadata is no longer provisional."
+        try:
+            from importlib.metadata import entry_points
+        except ImportError:
+            pass
+        else:
+            eps = entry_points(group=group_name)
+            # Only do this if this implementation of `importlib.metadata` is
+            # modern enough to not return a dict.
+            if not isinstance(eps, dict):
+                for entry_point in eps:
+                    yield (entry_point.name, entry_point.load)
+                return
 
     try:
         from pkg_resources import working_set
--- a/babel/messages/_compat.py
+++ b/babel/messages/_compat.py
@@ -1,0 +1,32 @@
+from functools import partial
+
+
+def find_entrypoints(group_name: str):
+    """
+    Find entrypoints of a given group using either `importlib.metadata` or the
+    older `pkg_resources` mechanism.
+
+    Yields tuples of the entrypoint name and a callable function that will
+    load the actual entrypoint.
+    """
+    try:
+        from importlib.metadata import entry_points
+    except ImportError:
+        pass
+    else:
+        eps = entry_points()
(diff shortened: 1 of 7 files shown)


akx:
cc @tomasr8 @podgorniy94

babel/messages/_compat.py
```diff
@@ -0,0 +1,32 @@
...
+            yield (entry_point.name, entry_point.load)
+        return
+
+    try:
+        from pkg_resources import working_set
+    except ImportError:
+        pass
+    else:
+        for entry_point in working_set.iter_entry_points(group_name):
+            yield (entry_point.name, partial(entry_point.load, require=True))
```
podgorniy94, line 32: As my colleague @j123b567 rightly noted, for versions before 3.10, it is better to use `pkg_resources`, as the `importlib.metadata` API is still unstable. 
You can use `pkg_resources` as the first handler, and in the second try block, check that the object type is not a dictionary before using `importlib.metadata`. Otherwise, simply continue the function execution. 
Consider this option as well.

tomasr8: Could you post this as a code suggestion so it's easier to see what you mean?

podgorniy94: > Could you post this as a code suggestion so it's easier to see what you mean?

yeah

podgorniy94:
```suggestion
    try:
        from pkg_resources import working_set
    except ImportError:
        pass
    else:
        for entry_point in working_set.iter_entry_points(group_name):
            yield (entry_point.name, partial(entry_point.load, require=True))
        return
    try:
        from importlib.metadata import entry_points
    except ImportError:
        pass
    else:
        eps = entry_points()
        if not isinstance(eps, dict):  # New structure in Python 3.10+
            group_eps = (ep for ep in eps if ep.group == group_name)
            for entry_point in group_eps:
                yield (entry_point.name, entry_point.load)
```

akx: Good idea – I pushed a commit with a variation on this idea, namely to only even try importlib.metadata on Python 3.10+, when we know it will be stable.

babel/messages/_compat.py
```diff
@@ -0,0 +1,31 @@
...
+    older `pkg_resources` mechanism.
+
+    Yields tuples of the entrypoint name and a callable function that will
+    load the actual entrypoint.
+    """
+    if sys.version_info >= (3, 10):
+        # "Changed in version 3.10: importlib.metadata is no longer provisional."
+        try:
+            from importlib.metadata import entry_points
+        except ImportError:
```
tomasr8, line 17: Is the try/except block needed if we have an explicit version check? `importlib.metadata` should always be available on 3.10+

babel/messages/_compat.py
podgorniy94: In my initial local implementation, there was a strict check for the Python version, but we decided to abandon this approach since various Python interpreters (e.g., PyPy) and, consequently, different standard modules can be used. My proposed version is universal and not tied to a specific version, but only to the availability of the library.

babel/messages/_compat.py
podgorniy94: > Is the try/except block needed if we have an explicit version check? `importlib.metadata` should always be available on 3.10+

https://github.com/python-babel/babel/pull/1102#discussion_r1684164348

akx: As @podgorniy94 mentioned, I think it's a good thing to double-check, for supporting some more esoteric environments.

podgorniy94: > As @podgorniy94 mentioned, I think it's a good thing to double-check, for supporting some more esoteric environments.

Actually, there's no need to check the Python version. We simply try `pkg_resources`, and if the library isn't available, then we attempt `importlib.metadata`. To avoid strict version checking for, as you said, esoteric environments where there might be different version labeling :)

akx: Yes and no – using `pkg_resources` where it's deprecated will raise a DeprecationWarning, which in some (in e.g. my work-work) environments deprecation warnings are hard errors in tests.

podgorniy94 (approved):

PR #1105, opened 2024-07-22
Enclose white spaces in references

### LINKED ISSUE #654: Extract messes up file names with whitespace
If a file with whitespace is extracted with the command line tool, the comments include it as 2 different files split on the space, and then sorts them individually. Example output:
```
#: landing.html:77 web/templates/logged_out/endnote-landing.html:131
#: web/templates/logged_out/wos web/templates/partials/home_page.html:319
```

Where one of the files is actually "wos landing.html".

Not an urgent issue, I only noticed this while looking at how to handle this edge case for my own code.

Dunedan:
Since version 0.22 gettext encloses file names in references which contain white spaces or tabs within First Strong Isolate (U+2068) and Pop Directional Isolate (U+2069). This commit adds the same behavior for Babel.

Fixes #654

### DIFF
--- a/babel/messages/pofile.py
+++ b/babel/messages/pofile.py
@@ -626,6 +626,11 @@ def generate_po(
 
             for filename, lineno in locations:
                 location = filename.replace(os.sep, '/')
+                if " " in location or "\t" in location:
+                    if not location.startswith("\u2068"):
+                        location = "\u2068" + location
+                    if not location.endswith("\u2069"):
+                        location += "\u2069"
                 if lineno and include_lineno:
                     location = f"{location}:{lineno:d}"
                 if location not in locs:
--- a/tests/messages/test_pofile.py
+++ b/tests/messages/test_pofile.py
@@ -841,6 +841,46 @@ msgstr ""'''
 msgid "foo"
 msgstr ""'''
 
+    def test_white_space_in_location(self):
+        catalog = Catalog()
+        catalog.add('foo', locations=[('main.py', 1)])
+        catalog.add('foo', locations=[('utils b.py', 3)])
+        buf = BytesIO()
+        pofile.write_po(buf, catalog, omit_header=True, include_lineno=True)
+        assert buf.getvalue().strip() == b'''#: main.py:1 \xe2\x81\xa8utils b.py\xe2\x81\xa9:3
+msgid "foo"
+msgstr ""'''
+
+    def test_white_space_in_location_already_enclosed(self):
+        catalog = Catalog()
+        catalog.add('foo', locations=[('main.py', 1)])
+        catalog.add('foo', locations=[('\u2068utils b.py\u2069', 3)])
+        buf = BytesIO()
+        pofile.write_po(buf, catalog, omit_header=True, include_lineno=True)
+        assert buf.getvalue().strip() == b'''#: main.py:1 \xe2\x81\xa8utils b.py\xe2\x81\xa9:3
+msgid "foo"
+msgstr ""'''
+
+    def test_tab_in_location(self):
+        catalog = Catalog()
+        catalog.add('foo', locations=[('main.py', 1)])
+        catalog.add('foo', locations=[('utils\tb.py', 3)])
+        buf = BytesIO()
+        pofile.write_po(buf, catalog, omit_header=True, include_lineno=True)
+        assert buf.getvalue().strip() == b'''#: main.py:1 \xe2\x81\xa8utils        b.py\xe2\x81\xa9:3
+msgid "foo"
+msgstr ""'''
+
+    def test_tab_in_location_already_enclosed(self):
+        catalog = Catalog()
+        catalog.add('foo', locations=[('main.py', 1)])
+        catalog.add('foo', locations=[('\u2068utils\tb.py\u2069', 3)])
+        buf = BytesIO()
+        pofile.write_po(buf, catalog, omit_header=True, include_lineno=True)
+        assert buf.getvalue().strip() == b'''#: main.py:1 \xe2\x81\xa8utils        b.py\xe2\x81\xa9:3
+msgid "foo"
+msgstr ""'''
+
 
 class PofileFunctionsTestCase(unittest.TestCase):
 


akx:
Hi! Thanks for the contribution – could you link to gettext changelog/code for this?

Dunedan:
> Hi! Thanks for the contribution – could you link to gettext changelog/code for this?

Sorry, of course. Here you go:

https://git.savannah.gnu.org/gitweb/?p=gettext.git;a=commitdiff;h=4cacad44e67171ce136cf5c1ae01093aa8e7a7ad
https://git.savannah.gnu.org/gitweb/?p=gettext.git;a=blob;f=NEWS;h=4f126cab47b2eb462fbb844f5c785ab905661c3f;hb=HEAD#l68

babel/messages/pofile.py
```diff
@@ -626,6 +626,11 @@ def _format_message(message, prefix=''):
 
             for filename, lineno in locations:
                 location = filename.replace(os.sep, '/')
+                if " " in location or "\t" in location:
+                    if not location.startswith("\u2068"):
+                        location = "\u2068" + location
+                    if not location.endswith("\u2069"):
+                        location += "\u2069"
```
akx, line 633: Could this be refactored into a separate module-internal helper function (`_maybe_enclose_filename` for instance?) that has references to the relevant Gettext code?

As it is, it's a bit non-obvious why this is here :)

akx (changes requested):
Beyond the inline comment – we should probably also strip these delimiters when parsing a .po file? A roundtrip test would be neat.

Dunedan:
> Beyond the inline comment – we should probably also strip these delimiters when parsing a .po file?

Yes of course. I didn't think about that initially, but this makes perfect sense. I added the necessary logic for that.

> A roundtrip test would be neat.

I added a bunch of additional tests, including a roundtrip test.

babel/messages/pofile.py
```diff
@@ -80,6 +80,47 @@ def denormalize(string: str) -> str:
...
+    gettext to enclose filenames with spaces and tabs in their names.
+    """
+    locations = []
+    location = ""
+    in_filename = False
+    for c in line:
+        if c == "\u2068":
+            if in_filename:
+                raise ValueError("reference contains more First Strong Isolate characters, "
+                                 "than Pop Directional Isolate characters")
```
tomasr8, line 97: This might be a bit clearer?
```suggestion
                raise ValueError("File location contains more First Strong Isolate characters, "
                                 "than Pop Directional Isolate characters")
```

tomasr8 (approved):
Just one nit, otherwise looks good to me, thanks!

babel/messages/pofile.py
Dunedan: I tried to stick to the wording gettext uses, that's why it's called reference there. While I agree that it might not be perfectly clear, I wonder if file location is really clearer.

tomasr8: The CLI uses a combination of `location`, `location comments` and `file location`, but I don't think `reference` is ever used to refer to locations. Maybe `location comment` is better?

Dunedan: I changed it to "location comment" now.

tomasr8:
@akx in case you missed this ;)

tomasr8:
> > @akx in case you missed this ;)
> 
> I was getting married and on holiday. Back in the proverbial office now :)

No worries then :) And Congrats! :tada:

babel/messages/pofile.py
```diff
@@ -80,6 +80,47 @@ def denormalize(string: str) -> str:
...
 
 
+def _extract_locations(line: str) -> list[str]:
+    """Extract locations from location comments.
+
+    Locations are extracted while properly handling First Strong
+    Isolate (U+2068) and Pop Directional Isolate (U+2069), used by
+    gettext to enclose filenames with spaces and tabs in their names.
+    """
+    locations = []
```
akx, line 90: I think it might be a good idea to have a fast path here. Most projects will likely not have the isolate characters in `line`...

```suggestion
    if not ("\u2068" in line or "\u2069" in line):
        return line.lstrip().split()
    locations = []
```

or similar (the `line.lstrip().split()` being what we used to do).

---

Testing against [this large-ish translation file](https://raw.githubusercontent.com/bit-team/backintime/dev/common/po/fi.po):

### readpobench.py

```python
import io

from babel.messages.pofile import read_po

po_data = open("fi.po", "rb").read()


def fun():
    assert read_po(io.BytesIO(po_data))


for x in range(500):
    fun()
```

```
$ hyperfine 'git checkout master && python readpobench.py' 'git checkout enclose-whitespaces && python readpobench.py' 'git checkout enclose-whitespaces-fast-path && python readpobench.py'
Benchmark 1: git checkout master && python readpobench.py
  Time (mean ± σ):      1.538 s ±  0.005 s    [User: 1.493 s, System: 0.042 s]
  Range (min … max):    1.529 s …  1.547 s    10 runs

Benchmark 2: git checkout enclose-whitespaces && python readpobench.py
  Time (mean ± σ):      1.744 s ±  0.022 s    [User: 1.690 s, System: 0.047 s]
  Range (min … max):    1.710 s …  1.784 s    10 runs

Benchmark 3: git checkout enclose-whitespaces-fast-path && python readpobench.py
  Time (mean ± σ):      1.555 s ±  0.017 s    [User: 1.508 s, System: 0.042 s]
  Range (min … max):    1.537 s …  1.589 s    10 runs

Summary
  git checkout master && python readpobench.py ran
    1.01 ± 0.01 times faster than git checkout enclose-whitespaces-fast-path && python readpobench.py
    1.13 ± 0.02 times faster than git checkout enclose-whitespaces && python readpobench.py
```

akx (changes requested):
One performance comment :)

babel/messages/pofile.py
Dunedan: Sounds good to me. I just pushed a commit with slightly modified version of that.

babel/messages/pofile.py
```diff
@@ -80,6 +80,50 @@ def denormalize(string: str) -> str:
...
 
 
+def _extract_locations(line: str) -> list[str]:
+    """Extract locations from location comments.
+
+    Locations are extracted while properly handling First Strong
+    Isolate (U+2068) and Pop Directional Isolate (U+2069), used by
+    gettext to enclose filenames with spaces and tabs in their names.
+    """
+    if "\u2068" not in line and "\u2069" not in line:
```
akx, line 90: TBH, I think looking for only one of these is enough for us – it should be safe to assume that a .po writer that writes these characters to begin with will write both of them.

```suggestion
    if "\u2068" not in line:
```

Dunedan: When checking only for First Strong Isolate and parsing a corrupt file, one might end up with the Pop Directional Isolate being part of the extracted locations:

```
>>> "first location\u2069 second_location".lstrip().split()
['first', 'location\u2069', 'second_location']
```

I believe raising an error would be more appropriate in that situation.

akx: Practically speaking, I don't think files will (often or ever) be corrupt like that.

But I measured having both checks and having only one check, and it's practically the same speed, all good.

Issue #1084, opened 2024-05-12
`KeyError` raised in `format_skeleton`  when using fuzzy matching


tomasr8:
## Overview Description

When using `format_skeleton`, a `KeyError` is raised despite using `fuzzy=True`.

## Steps to Reproduce

```python3
from datetime import datetime

from babel.dates import format_skeleton


dt = datetime(2012, 1, 1, 14, 30, 59)
format_skeleton("G", dt, locale="cs_CZ", fuzzy=True)
```

From the docs:

> fuzzy – If the skeleton is not found, allow choosing a skeleton that’s close enough to it.

The way I read it is that as long as I pass `fuzzy=True` (which is the default) a skeleton should _always_ be found and I should not need to worry about `KeyError`s in that case.

Even the example in the docs makes it seem a `KeyError` is only thrown with `fuzzy=False`:
```python3
format_skeleton('yMMd', t, fuzzy=False, locale='fi')  # yMMd is not in the Finnish locale, an error is thrown
```

If this is the expected behaviour, I think the docs should state that explicitly. If not, I would suggest adding `allow_different_fields=True` to the underlying `match_skeleton()` call.

## Additional Information
Babel version: `2.15`

tomasr8:
@akx What do you think? Should we fix this?

akx:
Yeah, I think so. My original commit for fuzzy skeletons (fuzzy skeletons, yikes what a horrible mental image) https://github.com/python-babel/babel/commit/cd70395b0f138b7f60b0866f5b2b164f1a025786 refers to [the ICU4J `getBestSkeleton` function](https://github.com/unicode-org/icu/blob/ee93218eabd72406d81f8a1c92a6b9249ceb4de9/icu4j/main/core/src/main/java/com/ibm/icu/text/DateIntervalInfo.java#L1081) as the reference implementation.

Apparently there's been at least one commit https://github.com/unicode-org/icu/commit/edaebfa64ee673ed580f464123d3afd4deffc693 that has improved that function since – we should maybe look at whether that improvement would in fact fix this?

tomasr8:
Looks like the only thing that changed is more format specifier replacements which we could add as well:

```python
    if 'k' in skeleton and not any('k' in option for option in options):
        skeleton = skeleton.replace('k', 'H')
    if 'K' in skeleton and not any('K' in option for option in options):
        skeleton = skeleton.replace('K', 'h')
    if 'a' in skeleton and not any('a' in option for option in options):
[...]

PR #1106, opened 2024-07-23
Add a mention to the docs that `format_skeleton(..., fuzzy=True)` may raise

tomasr8:
Based on the discussion here: #1084 

This should make it clearer that fuzzy matching might still fail to find a suitable pattern.


akx (approved):
LGTM!

PR #1107, opened 2024-07-25
Test on Python 3.13 beta releases

akx:


.github/workflows/ci.yml
```diff
@@ -35,6 +35,7 @@ jobs:
           - "3.11"
           - "3.12"
           - "pypy3.10"
+          - "3.13.0-beta.4"
```
hugovk, line 38: This will track the latest beta/release candidates (and also use the final release when that comes out):
```suggestion
          - "3.13-dev"
```

akx: Good point, thanks! I was just looking at the https://github.com/actions/python-versions/releases list :)

Issue #777, opened 2021-04-02
Parse settings from pyproject.toml


miigotu:
This project is the last of my dependencies that can only read from setup.cfg. the domains that I think should be used are something like:
```toml
;pyproject.toml
[ tool.babel.extract_messages ]
message_extractors="""{
    'gui': [
        ('**/js/*.min.js', 'ignore', None),
        ('**/js/*.js', 'javascript', {'input_encoding': 'utf-8'})
    ]}"""
;options
[ tool.babel.compile_catalog ]
;options
[ tool.babel.init_catalog ]
;options
[ tool.babel.update_catalog ]
;options
```

miigotu:
@akx @benselme @blagasz @kvesteri @mitsuhiko @sils 
Sorry for the ping but I'd really like to see this, and I'm willing to add a PR to make this happen. Before I waste my time, is there a policy or stopper against accepting a change pr that accomplishes this? I just don'y want to waste my time if so.

miigotu:
I'd also like to submit a patch for the f-string problem. There is not a nice way to do it, but I think we could add a hack for it (which includes a small speed expense) that depends on a config/pyproject setting or environment variable or something to enable. Possibly with a new/different import when you gettext_install. That way it would not be a breaking change, and would avoid slowing down every implementation unless they enabled the feature.

akx:
@miigotu There's no policy or stopper preventing a PR for pyproject.toml.

However: 
* care should be taken to keep Babel working even when no TOML capabilities are present,
* the format of the TOML configuration should be well-thought-out in advance. The draft in the original message here seems a bit icky with a Python literal being embedded in the TOML.

As for "the f-string problem"... I think that's quite unrelated here?

miigotu:
Yeah I messed with the spec for it a bit last night, and pulled a global options out. Toml allows multiple definitions of the same key, so 
[tool.babel.config.extractors]
type = python
match = "**/source/*.py"

tool.babel.config.extractors]
type = js
match = "**/js/*.js
```
Works, but I need to think about it more.

miigotu:
[...]

PR #1108, opened 2024-08-01
Initial support for reading mapping configuration as TOML

akx:
This PR offers to implement part of https://github.com/python-babel/babel/issues/777.

Namely, following this PR you would be able to run `pybabel extract -F babel.toml` – or even `pybabel extract -F pyproject.toml`, as we already will support `tool.babel` as well as `babel` for the namespace. (Auto-detecting a `pyproject.toml` instead of `setup.cfg` or `babel.cfg` is not implemented in this PR.)

I'm **requesting comments** on the proposed TOML format, which borrows the idea of [mappings-as-lists from Mypy's overrides configuration](https://mypy.readthedocs.io/en/latest/config_file.html#example-pyproject-toml).

```toml
[babel.extractors]
custom = "mypackage.module:myfunc"

# Python source files
[[babel.mappings]]
method = "python"
pattern = "**.py"

# Genshi templates
[[babel.mappings]]
method = "genshi"
pattern = "**/templates/**.html"
include_attrs = ""

[[babel.mappings]]
method = "genshi"
pattern = "**/templates/**.txt"
template_class = "genshi.template:TextTemplate"
encoding = "latin-1"

# Some custom extractor
[[babel.mappings]]
method = "custom"
pattern = "**/custom/*.*"
```

### DIFF
--- a/babel/messages/frontend.py
+++ b/babel/messages/frontend.py
@@ -1052,15 +1052,23 @@ def _parse_config_object(config: dict, *, filename="(unknown)"):
         if not isinstance(entry, dict):
             raise ValueError(f"{filename}: mappings[{idx}]: Expected a dictionary, got {type(entry)!r}")
         entry = entry.copy()
+
         method = entry.pop("method", None)
-        pattern = entry.pop("pattern", None)
         if not isinstance(method, str):
             raise ValueError(f"{filename}: mappings[{idx}]: 'method' must be a string, got {method!r}")
-        if not isinstance(pattern, str):
-            raise ValueError(f"{filename}: mappings[{idx}]: 'pattern' must be a string, got {pattern!r}")
         method = extractors.get(method, method)  # Map the extractor name to the callable now
-        method_map.append((pattern, method))
-        options_map[pattern] = entry
+
+        pattern = entry.pop("pattern", None)
+        if not isinstance(pattern, (list, str)):
+            raise ValueError(f"{filename}: mappings[{idx}]: 'pattern' must be a list or a string, got {pattern!r}")
+        if not isinstance(pattern, list):
+            pattern = [pattern]
+
+        for pat in pattern:
+            if not isinstance(pat, str):
+                raise ValueError(f"{filename}: mappings[{idx}]: 'pattern' elements must be strings, got {pat!r}")
+            method_map.append((pat, method))
+            options_map[pat] = entry
 
     return method_map, options_map
 
--- a/babel/messages/frontend.py
+++ b/babel/messages/frontend.py
@@ -23,7 +23,7 @@ import warnings
 from collections import OrderedDict
 from configparser import RawConfigParser
 from io import StringIO
-from typing import BinaryIO, Iterable
+from typing import BinaryIO, Iterable, Literal
 
 from babel import Locale, localedata
 from babel import __version__ as VERSION
@@ -537,10 +537,15 @@ class ExtractMessages(CommandMixin):
         if self.mapping_file:
             if self.mapping_file.endswith(".toml"):
                 with open(self.mapping_file, "rb") as fileobj:
+                    file_style = (
+                        "pyproject.toml"
+                        if os.path.basename(self.mapping_file) == "pyproject.toml"
+                        else "standalone"
+                    )
                     method_map, options_map = parse_mapping_toml(
                         fileobj,
                         filename=self.mapping_file,
-                        pyproject_toml_style=(os.path.basename(self.mapping_file) == "pyproject.toml"),
+                        style=file_style,
                     )
             else:
                 with open(self.mapping_file) as fileobj:
@@ -1078,14 +1083,13 @@ def _parse_config_object(config: dict, *, filename="(unknown)"):
 def parse_mapping_toml(
     fileobj: BinaryIO,
     filename: str = "(unknown)",
-    pyproject_toml_style: bool = False,
+    style: Literal["standalone", "pyproject.toml"] = "standalone",
 ):
     """Parse an extraction method mapping from a binary file-like object.
 
     :param fileobj: a readable binary file-like object containing the configuration TOML to parse
     :param filename: the name of the file being parsed, for error messages
-    :param pyproject_toml_style: whether the file is in the style of a `pyproject.toml` file,
-                                 i.e. whether to look for `tool.babel` instead of `babel`.
+    :param style: whether the file is in the style of a `pyproject.toml` file, i.e. whether to look for `tool.babel`.
     :see: `extract_from_directory`
     """
     try:
@@ -1097,16 +1101,18 @@ def parse_mapping_toml(
             raise ImportError("tomli or tomllib is required to parse TOML files") from ie
 
     parsed_data = tomllib.load(fileobj)
-    if pyproject_toml_style:
+    if style == "pyproject.toml":
         try:
             babel_data = parsed_data["tool"]["babel"]
         except (TypeError, KeyError) as e:
             raise ValueError(f"{filename}: No 'tool.babel' section found in file") from e
-    else:
-        try:
-            babel_data = parsed_data["babel"]
-        except (TypeError, KeyError) as e:
-            raise ValueError(f"{filename}: No 'babel' section found in file") from e
+    elif style == "standalone":
+        babel_data = parsed_data
+        if "babel" in babel_data:
+            raise ValueError(f"{filename}: 'babel' should not be present in a stand-alone configuration file")
+    else:  # pragma: no cover
+        raise ValueError(f"Unknown TOML style {style!r}")
+
     return _parse_config_object(babel_data, filename=filename)
 
 
--- a/babel/messages/frontend.py
+++ b/babel/messages/frontend.py
@@ -548,7 +548,7 @@ class ExtractMessages(CommandMixin):
                         if os.path.basename(self.mapping_file) == "pyproject.toml"
                         else "standalone"
                     )
-                    method_map, options_map = parse_mapping_toml(
+                    method_map, options_map = _parse_mapping_toml(
                         fileobj,
                         filename=self.mapping_file,
                         style=file_style,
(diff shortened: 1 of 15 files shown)


akx:
cc @zharktas, @tomasr8, @eemeli – please weigh in if you have ideas :)

akx:
Oh, and cc @miigotu too of course as the opener of #777 😄

tomasr8:
Looks great! I was just wondering how you can specify ignore paths. Would this work?
```toml
[[babel.mappings]]
method = "ignore"
pattern = "test/*"
```
Can the pattern be a list?

babel/messages/frontend.py
```diff
@@ -1053,6 +1025,68 @@ def parse_mapping(fileobj, filename=None):
...
+            import tomllib
+        except ImportError as ie:
+            raise ImportError("tomli or tomllib is required to parse TOML files") from ie
+
+    parsed_data = tomllib.load(fileobj)
+    babel_data = None
+    if "babel" in parsed_data:
+        babel_data = parsed_data["babel"]
+    elif "tool" in parsed_data and isinstance(parsed_data["tool"], dict):
+        babel_data = parsed_data["tool"].get("babel")
```
eemeli, line 1084: I find supporting both `[babel]` and `[tool.babel]` to be surprising, and not in line with usual `pyproject.toml` usage. To still support both `babel.toml` and `pyproject.toml` configs, the expected path could depend on the file name?

akx: Yeah, I guess that's a little surprising and I think I was going from where the aita is the matalin here. I referred to how `ruff` does this thing, and they do have separate code paths for dealing with `ruff.toml` and `pyproject.toml`, so I suppose we can just as well too.

That said, a `pyproject.toml` with a top-level `babel` key would probably fail [`validate-pyproject`](https://pypi.org/project/validate-pyproject/) but you're right - better not be sloppy now that we have the chance not to be.

akx:
> Looks great! I was just wondering how you can specify ignore paths. 

By not specifying them in a mapping at all... 😅 There's also the command-line `--ignore-dirs` option for wholesale directory ignoring.

But if your imagined use-case is "extract from all HTML files with `jinja2`, but never extract anything from `test*.html`", that would be a separate issue. Perfectly doable with a no-op extractor (that we don't yet have), since the list of mappings is in priority order, but that'd also work for .cfg style mapping files.

> Can the pattern be a list?

Might as well support that too!

eemeli (approved):

babel/messages/frontend.py
```diff
@@ -1053,6 +1029,85 @@ def parse_mapping(fileobj, filename=None):
...
+        raise ValueError(f"{filename}: extractors: Expected a dictionary, got {type(extractors_read)!r}")
+    for method, callable_spec in extractors_read.items():
+        if not isinstance(method, str):
+            raise ValueError(f"{filename}: extractors: Extraction method must be a string, got {method!r}")
+        if not isinstance(callable_spec, str):
+            raise ValueError(f"{filename}: extractors: Callable specification must be a string, got {callable_spec!r}")
+        extractors[method] = callable_spec
+
+
+    mappings_read = config.get("mappings") or []
```
tomasr8, line 1048: For consistency with `extractors_read` above:
```suggestion
    mappings_read = config.get("mappings", [])
```

babel/messages/frontend.py
```diff
@@ -1053,6 +1029,85 @@ def parse_mapping(fileobj, filename=None):
...
+    filename: str = "(unknown)",
+    pyproject_toml_style: bool = False,
+):
+    """Parse an extraction method mapping from a binary file-like object.
+
+    :param fileobj: a readable binary file-like object containing the configuration TOML to parse
+    :param filename: the name of the file being parsed, for error messages
+    :param pyproject_toml_style: whether the file is in the style of a `pyproject.toml` file,
+                                 i.e. whether to look for `tool.babel` instead of `babel`.
+    :see: `extract_from_directory`
```
tomasr8, line 1087: `extract_from_directory` does not seem to exist anymore :thinking:

tomasr8 (approved):
> But if your imagined use-case is "extract from all HTML files with jinja2, but never extract anything from test*.html", that would be a separate issue. Perfectly doable with a no-op extractor (that we don't yet have)

I'm a bit tired so maybe I missed something, but there already exists a noop extractor - `extract_nothing` so for example this cfg currently works as expected:

```toml
[ignore: **/*_test.py]

[python: **.py]
```

My previous question was about whether this will keep working with toml and I just confimed that it does i.e. this works:

```toml
[[babel.mappings]]
method = "ignore"
pattern = "**/*_test.py"

[[babel.mappings]]
method = "python"
pattern = "**/*.py"
```

babel/messages/frontend.py
```diff
@@ -1053,6 +1029,85 @@ def parse_mapping(fileobj, filename=None):
...
+    """
+    try:
+        import tomli as tomllib
+    except ImportError:
+        try:
+            import tomllib
```
AA-Turner, line 1093: The stdlib module should probably be preferred?
```suggestion
    try:
        import tomllib
    except ImportError:
        try:
            import tomli as tomllib
```

AA-Turner (commented):
Entirely unsolicited, but an option might be:

```toml
[extractors]
custom = "mypackage.module:myfunc"

# Python source files
[[mappings.python]]
pattern = "**.py"

# Genshi templates
[[mappings.genshi]]
pattern = "**/templates/**.html"
include_attrs = ""

[[mappings.genshi]]
pattern = "**/templates/**.txt"
template_class = "genshi.template:TextTemplate"
encoding = "latin-1"

# Some custom extractor
[[mappings."custom extractor"]]
pattern = "**/custom/*.*"
```

This makes it clearer when scanning the file which table is for which method -- something the current `.cfg` file does well at. I would also drop the ``babel.`` prefix for the ``babel.toml`` file -- it is needlessly explicit (see e.g. `.ruff.toml` which drops the full `tool.ruff` prefix).

A

akx:
> I'm a bit tired so maybe I missed something, but there already exists a noop extractor - `extract_nothing` so for example this cfg currently works as expected:

My bad, it's not you being tired. I had forgotten there's a special `if method == 'ignore':` case in `extract_from_file` that handles that (as well as the `'ignore': extract_nothing` case in `_BUILTIN_EXTRACTORS`)...

akx:
> Entirely unsolicited,

Not at all, thank you for the input! Much appreciated.

> but an option might be: ...

I'm not completely sold on that – feels like it becomes too easy to accidentally do `[mappings.python]` if you expect to only have a single mapping for `python`, and you'd then get a (possibly inscrutable) parsing type error.

> I would also drop the babel. prefix for the babel.toml file -- it is needlessly explicit (see e.g. .ruff.toml which drops the full tool.ruff prefix).

That's probably a good idea.

AA-Turner:
> I'm not completely sold on that – feels like it becomes too easy to accidentally do `[mappings.python]` if you expect to only have a single mapping for `python`, and you'd then get a (possibly inscrutable) parsing type error.

I think this is a problem regardless. For example [this PR](https://github.com/python/python-docs-theme/pull/195) adds a config file with one entry, ``[jinja2: **.html]``. Someone without knowledge of TOML's quirks might write 

```toml
[mapping]
method = "jinja2"
pattern = "**.html"
```

which looks right, but would instantly fail. Worse would be:

```toml
[babel.mappings]
method = "genshi"
pattern = "**/templates/**.html"
include_attrs = ""

[babel.mappings]
method = "genshi"
pattern = "**/templates/**.txt"
template_class = "genshi.template:TextTemplate"
encoding = "latin-1"
```

This is luckily an [error](https://toml.io/en/v1.0.0#table) in TOML, but still confusing. There should be a clear and early error message if ``isinstance(config['mapping'], dict) is True`` (or in my proposal, ``any(isinstance(k, dict) for k in config['mapping'].keys())``).

I think with the clear error message (that I posit is needed regardless of the option taken), my proposal still has merit. However, it does have drawbacks as you note.

akx:
@AA-Turner @tomasr8 Do you have the time to take another look at this? Thanks!

tomasr8 (approved):
Looks good to me :) Should we add some sample toml configs to the [docs](https://babel.pocoo.org/en/latest/messages.html#extraction-method-mapping-and-configuration) as well?

babel/messages/frontend.py
```diff
@@ -1053,6 +1039,92 @@ def parse_mapping(fileobj, filename=None):
...
+            options_map[pat] = entry
+
+    return method_map, options_map
+
+
+def parse_mapping_toml(
+    fileobj: BinaryIO,
+    filename: str = "(unknown)",
+    style: Literal["standalone", "pyproject.toml"] = "standalone",
+):
```
AA-Turner, line 1093: I understand if this is to preserve interface homogeneity, but for a new method I would suggest accepting a path (ideally `pathlib.Path` or `os.PathLike`) instead of a file object. This means that you can close the object much earlier (when `tomllib.load()` finishes) and the caller doesn't need to worry about binary-vs-text mode for opening.

The side benefit would be that the ``style`` parameter could then be dropped, as you would still have the full path within ``parse_mapping_toml()``.

AA-Turner (approved):
When I moved ``theme.conf`` to ``theme.toml`` in Sphinx, I added a conversion helper (`python -m sphinx.theming conf_to_toml <path>`) for consumers to use. Perhaps something similar could be helpful here, if so I could investigate a PR?

A

babel/messages/frontend.py
tomasr8: +1 for pathlib!

akx: Yeah, kind of makes sense (though that'd make the current tests a little harder to implement).

I'll split the difference for the time being and make this function a private API for this release (a820cf5b59363aa5fa4d4019f32b9c65620c60f7) – I doubt anyone would use it from userland anyway for now, and if they do, they will hopefully have read the warning admonition in the docstring!

akx:
Thank you for the wonderful comments and review! ❤️

PR #1113, opened 2024-08-06
Do not allow substituting alternates or drafts in derived locales

### LINKED ISSUE #1112: German translation for Switzerland for country Turkey differs from Germany
## Overview
The country name / translation for the country turkey differs for locale 'de_CH' from locale 'de_AT' or 'de_DE'. Why? In my opinion we say `Türkei` as well.

## Steps to Reproduce

from babel import Locale
_locale = Locale.parse('de_CH')
_locale.territories.get('TR')
'Türkiye'
_locale = Locale.parse('de_DE')
_locale.territories.get('TR')
'Türkei'
_locale = Locale.parse('de_AT')
_locale.territories.get('TR')
'Türkei'

akx:
For more coverage, we've allowed using alternate or draft values when no officially accepted values have been present.

However, non-global (i.e. e.g. `de_CH` locales) may have an alternate spelling for what would be an alternate in the parent locale (and that alternate hasn't been imported into the parent), and the import would have then accepted the alternate from the child locale as a non-alternate.

Fixes #1112.

### DIFF
--- a/scripts/download_import_cldr.py
+++ b/scripts/download_import_cldr.py
@@ -79,7 +79,9 @@ def main():
     subprocess.check_call([
         sys.executable,
         os.path.join(scripts_path, 'import_cldr.py'),
-        common_path])
+        common_path,
+        *sys.argv[1:],
+    ])
 
 
 if __name__ == '__main__':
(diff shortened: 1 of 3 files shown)


tomasr8 (approved):
Not an expert in this but it looks reasonable to me :)

PR #1121, opened 2024-08-21
fix: check_and_call_extract_file uses the first matching method and options, instead of the first matching method and last matching options

jpmckinney:
Otherwise, if I want to use a different method for one file AND use options, I can't do, for example:

```ini
[myextractor: path/file.py]
myoption = myvalue

[python: path/**.py]
```

The current code will match `path/file.py` to `myextractor`, but it'll use the empty options from `python`...

### DIFF
--- a/babel/messages/extract.py
+++ b/babel/messages/extract.py
@@ -276,6 +276,7 @@ def check_and_call_extract_file(
         for opattern, odict in options_map.items():
             if pathmatch(opattern, filename):
                 options = odict
+                break
         if callback:
             callback(filename, method, options)
         for message_tuple in extract_from_file(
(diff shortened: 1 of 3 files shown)


akx (changes requested):
LGTM, but could you add a test for this? Thanks!

jpmckinney:
I would love to, but it's very hard to figure out where/how to test this. Basically, I'd need some scaffolding to be able to provide a configuration file, source files and a custom extractor and test whether the output is as expected.

akx (approved):
Sorry for the delay, work-work has been taking up resources...

I took the liberty of adding a test!

PR #1130, opened 2024-09-22
Replace `str.index` with `str.find`

tomasr8:
`str.index` raises a `ValueError` when the substring is not found so there is no way to even check the return value as the code currently tries to do. Replacing it with `str.find` makes the code work as intended.


akx (approved):
Good catch!

Issue #1134, opened 2024-09-24
List index out of range error in PoFileParser._add_message


gabe-sherman:
The below code triggers a list index out of range error when provided a malformed input. This occurs in _add_message at line 235 in babel/messages/pofile.py.
```
import sys
import io
import babel.messages.pofile
import babel

def main():
    input = io.StringIO(open(sys.argv[1], "r").read())
    babel.messages.pofile.read_po(input)
    

if __name__ == "__main__":
    main()
```

### Version
babel version 2.16.0

### POC File
https://github.com/FuturesLab/POC/blob/main/babel/poc-01

### How to trigger:
`python filename.py poc-01`

### Trace report
```
Traceback (most recent call last):
  File "rep.py", line 11, in <module>
    main()
  File "rep.py", line 7, in main
    babel.messages.pofile.read_po(input)
  File "/home/gabe/extras/atheris_venv/lib/python3.8/site-packages/babel/messages/pofile.py", line 434, in read_po
    parser.parse(fileobj)
  File "/home/gabe/extras/atheris_venv/lib/python3.8/site-packages/babel/messages/pofile.py", line 361, in parse
    self._finish_current_message()
  File "/home/gabe/extras/atheris_venv/lib/python3.8/site-packages/babel/messages/pofile.py", line 250, in _finish_current_message
    self._add_message()
  File "/home/gabe/extras/atheris_venv/lib/python3.8/site-packages/babel/messages/pofile.py", line 235, in _add_message
    string = self.translations[0][1].denormalize()
IndexError: list index out of range
```

gabe-sherman:
Hey Tomas! Thanks for the response. I don't have a deep understanding of the way the PO files should be processed so a full PR may be challenging, but if I'm able to get some time I'll certainly check it out :)

tomasr8:
I think checking if `self.translations` is empty inside `_finish_current_message()` and raising a `PoFileError` if that's the case might work. In any case, feel free to ask if you need help :)

gabe-sherman:
Sounds good! Should the process exit when self.translations is empty at the beginning of _add_message to avoid any attempts to access it, or should this check only occur in the block where it's actually being accessed?

gabe-sherman:
Sounds good, thanks! Should the process exit when this is detected or should we only raise the warning and leave the exception raising up to the value of self.abort_invalid.

tomasr8:
[...]

PR #1135, opened 2024-09-25
Fix of list index out of range error in PoFileParser.add_message when translations is empty

gabe-sherman:
Addresses https://github.com/python-babel/babel/issues/1134

Previously, the `add_message` function would try to access the PoFileParser object's translations list without any checks. However, when an invalid po file is provided, this can lead to an index out of range error.

This PR adds a check to the `_finish_current_message` function and raises an `invalid_pofile` exception when the messages list is populated and translations is empty. This also adds a dummy translation to the translations list to avoid the index out of range error when an `invalid_pofile` exception is treated as a warning rather than raised.

### DIFF
--- a/babel/messages/pofile.py
+++ b/babel/messages/pofile.py
@@ -249,7 +249,7 @@ class PoFileParser:
         if self.messages:
             if not len(self.translations):
                 self.translations.append([0, _NormalizedString("")])
-                self._invalid_pofile("", self.offset, "invalid PO file provided")
+                self._invalid_pofile("", self.offset, "invalid po file provided")
             self._add_message()
 
     def _process_message_line(self, lineno, line, obsolete=False) -> None:
--- a/babel/messages/pofile.py
+++ b/babel/messages/pofile.py
@@ -249,7 +249,7 @@ class PoFileParser:
         if self.messages:
             if not len(self.translations):
                 self.translations.append([0, _NormalizedString("")])
-                self._invalid_pofile("", self.offset, "invalid po file provided")
+                self._invalid_pofile("", self.offset, f"missing msgstr for msgid: {self.messages[0].denormalize()}")
             self._add_message()
 
     def _process_message_line(self, lineno, line, obsolete=False) -> None:
--- a/babel/messages/pofile.py
+++ b/babel/messages/pofile.py
@@ -249,7 +249,7 @@ class PoFileParser:
         if self.messages:
             if not len(self.translations):
                 self.translations.append([0, _NormalizedString("")])
-                self._invalid_pofile("", self.offset, f"missing msgstr for msgid: {self.messages[0].denormalize()}")
+                self._invalid_pofile("", self.offset, f"missing msgstr for msgid: '{self.messages[0].denormalize()}'")
             self._add_message()
 
     def _process_message_line(self, lineno, line, obsolete=False) -> None:
--- a/babel/messages/pofile.py
+++ b/babel/messages/pofile.py
@@ -247,6 +247,9 @@ class PoFileParser:
 
     def _finish_current_message(self) -> None:
         if self.messages:
+            if not len(self.translations):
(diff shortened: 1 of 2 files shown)


tomasr8 (commented):
Thanks! I left two comments :)

edit: I'd also add a test for this case

gabe-sherman:
I updated the error message and added a test. Let me know if there's any other changes that should be made!

tomasr8:
Thanks! I'll take a look later today or tomorrow (ping me in case I forget)

babel/messages/pofile.py
```diff
@@ -247,6 +247,9 @@ def _add_message(self) -> None:
 
     def _finish_current_message(self) -> None:
         if self.messages:
+            if not self.translations:
+                self.translations.append([0, _NormalizedString("")])
+                self._invalid_pofile("", self.offset, f"missing msgstr for msgid: '{self.messages[0].denormalize()}'")
```
tomasr8, line 252:
```suggestion
                self._invalid_pofile("", self.offset, f"missing msgstr for msgid '{self.messages[0].denormalize()}'")
```

tests/messages/test_pofile.py
```diff
@@ -1014,3 +1014,10 @@ def test_issue_1087():
 "Language: \n"
 ''')
     assert pofile.read_po(buf).locale is None
+
+def test_issue_1134():
+    buf = StringIO('''msgid "this is an invalid po file"
+msgstr "hello world"
+msgid "msgid without str"''')
```
tomasr8, line 1021: It's IMO easier to understand what a test does when it's as short as possible:

```suggestion 
    buf = StringIO('''
msgid "foo"
"''')
```

tests/messages/test_pofile.py
```diff
@@ -1014,3 +1014,10 @@ def test_issue_1087():
...
+    with pytest.raises(pofile.PoFileError):
```
tomasr8, line 1022: `PoFileError` is raised for various reasons, I'd also assert that the exception message matches ours:

```python
with pytest.raises(pofile.PoFileError) as excinfo:
    pofile.read_po(buf, abort_invalid=True)
assert str(excinfo.value) == "..." 
```

tests/messages/test_pofile.py
```diff
@@ -1014,3 +1014,10 @@ def test_issue_1087():
...
+        pofile.read_po(buf, abort_invalid=True)
```
tomasr8, line 1023: I'd also add a test when `abort_invalid` is False and maybe one test when the input includes `msgid_plural` e.g.

```
msgid "foo"
msgid_plural "foos"
```

(Also, there's a missing newline at the end of this file ;))

gabe-sherman:
Sorry for the delay, it's been a busy couple of days. I just added a couple more test cases!

tests/messages/test_pofile.py
```diff
@@ -1014,3 +1014,32 @@ def test_issue_1087():
...
+
+    buf = StringIO('''
+msgid "foo"
+msgid_plural "foos"
+"''')
+
+    # Catalog not created, aborted with PoFileError
+    with pytest.raises(pofile.PoFileError) as e:
+        pofile.read_po(buf, abort_invalid=True)
+    assert(str(e.value)) == "missing msgstr for msgid 'foo' on 1"
```
tomasr8, line 1045: Missing newline ;)

tomasr8 (commented):
Looking good, just some test nitpicks :)

gabe-sherman:
Made those extra changes. Thanks for sticking with me as I work through everything!

tomasr8:
> Made those extra changes. Thanks for sticking with me as I work through everything!

Thank _you_ for working on the fix! It's looking good to me but I don't have commit rights in this repo so I'm pinging @akx :slightly_smiling_face:

tomasr8 (approved):

PR #1136, opened 2024-10-03
Fix #774: Allow object methods to be used as extraction-keywords

### LINKED ISSUE #774: Allow object methods to be used as extraction-keywords
I'm having trouble getting certain strings to attach that are inside non-default-keyword member functions.

My code has a level of encapsulation to it, where I sometimes store the `gettext` function inside the class as a method, like so:

```
_ = gettext(...)
class Foobar:
  def __init__(self, translator):
     self.i18n = translator
   def send(self):
    return self.i18n("Hello World")
  f = Foobar(translator=_)
  print(f.send())
  ```
  
I've added the `i18n` function to `setup.cfg`:
  ```
[extract_messages]
...
keywords = _ i18n
...
```

I've also tried adding `"self.i18n"`, and `Foobar.i18n` to keywords, but this failed to extract the string as well.

I can go through and manually wrap every single string with a "pure" `N_()` function, but the code has many strings and that would be highly verbose.

embray:
Proof of concept fix for #774.  I did this quickly using `peekable()` from the more-itertools package, but if adding a new dependency just for this is (understandably) undesirable, it could be fixed just as easily with slight improvements to the generate tokens loop (in fact I had a version of that working earlier, but it gets a bit hairier with the nested case).

This would conflict with #1127, so if this is otherwise acceptable as a feature can rebase on top of that and without using more-itertools.

### DIFF
--- a/babel/messages/extract.py
+++ b/babel/messages/extract.py
@@ -36,6 +36,8 @@ from textwrap import dedent
 from tokenize import COMMENT, NAME, OP, STRING, generate_tokens
 from typing import TYPE_CHECKING, Any
 
+from more_itertools import peekable
+
 from babel.messages._compat import find_entrypoints
 from babel.util import parse_encoding, parse_future_flags, pathmatch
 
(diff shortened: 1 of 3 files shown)


akx:
Thank you! Yeah, this makes sense, but we definitely don't want an extra dependency :)

Maybe we can vendor (a subset of) `peekable` if that makes implementing this easier. Tests should also ensure multi-dotted names such as `my.corporate.app.translations.execute()` work :)

embray:
Great, given the go-ahead I'll rework this. peekable() is nice but I don't think it's really needed either.

I like how your example pokes fun at the "Enterprise-y" nature of this ;). But I swear I had a really good use case for this and wanted to make it work.

PR #1140, opened 2024-10-19
Mark `wraptext` deprecated; use `TextWrapper` directly in `write_po`

akx:
Related to #1120.

### DIFF
--- a/babel/util.py
+++ b/babel/util.py
@@ -15,6 +15,7 @@ import datetime
 import os
 import re
 import textwrap
+import warnings
 from collections.abc import Generator, Iterable
 from typing import IO, Any, TypeVar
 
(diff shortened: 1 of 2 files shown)


babel/util.py
```diff
@@ -217,6 +218,12 @@ def wraptext(text: str, width: int = 70, initial_indent: str = '', subsequent_in
     :param subsequent_indent: string that will be prepended to all lines save
                               the first of wrapped output
     """
+    warnings.warn(
+        "babel.util.wraptext is deprecated and will be removed in a future version of Babel. "
+        "If you need this functionality, use the `babel.util.TextWrapper` class directly.",
+        PendingDeprecationWarning,
```
tomasr8, line 224: Wouldn't a normal `DeprecationWarning` be more appropriate here, since this would be a normal deprecation and not a pending deprecation?

babel/util.py
```diff
@@ -217,6 +218,12 @@ def wraptext(text: str, width: int = 70, initial_indent: str = '', subsequent_in
     :param subsequent_indent: string that will be prepended to all lines save
                               the first of wrapped output
     """
+    warnings.warn(
+        "babel.util.wraptext is deprecated and will be removed in a future version of Babel. "
```
tomasr8, line 222: To be consistent with the rest of the text:
```suggestion
        "`babel.util.wraptext` is deprecated and will be removed in a future version of Babel. "
```

tomasr8 (commented):
Two small nits, otherwise LGTM!

akx:
@tomasr8 Good comments, thanks! 😄

PR #1142, opened 2024-10-28
Fix `pybabel update --init-missing` not creating parent directory

### LINKED ISSUE #1139: option --init-missing of `pybabel update` should also create the corresponding directory
I use the option `--init-missing` of `pybabel update` to avoid redundant `pybabel init` command. However, this option only creates the *.po file, and will fail if the target directory (`[locale]/LC_MESSAGES`) does not exist yet:
```bash
creating catalog .\en_US\LC_MESSAGES\strings.po based on strings.pot
  File "<frozen runpy>", line 198, in _run_module_as_main
  File "...\.venv\Scripts\pybabel.exe\__main__.py", line 8, in <module>
  File "...\.venv\Lib\site-packages\babel\messages\frontend.py", line 999, in main
  File "...\.venv\Lib\site-packages\babel\messages\frontend.py", line 925, in run
    return cmdinst.run()
           ^^^^^^^^^^^^^
  File "...\.venv\Lib\site-packages\babel\messages\frontend.py", line 795, in run
    with open(filename, 'wb') as outfile:
         ^^^^^^^^^^^^^^^^^^^^
FileNotFoundError: [Errno 2] No such file or directory: '.\\en_US\\LC_MESSAGES\\strings.po'
```
This behavior makes the option less useful. I suggest invoke `os.makedirs` in `UpdateCatalog.finalize_options`, just like in `InitCatalog.finalize_options`:
https://github.com/python-babel/babel/blob/f91754b01cb9f32b83aeaa80b74ed10b5dfccb6a/babel/messages/frontend.py#L629-L630
but with extra `exist_ok=True` argument, as the directory may already exist for the update command.

Or it might be better to reuse the logic in `InitCatalog`.

du33169:
Replaced the redundant init logic in `UpdateCatalog.run()` with a temporary `InitCatalog` instance to ensure the parent directory of the output file will be created. (close #1139)

### DIFF
--- a/babel/messages/frontend.py
+++ b/babel/messages/frontend.py
@@ -626,8 +626,8 @@ class InitCatalog(CommandMixin):
             self.output_file = os.path.join(self.output_dir, self.locale,
                                             'LC_MESSAGES', f"{self.domain}.po")
 
-        if not os.path.exists(os.path.dirname(self.output_file)):
-            os.makedirs(os.path.dirname(self.output_file))
+        os.makedirs(os.path.dirname(self.output_file),exist_ok=True)
+
         if self.no_wrap and self.width:
             raise OptionError("'--no-wrap' and '--width' are mutually exclusive")
         if not self.no_wrap and not self.width:
--- a/babel/messages/frontend.py
+++ b/babel/messages/frontend.py
@@ -779,21 +779,17 @@ class UpdateCatalog(CommandMixin):
                 if self.check:
                     check_status[filename] = False
                     continue
-                self.log.info(
-                    'creating catalog %s based on %s', filename, self.input_file,
-                )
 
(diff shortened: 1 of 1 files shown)


du33169:
> Could we get some sort of test for this? Does the (re)use of `InitCatalog` do exactly the same thing as the old code?

The original logic replaced in `UpdateCatalog` is the same as `InitCatalog.run()`. To resolve the corresponding issue, we need the directory creation logic in `InitCatalog.finalize_options()`. I haven't investigated on what else will be done in the constructor. 

Will it be better to elevate the logic of `InitCatalog.run()` and  `InitCatalog.finalize_options()` to classmethods?
Or should we make it simple, by just adding the directory creation logic again, and leave the redundancy for now?

PR #1145, opened 2024-11-04
Use CLDR 46

tomasr8:
[CLDR 46](https://cldr.unicode.org/downloads/cldr-46) is out now, let's upgrade.

Looking at the release notes, I don't think there's anything we need to update in babel, but another pair of eyes would be nice :)

### DIFF
--- a/scripts/download_import_cldr.py
+++ b/scripts/download_import_cldr.py
@@ -9,10 +9,10 @@ import sys
 import zipfile
 from urllib.request import urlretrieve
 
-URL = 'https://unicode.org/Public/cldr/45/cldr-common-45.0.zip'
-FILENAME = 'cldr-common-45.0.zip'
+URL = 'https://unicode.org/Public/cldr/46/cldr-common-46.0.zip'
+FILENAME = 'cldr-common-46.0.zip'
 # Via https://unicode.org/Public/cldr/45/hashes/SHASUM512.txt
-FILESUM = '638123882bd29911fc9492ec152926572fec48eb6c1f5dd706aee3e59cad8be4963a334bb7a09a645dbedc3356f60ef7ac2ef7ab4ccf2c8926b547782175603c'
+FILESUM = 'e0729de0d7fa8b7db620014ed16129a14f3be2f45191626c1a26778b2aa0eabe488752ff5e39da645c2dc7e8046b25a0eeef8e9ae472d26abe05a0efc8059654'
 BLKSIZE = 131072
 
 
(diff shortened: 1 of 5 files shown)


akx:
Looking at the release notes,

> Added `alt='official'` to represent cases where an official value differs from the customary value.

may require changes in the importer (probably to ignore `"official"` in favor of the customary value).

When upgrading the CLDR, I've generally run the importer with the `-j` flag with the old CLDR so it zoops out human-readable JSON, move it aside, then do the same with the new CLDR and diff for interesting changes in languages I know in e.g. Beyond Compare or your other favorite diff tool.

> Added units: `portion-per-1e9` (aka per-billion), `night` (for hotel stays), `light-speed` (as an internal prefix for light-second, light-minute, etc.)

Could be fun to see if (some of) these are supported in some locales and add some unit formatting tests for them. 

The new emoji search stuff would also be nice to incorporate in some form (but that'd be a follow-up feature PR).

tomasr8:
> Looking at the release notes,
> 
> > Added `alt='official'` to represent cases where an official value differs from the customary value.
> 
> may require changes in the importer (probably to ignore `"official"` in favor of the customary value).

I'll try to see if there's some recommendation which value should be used in case both are present.

> When upgrading the CLDR, I've generally run the importer with the `-j` flag with the old CLDR so it zoops out human-readable JSON, move it aside, then do the same with the new CLDR and diff for interesting changes in languages I know in e.g. Beyond Compare or your other favorite diff tool.

Thanks for the tip! I'll try that :)

> > Added units: `portion-per-1e9` (aka per-billion), `night` (for hotel stays), `light-speed` (as an internal prefix for light-second, light-minute, etc.)
> 
> Could be fun to see if (some of) these are supported in some locales and add some unit formatting tests for them.

I'll check if there are any and try to add some tests

> The new emoji search stuff would also be nice to incorporate in some form (but that'd be a follow-up feature PR).

Agreed, but yeah probably as a separate issue, I didn't really dig into that too much yet

tomasr8:
The only `alt="official"` that I found is in `localeDisplayNames`:

```xml
<language type="mus">Muscogee</language>
<language type="mus" alt="official">Mvskoke</language>
```
We already ignore languages with the `alt` attribute, but I added a test for it just in case.
Based on the discussion here, I think it's safe to keep using the customary value: https://unicode-org.atlassian.net/browse/CLDR-17761

I also added tests for the new units - I didn't know where to put them, so I created a new file `test_units.py` for them.

tomasr8:
Forgot to add, in the two locales I checked with the `-j` flag, I didn't see any large differences besides the new units, some timezone updates and spelling changes.

akx (approved):
Good stuff, thank you!

PR #1146, opened 2024-11-05
Changed get_week_number() to fix a problem with the locale en_AU

RomainBarbason:
This is my first pull-request ever, so don't hesitate to give me feedback on anything

If my analysis is correct, get_week_number() made it impossible for iso-calendar to show their last week, because the week_number was fixed from 53 to 1, while it should be fixed to 52.
That's why the year wasn't updated and the result was "W1 2023", it should be "W52 2023".

Hope this helps !

Fixes [#1133 ](https://github.com/python-babel/babel/issues/1133#issue-2544396234)

### DIFF
--- a/babel/dates.py
+++ b/babel/dates.py
@@ -1659,7 +1659,7 @@ class DateTimeFormat:
             # weeks, thus the weeknumber must be 53-52=1.
             max_weeks = datetime.date(year=self.value.year, day=28, month=12).isocalendar()[1]
             if week_number > max_weeks:
-                week_number -= max_weeks
+                week_number -= 1
 
         return week_number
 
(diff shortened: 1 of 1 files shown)


tomasr8:
Thanks for the PR! Do you think you could add some tests for this fix? Somewhere here might be a good place:

https://github.com/python-babel/babel/blob/313277af993cebfbae49d385327fc7f45e32f737/tests/test_dates.py#L715-L720

akx:
Yep, as Tomas said, a test would be very much appreciated - the week numbering logic seems to be pretty finicky all in all...

RomainBarbason:
Hi ! 
Sorry for the long wait, I had a lot of work these past weeks.
I added the tests you suggested.

PR #1148, opened 2024-11-09
Simplify `read_mo` logic regarding `catalog.charset`

tomasr8:
`Catalog.charset` is set to `utf-8` when no charset is provided (which is the case here) so the `if` check is redundant.


akx (approved):
Yeah, this should be fine. If someone has somehow explicitly set `cat.charset = None` after initializing a cat, that's on them, I'd say 😄

PR #1149, opened 2024-11-09
Replace `OrderedDict` with just `dict`

tomasr8:
Since python 3.7, `dict` preserves insertion order. I think we can safely replace `OrderedDict` with `dict`.


akx (approved):
Yep, absolutely!

PR #1151, opened 2024-11-17
Check pofile string delimiters

rtobar:
This PR adds checks to the pofile parser code to validate that message strings are correctly delimited by double quotes. Keeping with the current design, an error is only raised if requested, otherwise a warning is printed, the faulty lines are corrected and parsing goes on.

I found this issue while processing a pofile used in the Spanish translation of the CPython documentation. One of our files was incorrectly written, and from all our tooling only the `msgcat`  tool of GNU's `gettext` package complained, while `babel`, `polib` and others didn't. See https://github.com/python/python-docs-es/pull/2873, https://github.com/izimobil/polib/pull/161 and https://git.afpy.org/AFPy/powrap/pulls/4 for further reference. 

While implementing this change I found that the `_NormalizedString` class not only was used to contain message lines, but also participated in the parsing process (and hid some parsing as well). I thus broke down my changes into three separate commits:
 * I first clarified the usage of the *current* `_NormalizedString` class across the codebase (see details in commit).
 * I then added the double quote delimitation check logic I wanted to add to the parser
 * Now that all strings have the same form, I more formally constrained how `_NormalizedString` behaves

Along the way I also implemented three small quality-of-life changes. They are included as the first three commits of this PR, happy to submit these separately if required:
 * Avoid re-compiling a regular expression
 * Remove a duplicate test assertion
 * Perform a better assertion in a particular test, allegedly what was intended in the first place

### DIFF
--- a/babel/messages/pofile.py
+++ b/babel/messages/pofile.py
@@ -26,6 +26,8 @@ if TYPE_CHECKING:
     from typing_extensions import Literal
 
 
+_ESCAPED_CHARACTER_PATTERN = re.compile(r'\\([\\trn"])')
+
 def unescape(string: str) -> str:
     r"""Reverse `escape` the given string.
 
@@ -46,7 +48,7 @@ def unescape(string: str) -> str:
             return '\r'
         # m is \ or "
         return m
-    return re.compile(r'\\([\\trn"])').sub(replace_escapes, string[1:-1])
+    return _ESCAPED_CHARACTER_PATTERN.sub(replace_escapes, string[1:-1])
 
 
 def denormalize(string: str) -> str:
--- a/babel/messages/pofile.py
+++ b/babel/messages/pofile.py
@@ -144,7 +144,8 @@ class _NormalizedString:
             self.append(arg)
 
     def append(self, s: str) -> None:
-        self._strs.append(s.strip())
+        assert s[0] == '"' and s[-1] == '"'
+        self._strs.append(s)
 
     def denormalize(self) -> str:
         return ''.join(map(unescape, self._strs))
--- a/babel/messages/pofile.py
+++ b/babel/messages/pofile.py
@@ -278,6 +278,17 @@ class PoFileParser:
 
         self.obsolete = obsolete
 
+        def _normalized_string(s):
+            # whole lines are already stripped on both ends
+            s = s.lstrip()
+            if s[0] != '"':
+                self._invalid_pofile(line, lineno, "String must be delimited on the left by double quotes")
+                s = '"' + s
+            if s[-1] != '"':
+                self._invalid_pofile(line, lineno, "String must be delimited on the left by double quotes")
+                s += '"'
+            return _NormalizedString(s)
+
         # The line that has the msgid is stored as the offset of the msg
         # should this be the msgctxt if it has one?
         if keyword == 'msgid':
@@ -286,20 +297,20 @@ class PoFileParser:
         if keyword in ['msgid', 'msgid_plural']:
             self.in_msgctxt = False
             self.in_msgid = True
-            self.messages.append(_NormalizedString(arg))
+            self.messages.append(_normalized_string(arg))
 
         elif keyword == 'msgstr':
             self.in_msgid = False
             self.in_msgstr = True
             if arg.startswith('['):
                 idx, msg = arg[1:].split(']', 1)
-                self.translations.append([int(idx), _NormalizedString(msg)])
+                self.translations.append([int(idx), _normalized_string(msg)])
             else:
-                self.translations.append([0, _NormalizedString(arg)])
+                self.translations.append([0, _normalized_string(arg)])
 
         elif keyword == 'msgctxt':
             self.in_msgctxt = True
-            self.context = _NormalizedString(arg)
+            self.context = _normalized_string(arg)
 
     def _process_string_continuation_line(self, line, lineno) -> None:
         if self.in_msgid:
@@ -311,6 +322,10 @@ class PoFileParser:
         else:
             self._invalid_pofile(line, lineno, "Got line starting with \" but not in msgid, msgstr or msgctxt")
             return
+        assert line[0] == '"'
+        if line[-1] != '"':
+            self._invalid_pofile(line, lineno, "String must be delimited by double quotes")
+            line += '"'
         s.append(line)
 
     def _process_comment(self, line) -> None:
--- a/babel/messages/pofile.py
+++ b/babel/messages/pofile.py
@@ -251,7 +251,7 @@ class PoFileParser:
         if self.messages:
             if not self.translations:
                 self._invalid_pofile("", self.offset, f"missing msgstr for msgid '{self.messages[0].denormalize()}'")
-                self.translations.append([0, _NormalizedString("")])
+                self.translations.append([0, _NormalizedString('""')])
             self._add_message()
 
     def _process_message_line(self, lineno, line, obsolete=False) -> None:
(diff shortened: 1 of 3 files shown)


rtobar:
Gentle ping, at least to kick off CI and check if there's any obvious mistakes to be fixed

babel/messages/pofile.py
```diff
@@ -276,6 +279,17 @@ def _process_keyword_line(self, lineno, line, obsolete=False) -> None:
...
         self.obsolete = obsolete
 
+        def _normalized_string(s):
+            # whole lines are already stripped on both ends
+            s = s.lstrip()
+            if s[0] != '"':
+                self._invalid_pofile(line, lineno, "String must be delimited on the left by double quotes")
+                s = '"' + s
+            if s[-1] != '"':
+                self._invalid_pofile(line, lineno, "String must be delimited on the left by double quotes")
```
akx, line 289: Copy-paste typo 😄 

```suggestion
                self._invalid_pofile(line, lineno, "String must be delimited on the right by double quotes")
```

babel/messages/pofile.py
```diff
@@ -249,7 +252,7 @@ def _finish_current_message(self) -> None:
         if self.messages:
             if not self.translations:
                 self._invalid_pofile("", self.offset, f"missing msgstr for msgid '{self.messages[0].denormalize()}'")
-                self.translations.append([0, _NormalizedString("")])
+                self.translations.append([0, _NormalizedString('""')])
```
akx, line 255: Wouldn't
```suggestion
                self.translations.append([0, _NormalizedString()])
```
do?

babel/messages/pofile.py
```diff
@@ -276,6 +279,17 @@ def _process_keyword_line(self, lineno, line, obsolete=False) -> None:
 
         self.obsolete = obsolete
 
+        def _normalized_string(s):
+            # whole lines are already stripped on both ends
+            s = s.lstrip()
+            if s[0] != '"':
+                self._invalid_pofile(line, lineno, "String must be delimited on the left by double quotes")
+                s = '"' + s
+            if s[-1] != '"':
```
akx, line 288: These would raise an ugly `IndexError` if `s` is falsy – can we be sure that'll never be the case?

babel/messages/pofile.py
```diff
@@ -309,6 +323,10 @@ def _process_string_continuation_line(self, line, lineno) -> None:
         else:
             self._invalid_pofile(line, lineno, "Got line starting with \" but not in msgid, msgstr or msgctxt")
             return
+        assert line[0] == '"'
+        if line[-1] != '"':
+            self._invalid_pofile(line, lineno, "String must be delimited by double quotes")
```
akx, line 328: Would it be useful to note that this is a continuation context? Also, this message wouldn't be shown if the above assertion raises (unless asserts have been disabled with `-O`...), so maybe the message could be more specific?
```suggestion
            self._invalid_pofile(line, lineno, "Continuation string must end with double quotes")
```

babel/messages/pofile.py
```diff
@@ -142,7 +144,8 @@ def __init__(self, *args: str) -> None:
             self.append(arg)
 
     def append(self, s: str) -> None:
-        self._strs.append(s.strip())
+        assert s[0] == '"' and s[-1] == '"'
```
akx, line 147: Since this looks to be a hard constraint for `_NormalizedString` to denormalize correctly (since `unescape` always strips the first and last character), this should probably not be an `assert` that could get stripped out by `-O` modes.

```suggestion
        if not (s and s[0] == '"' and s[-1] == '"'):
             raise ValueError(f"_NormalizedString components must be double quoted; got {s!r}")
```
maybe?

akx (changes requested):
Hi! Sorry for the delay and thanks for the contribution! ❤️ 

Some comments within – very nice work on the very detailed commit messages; they were a joy to read!

babel/messages/pofile.py
rtobar: Happy to change the message. I added the assert because it's an invariant (this function is allied only if the string starts with double quotes), and thus we don't need to perform the check on the first character, only the last.

babel/messages/pofile.py
rtobar: Oops!

babel/messages/pofile.py
rtobar: I guess that depends on how you want to treat this class. Since this is a non-user-facing class, I thought I'd model this requirement as an assertion the code does on its input, and that babel developers shouldn't break -- users should never see this failing, assertions enabled or disabled. In other words, if users were to see such an exception, there'd be nothing they could do about it.

Taking one step behind, I think ideally I'd rather see this class take strings with the double quotes already removed in the first place, which is way less confusing. However I think I saw the `unescape` function being used somewhere else, and I didn't want to introduce breaking behaviour. I'm also not sure whether the function is part of babel's public API (judging by the name it might be).

babel/messages/pofile.py
rtobar: Mmmm... Can't recall without looking at this in a computer and running tests, but I have vague memories of this needing a value so the `_NormalizedString` instance's `__bool__` method could return `True`

babel/messages/pofile.py
rtobar: I thought we could.

- Each line is first stripped
- Keyword lines that trigger this code have a keyword followed by a `[` or by.a space (which implies at least another character after the space, because of the previous step)
- The incoming `s` here is `arg` in the enclosing context, i.e. what comes after the keyword, so either at least a `[` or at least a space and another character; hence, never empty after lstrip
  - The only problem here is that we could eventually be handling a line literally looking like `msgstr[1]` (or similarly invalid lines with no correct content after `]`) and we would run into the issue you mention. But OTOH it looks like the parsing of msgstr plural entries is currently quite fragile anyway (e.g., if no `]` is present, or what's between is not a correct number, or nothing in between, etc...) you would get other sorts of errors.

So: would you rather keep this change as-is and deal with msgstr plurals separately to add more validations, or should I go with a slice syntax to avoid any potential index errors?

rtobar (commented):
Thanks @akx for the review, it's much appreciated.

I'm far from computers.at the moment, so I won't be pushing anything until next week, at least. In the meantime please see my replies. I'll also attend to the coverage report that was raised, I thought I had covered all bases with the tests.

Do you prefer force-pushes (I know I do) or further commits?

babel/messages/pofile.py
rtobar: It seems this was just me confusing myself with how `_NormalizedString` works. I've checked and I can safely (all tests run correctly) replace all instances of `_NormalizedString('""')` with `_NormalizedString()`, even a couple that weren't included in the initial changeset.

babel/messages/pofile.py
rtobar: OK, I've opted for adding further checks to msgstr plural entries, ensuring that by the point the message itself is being parsed we can be sure it's not empty. I've also added an `assert` just before accessing the `0` and `-1` elements to denote that the string is expected to be not empty at that point.

babel/messages/pofile.py
rtobar: Leaving as-is at the moment, let me know if you want this to change.

rtobar:
@akx thanks for the patience .I finally found some time today to work on this. I opted for adding yet more checks (and tests) for msgstr plural entries, hence ensuring their messages are as well-formed as the rest. I push-forced, and I marked all conversations as resolved, mostly to keep track of what I needed to work on, so don't necessarily interpret that as me not being open to further suggestions.

Issue #1143, opened 2024-10-31
babel does not support sphinx 8


fundawang:
## Overview Description
When building docs of babel 2.16 with sphinx 8.1.3, it failed with following messages:
```
2024-10-31 17:21:22 + pushd docs
2024-10-31 17:21:22 + make SPHINXBUILD=sphinx-build-3 BUILDDIR=/home/lkp/rpmbuild/BUILD/babel-2.16.0/built-docs html
2024-10-31 17:21:22 ~/rpmbuild/BUILD/babel-2.16.0/docs ~/rpmbuild/BUILD/babel-2.16.0
2024-10-31 17:21:22 sphinx-build-3 -b html -d /home/lkp/rpmbuild/BUILD/babel-2.16.0/built-docs/doctrees   . /home/lkp/rpmbuild/BUILD/babel-2.16.0/built-docs/html
2024-10-31 17:21:22 Running Sphinx v8.1.3
2024-10-31 17:21:22 loading translations [en]... done
2024-10-31 17:21:23 making output directory... done
2024-10-31 17:21:23 Converting `source_suffix = '.rst'` to `source_suffix = {'.rst': 'restructuredtext'}`.
2024-10-31 17:21:23 ERROR: Invalid value `None` in intersphinx_mapping['https://docs.python.org/3/']. Expected a two-element tuple or list.
2024-10-31 17:21:23 
2024-10-31 17:21:23 Configuration error:
2024-10-31 17:21:23 Invalid `intersphinx_mapping` configuration (1 error).
2024-10-31 17:21:23 make: *** [Makefile:45: html] Error 2
```

PR #1155, opened 2024-11-20
Adjust docs/conf.py to add compatibility with sphinx 8

hrnciar:
Fixes: #1143

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -252,7 +252,7 @@ texinfo_documents = [
 #texinfo_show_urls = 'footnote'
 
 intersphinx_mapping = {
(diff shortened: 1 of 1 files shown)


docs/conf.py
```diff
@@ -252,7 +252,7 @@
 #texinfo_show_urls = 'footnote'
 
 intersphinx_mapping = {
-    'https://docs.python.org/3/': None,
+    'python': ('https://docs.python.org/3/', None)
```
akx, line 255:
```suggestion
    'python': ('https://docs.python.org/3/', None),
```

akx (approved):
I wonder how legacy the old format has been...

Issue #1011, opened 2023-06-11
Feature request: implement `pybabel lint`


jeanas:
A common error that translators do is translating the format field names, while they should remain untranslated. For example: translating `"File {file} not found"` by `"Fichier {fichier} non trouvé"` instead of `"Fichier {file} non trouvé"`.

There is a script in this repository that checks for such errors: https://github.com/himdel/lint-po. However, it does its own PO file parsing that looks a bit fishy. I'd rather have this functionality in babel, with a well-tested PO parser.

Quix0r:
That would be much appreciated.

akx:
Sounds like a great idea :)

Issue #333, opened 2016-01-18
New style python-brace-format translations


sassanh:
Can we have support for python-brace-format translations?

akx:
This seems related to #35, #80.

sassanh:
It may be related but not included in those issues.

septatrix:
Any progress on this? Would you also consider adding this to babel 2 already if one would create a PR? Even GNU `xgettext` supports string tagged with `f"..."`...

stianjensen:
This would still be great to have, to be able to statically and safely lint that translations actually use format placeholders correctly. I see that a PR is open to start getting support for this now, but the project is perhaps currently lacking active maintainers to review it?

PR #1168, opened 2025-01-11
Add a new CLI command to lint PO files

tomasr8:
Partially addresses https://github.com/python-babel/babel/issues/1011

This adds a new `lint` command to babel:

```sh
$ pybabel lint --help                                
Usage: pybabel lint [options] <input-paths>

check message catalogs for common problems

Options:
  -h, --help  show this help message and exit
```

Example output:

```sh
$ pybabel lint messages.po
messages.po:20: placeholders in msgid differ from placeholders in msgstr:
	placeholders in msgid but missing in msgstr: {foo}
	placeholders in msgstr but missing in msgid: {bar}
```

The implementation is based on my team's [implementation in Indico](https://github.com/indico/indico/blob/5167d8c065e04ca468764babeffd9f5287a3af60/indico/cli/i18n.py#L666-L686).

Some notes:
- Currently only f-string placeholders are checked. I didn't implement the old %-style formatting to keep the diff smaller, but I can add it later.
- There are currently no CLI options besides the `input-paths`. We might want to consider adding a parameter to ignore a specific file/msgid but that can again be added later.

Feedback welcome!

(@akx would you mind taking a look?)

### DIFF
--- a/babel/messages/frontend.py
+++ b/babel/messages/frontend.py
@@ -20,13 +20,18 @@ import shutil
 import sys
 import tempfile
 import warnings
+from collections.abc import Generator
 from configparser import RawConfigParser
+from dataclasses import dataclass
 from io import StringIO
+from pathlib import Path
+from string import Formatter
 from typing import BinaryIO, Iterable, Literal
 
 from babel import Locale, localedata
 from babel import __version__ as VERSION
 from babel.core import UnknownLocaleError
+from babel.messages import Message
 from babel.messages.catalog import DEFAULT_HEADER, Catalog
 from babel.messages.extract import (
     DEFAULT_KEYWORDS,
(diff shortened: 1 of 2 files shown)


jun66j5:
Why the new checker is not added to babel/messages/checkers.py? It could check using `Message.check()` method if added. Also, checker for %-styling formatting is already implemented.

(Side note: Trac provides `check_catalog` command via `setup.py` to check catalogs using the method.)

tomasr8:
> Why the new checker is not added to babel/messages/checkers.py?

Yeah, that might be a good idea. I'll look into that :)

tomasr8:
Perhaps it might make more sense to start with implementing the [`python-brace-format` flag](https://github.com/python-babel/babel/issues/333)

PR #1169, opened 2025-01-12
Add `Message.python_brace_format`

tomasr8:
Relevant issue: https://github.com/python-babel/babel/issues/333

Part 1 of adding support for checking f-string parameters. This simply adds a new property on the `Message` object analogous to `python_format` which sets the corresponding flag. In a followup PR I'm going to add a checker for the f-string format (reusing my implementation from https://github.com/python-babel/babel/pull/1168)

@akx (I can't ask for a review so I'm tagging you ;))

### DIFF
--- a/babel/messages/catalog.py
+++ b/babel/messages/catalog.py
@@ -71,12 +71,16 @@ PYTHON_FORMAT = re.compile(r'''
 
 
 def _has_python_brace_format(string: str) -> bool:
+    if "{" not in string:
+        return False
     fmt = Formatter()
     try:
-        parsed = list(fmt.parse(string))
+        # `fmt.parse` returns 3-or-4-tuples of the form
+        # `(literal_text, field_name, format_spec, conversion)`;
+        # if `field_name` is set, this smells like brace format
+        return any(t[1] is not None for t in fmt.parse(string))
     except ValueError:
         return False
-    return any(True for _, field_name, *_ in parsed if field_name is not None)
 
 
 def _parse_datetime_header(value: str) -> datetime.datetime:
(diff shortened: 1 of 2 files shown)


babel/messages/catalog.py
```diff
@@ -69,6 +70,15 @@ def get_close_matches(word, possibilities, n=3, cutoff=0.6):
 ''', re.VERBOSE)
 
 
+def _has_python_brace_format(string: str) -> bool:
+    fmt = Formatter()
+    try:
+        parsed = list(fmt.parse(string))
+    except ValueError:
+        return False
+    return any(True for _, field_name, *_ in parsed if field_name is not None)
```
akx, line 79: How about
```suggestion
def _has_python_brace_format(string: str) -> bool:
    if "{" not in string:
        return False
    fmt = Formatter()
    try:
        # `fmt.parse` returns 3-or-4-tuples of the form
        # `(literal_text, field_name, format_spec, conversion)`;
        # if `field_name` is set, this smells like brace format
        return any(t[1] is not None for t in fmt.parse(string))
    except ValueError:
        return False
```
?

* fast path first
* no reason to gather the result to a list first that I can think of

tomasr8: Actually, there is a subtle difference. Consider the string `{} {`. If you try to format it, you'll get a `ValueError` because of the unclosed second brace which makes it invalid. However, `_has_python_brace_format` will return `True` because the `any` short-circuits before the parser gets to the invalid brace.

This might be a bit surprising. Perhaps we should only return `True` if the whole string is valid?

akx: Mm, fair! Maybe the `any` could be rewritten as a `not all(...)` (with a suitable comment above it to explain why the strange construction!) so we do always parse the entire thing?

tomasr8: I went for a loop in the end because I find it to be a bit easier to read compared to `not all(...)` but I'm happy to change it if you prefer!

akx (approved):
LGTM. A verbose for loop with comments is better than a hard-to-comprehend `any()`/`all()`/`not all()` expression!

PR #1173, opened 2025-01-14
Prefer LC_MONETARY when formatting currencies

akx:
> POSIX defines the [`LC_MONETARY` category](https://pubs.opengroup.org/onlinepubs/9699919799/basedefs/V1_chap07.html) to be used for monetary formatting, hence Babel could be more standards-compliant in this regard.

Closes #1051 (supersedes it).

### DIFF
--- a/babel/core.py
+++ b/babel/core.py
@@ -1089,7 +1089,10 @@ class Locale:
         return self._data['unit_display_names']
 
 
-def default_locale(category: str | None = None, aliases: Mapping[str, str] = LOCALE_ALIASES) -> str | None:
+def default_locale(
+    category: str | tuple[str, ...] | list[str] | None = None,
+    aliases: Mapping[str, str] = LOCALE_ALIASES,
+) -> str | None:
     """Returns the system default locale for a given category, based on
     environment variables.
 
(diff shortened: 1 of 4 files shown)


tomasr8:
I'll take a look later today ;)

babel/core.py
```diff
@@ -1113,11 +1116,20 @@ def default_locale(category: str | None = None, aliases: Mapping[str, str] = LOC
...
+    varnames = ('LANGUAGE', 'LC_ALL', 'LC_CTYPE', 'LANG')
+    if category:
+        if isinstance(category, str):
+            varnames = (category, *varnames)
+        elif isinstance(category, (list, tuple)):
+            varnames = (*category, *varnames)
+        else:
+            raise TypeError(f"Invalid type for category: {category!r}")
+
+    for name in varnames:
```
tomasr8, line 1132: This will technically try to look up an empty string now when using `default_locale('')` or `default_locale([''])` but I guess that's a rare-enough case?

babel/core.py
```diff
@@ -1113,11 +1116,20 @@ def default_locale(category: str | None = None, aliases: Mapping[str, str] = LOC
...
     """
-    varnames = (category, 'LANGUAGE', 'LC_ALL', 'LC_CTYPE', 'LANG')
-    for name in filter(None, varnames):
+
+    varnames = ('LANGUAGE', 'LC_ALL', 'LC_CTYPE', 'LANG')
+    if category:
+        if isinstance(category, str):
+            varnames = (category, *varnames)
+        elif isinstance(category, (list, tuple)):
+            varnames = (*category, *varnames)
```
tomasr8, line 1128: Should calls like `default_locale([None])` work without raising? (`os.getenv` raises for `None`)

tomasr8 (approved):
I just have two questions, the rest looks good to me :)

babel/core.py
akx: Added a tactical `if not name: continue` into the next loop that covers this. 👍

babel/core.py
akx: Added a tactical `if not name: continue` into the next loop that covers this. 👍

PR #1174, opened 2025-01-16
Remove u string prefix from docs

verhovsky:
I did a search and replace for `u"` and `(?<!')u'` and checked each case. I also update some strings to include the character instead of `\u` literals and I also noticed some literals in docstrings were not escaped correctly (you need to escape the backslash unless it's a `r"""` docstring, otherwise the docstring will contain the actual character instead of the literal)

### DIFF
--- a/babel/lists.py
+++ b/babel/lists.py
@@ -49,11 +49,11 @@ def format_list(
     Format the items in `lst` as a list.
 
     >>> format_list(['apples', 'oranges', 'pears'], locale='en')
-    u'apples, oranges, and pears'
+    'apples, oranges, and pears'
     >>> format_list(['apples', 'oranges', 'pears'], locale='zh')
-    u'apples\u3001oranges\u548cpears'
+    'apples\u3001oranges\u548cpears'
     >>> format_list(['omena', 'peruna', 'aplari'], style='or', locale='fi')
-    u'omena, peruna tai aplari'
+    'omena, peruna tai aplari'
 
     Not all styles are necessarily available in all locales.
     The function will attempt to fall back to replacement styles according to the rules
--- a/babel/lists.py
+++ b/babel/lists.py
@@ -51,7 +51,7 @@ def format_list(
     >>> format_list(['apples', 'oranges', 'pears'], locale='en')
     'apples, oranges, and pears'
     >>> format_list(['apples', 'oranges', 'pears'], locale='zh')
-    'apples\u3001oranges\u548cpears'
+    'apples、oranges和pears'
     >>> format_list(['omena', 'peruna', 'aplari'], style='or', locale='fi')
     'omena, peruna tai aplari'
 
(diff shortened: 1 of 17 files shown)


babel/lists.py
```diff
@@ -49,11 +49,11 @@ def format_list(
     Format the items in `lst` as a list.
 
     >>> format_list(['apples', 'oranges', 'pears'], locale='en')
-    u'apples, oranges, and pears'
+    'apples, oranges, and pears'
     >>> format_list(['apples', 'oranges', 'pears'], locale='zh')
-    u'apples\u3001oranges\u548cpears'
+    'apples、oranges和pears'
```
akx, line 54: Why do some of the changes have escaped sequences replaced with characters and others don't? 🤔 

I think I'd rather not change any escape sequences to the unescaped versions, just so it's easy to see e.g. the difference between - – — 😄

babel/support.py
```diff
@@ -94,7 +94,7 @@ def datetime(
         >>> from babel.dates import get_timezone
         >>> fmt = Format('en_US', tzinfo=get_timezone('US/Eastern'))
         >>> fmt.datetime(datetime(2007, 4, 1, 15, 30))
-        u'Apr 1, 2007, 11:30:00\u202fAM'
+        'Apr 1, 2007, 11:30:00\\u202fAM'
```
akx, line 97: Why does a single slash turn into two here? (And in other places in the diff.)

akx (changes requested):
Thank you for the PR! Generally I'm on board with this, but there are some inconsistencies and weirdness here.

babel/support.py
verhovsky: I explained in the PR, it's because this is a string within a string (a docstring).

> I also noticed some literals in docstrings were not escaped correctly (you need to escape the backslash unless it's a r""" docstring, otherwise the docstring will contain the actual character instead of the literal)

babel/lists.py
verhovsky: I also separated the 3 things I did into 3 separate commits for easier review.

verhovsky:
It would be good to add doctest to babel to check that the REPL output in the examples actually matches running the code, but I'm not going to work on that.

akx:
> It would be good to add doctest to babel to check that the REPL output in the examples actually matches what you get if you were to run the code, but I'm not going to work on that.

We do have that... https://github.com/python-babel/babel/blob/98b9562c05e5276038c27ec12c12f3e92dc027b6/conftest.py#L22-L24

verhovsky:
Turns out doctest thinks this is the same string

```py
['E d.\\u2009–\\u2009', 'E d.M.']
```

and

```py
['E d.\u2009–\u2009', 'E d.M.']
```

which is surprising because they obviously are not

```pycon
>>> 'E d.\\u2009–\\u2009' == 'E d.\u2009–\u2009'
False
```

PR #1175, opened 2025-01-16
Import `Literal` from the typing module

tomasr8:
[`typing.Literal`](https://docs.python.org/3/library/typing.html#typing.Literal) was added in 3.8. We can import it directly instead of relying on `typing_extensions`.

Edit: `TypedDict` as well

cc @akx


akx (approved):
quite literal!

PR #1176, opened 2025-01-17
Increase test coverage of the `python_format` checker

tomasr8:
There a few bugs that I found in the `python_format` checker while working on the linter that I'd like to fix (e.g. one is that it does not validate `msgstr[N]` for `N >= 2`).

Though before I start fixing anything, I want to make sure we have the checker fully covered with tests as it would be pretty easy to accidentally introduce other bugs.

cc @akx

### DIFF
--- a/babel/messages/checkers.py
+++ b/babel/messages/checkers.py
@@ -65,9 +65,6 @@ def _validate_format(format: str, alternative: str) -> None:
     arguments are not interchangeable as `alternative` may contain less
     placeholders if `format` uses named placeholders.
 
-    The behavior of this function is undefined if the string does not use
-    string formatting.
-
     If the string formatting of `alternative` is compatible to `format` the
     function returns `None`, otherwise a `TranslationError` is raised.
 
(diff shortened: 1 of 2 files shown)


babel/messages/checkers.py
```diff
@@ -65,9 +65,6 @@ def _validate_format(format: str, alternative: str) -> None:
     arguments are not interchangeable as `alternative` may contain less
     placeholders if `format` uses named placeholders.
 
-    The behavior of this function is undefined if the string does not use
```
tomasr8, line 68: I think this is wrong - `python_format` happily calls this method when `format` does not contain any format placeholders (for example when `msgid` contains them but `msgid_plural` does not, `python_format` will still call `_validate_format(msgid, msgstr)`).

tomasr8:
`+0.17%` in overall coverage :sweat_smile:

akx (approved):
LGTM! Every little (sensible) coverage increase is for the better.

PR #1187, opened 2025-02-03
Move pytest config to `pyproject.toml`

tomasr8:
I think it's better to centralize the tool configs in one file rather than have it scattered across multiple files. What do you think?


akx (approved):
LGTM. I'm also planning on switching builds to Hatch/Hatchling at some point (which will also take care of #850), so `setup.cfg` will be absolutely gone at that point.

PR #1188, opened 2025-02-04
Validate all plurals in python format checker

tomasr8:
Previously, the format checker would only validate the first 2 plurals. If a locale had more than two plurals, those would not be checked. This PR fixes that.


tomasr8:
@akx Do you think you'd have time to take a look at this at some point? I have a few followup PRs depending on this leading to the full lint CLI (though I can open them all already if you prefer?) :slightly_smiling_face:

StanFromIreland (approved):
Looks good to me :-)

akx (approved):
Lgtm! Sorry, I'd missed the ping @tomasr8 😩

PR #1189, opened 2025-02-08
Update dates.rst with current unicode.org tr35 link

clach04:


akx (approved):
LGTM!

akx:
> The pypy workflow is failing for some reason:

@tomasr8 Yeah, that's weird – it doesn't happen locally for me on that version of PyPy. #1197 works around that.

PR #1199, opened 2025-03-15
Redesign load_po to use Lex/Yacc state transitions style, to ease 

hoangduytran:
maintenance and also speed up the parsing using multiprocessing option, including debugging, abort if invalid in multiprocessing as well as in linear processing mode, plus custom hook to system's exception handling to not show the frame stacks tracing. Note the number of batches is based on the number of batches = (CPU cores  - 2) / batch division reduction. The speed for 10.8MB PO file loading just takes about 1.7seconds in multi-processing mode
Using 11 batch(es) (cpu_count: 24)
number of batches: 11, batch_size: 3222
1.623452367
Execution of run took 0:00:01.62.

### DIFF
--- a/babel/messages/block_utils.py
+++ b/babel/messages/block_utils.py
@@ -1,0 +1,57 @@
+"""
+babel.messages.block_utils
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+Utility functions for splitting PO file content into blocks.
+This module provides a function to split the content of a PO file into individual
+blocks representing separate message entries. Each block is separated by one or more
+blank lines and is returned as a tuple containing the starting line number and the block text.
+
+:copyright: (c) 2025 by the Babel Team.
+:license: BSD, see LICENSE for more details.
+"""
+
+
+def split_into_blocks(content: str):
+    """
+    Splits the given content into blocks separated by blank lines.
+
+    Each block is returned as a tuple (start_line, block_text), where:
+      - start_line: The line number in the original content where the block begins.
+      - block_text: The complete text of the block, with original line breaks preserved.
+
+    The function processes the content line by line:
+      1. It splits the content into individual lines.
+      2. It accumulates consecutive non-blank lines into a current block.
+      3. When a blank line is encountered (or at the end of the content), the current block
+         is finished and stored along with its starting line number.
+      4. The starting line number for the next block is updated.
+
+    This function is shared between the Catalog parser and the Messages parser.
+
+    :param content: The complete text content of a PO file.
+    :return: A list of tuples, each tuple containing (start_line, block_text) for a block.
+    """
+    lines = content.splitlines()  # Split the content into individual lines.
+    blocks = []  # List to store the resulting blocks.
+    current_block = []  # List to accumulate lines for the current block.
+    current_start_line = 1  # Line number where the current block starts.
+    line_number = 1  # Counter for the current line number.
+
+    for line in lines:
+        if line.strip() == "":  # Check if the line is blank.
+            if current_block:
+                # If the current block has content, join it into a single string and add to blocks.
+                blocks.append((current_start_line, "\n".join(current_block)))
+                current_block = []  # Reset the current block.
+            # Update the start line for the next block (line after the blank line).
+            current_start_line = line_number + 1
+        else:
+            # Add non-blank lines to the current block.
+            current_block.append(line)
+        line_number += 1
+
+    # If there is any remaining content in the current block, add it as the final block.
+    if current_block:
+        blocks.append((current_start_line, "\n".join(current_block)))
+    return blocks
--- a/babel/messages/catalog_parser.py
+++ b/babel/messages/catalog_parser.py
@@ -23,13 +23,14 @@ DEFAULT_CAT_STRINGS = {
     'Last-Translator': 'FULL NAME <EMAIL@ADDRESS>',
     'Language-Team': 'LANGUAGE <LL@li.org>',
     'Language': 'en',
-    'Language': 'en',
     'Plural-Forms': 'nplurals=1; plural=0;',
     'MIME-Version': '1.0',
     'Content-Type': 'text/plain; charset=utf-8',
     'Content-Transfer-Encoding': '8bit',
     'Generated-By': f'Babel {VERSION}\n',
+    'X-Generator': 'Poedit 3.5\n',
 }
+
 DEFAULT_CAT_STRING_LIST=list(DEFAULT_CAT_STRINGS.keys())
 HEADER_SEPARATOR = ':'
 
--- a/babel/messages/catalog_parser.py
+++ b/babel/messages/catalog_parser.py
@@ -6,7 +6,7 @@ Parsing of gettext PO file header blocks to construct a Babel Catalog instance.
 This module provides a function to parse only the header block (first block)
 of a pre-split PO file and extract metadata to initialize a Catalog.
 
-:copyright: (c) 2013-2024 by the Babel Team.
+:copyright: (c) 2025 by the Babel Team.
 :license: BSD, see LICENSE for more details.
 """
 
--- a/babel/messages/catalog_parser.py
+++ b/babel/messages/catalog_parser.py
@@ -6,13 +6,32 @@ Parsing of gettext PO file header blocks to construct a Babel Catalog instance.
 This module provides a function to parse only the header block (first block)
 of a pre-split PO file and extract metadata to initialize a Catalog.
 
-:copyright: (c) 2025 by the Babel Team.
+:copyright: (c) 2013-2024 by the Babel Team.
 :license: BSD, see LICENSE for more details.
 """
 
 from babel.messages import Catalog
-from babel.messages.msg_parser import process_block, parse_header_msgstr
-from babel.messages.parse_utils import get_char_set, get_int_kwarg, get_str_kwargs, get_boolean_kwarg
+from babel.messages.catalog import VERSION
+from babel.messages.msg_parser import process_block, parse_header_msgstr, printErrors
+from babel.messages.parse_utils import get_char_set
+
+DEFAULT_CAT_STRINGS = {
+    'Project-Id-Version': 'Foobar 1.0',
+    'Report-Msgid-Bugs-To': 'EMAIL@ADDRESS',
+    'POT-Creation-Date': '1990-04-01 15:30+0000',
+    'PO-Revision-Date': 'YEAR-MO-DA HO:MI+ZONE',
+    'Last-Translator': 'FULL NAME <EMAIL@ADDRESS>',
+    'Language-Team': 'LANGUAGE <LL@li.org>',
+    'Language': 'en',
+    'Language': 'en',
+    'Plural-Forms': 'nplurals=1; plural=0;',
+    'MIME-Version': '1.0',
+    'Content-Type': 'text/plain; charset=utf-8',
+    'Content-Transfer-Encoding': '8bit',
+    'Generated-By': f'Babel {VERSION}\n',
+}
+DEFAULT_CAT_STRING_LIST=list(DEFAULT_CAT_STRINGS.keys())
+HEADER_SEPARATOR = ':'
 
 def parse_catalog(blocks: list,
                   default_catalog: Catalog,
@@ -34,7 +53,16 @@ def parse_catalog(blocks: list,
 
     # Use only the first block (the header block).
     first_block_start, first_block = blocks[0]
-    header_msg = process_block(first_block, first_block_start)
+    header_msg = process_block(first_block,
+                               first_block_start,
+                               is_catalog_header=True,
+                               valid_catalog_string_list=DEFAULT_CAT_STRING_LIST,
+                               header_separator = HEADER_SEPARATOR,
+                               )
+    if not header_msg:
+        # print errors if any
+        printErrors()
+        return default_catalog
 
     # The header message's string (msgstr) is parsed into key-value pairs.
     header_text = header_msg.string
@@ -51,8 +79,7 @@ def parse_catalog(blocks: list,
 
     # Create and configure a Catalog instance.
     catalog = default_catalog
-    header_cm = [f'# {cm}' for cm in header_msg.user_comments]
-    catalog._header_comment = '\n'.join(header_cm)
+    # catalog._header_comment = '\n'.join(header_msg.user_comments)
     catalog._set_mime_headers(header_dict.items())
     # Set the catalog header comment using the user_comments from the header message.
     catalog.fuzzy = fuzzy
--- a/babel/messages/catalog_parser.py
+++ b/babel/messages/catalog_parser.py
@@ -11,28 +11,8 @@ of a pre-split PO file and extract metadata to initialize a Catalog.
 """
 
 from babel.messages import Catalog
-from babel.messages.catalog import VERSION
 from babel.messages.msg_parser import process_block, parse_header_msgstr, printErrors
-from babel.messages.parse_utils import get_char_set
-
-DEFAULT_CAT_STRINGS = {
-    'Project-Id-Version': 'Foobar 1.0',
-    'Report-Msgid-Bugs-To': 'EMAIL@ADDRESS',
-    'POT-Creation-Date': '1990-04-01 15:30+0000',
-    'PO-Revision-Date': 'YEAR-MO-DA HO:MI+ZONE',
-    'Last-Translator': 'FULL NAME <EMAIL@ADDRESS>',
-    'Language-Team': 'LANGUAGE <LL@li.org>',
-    'Language': 'en',
-    'Plural-Forms': 'nplurals=1; plural=0;',
-    'MIME-Version': '1.0',
-    'Content-Type': 'text/plain; charset=utf-8',
-    'Content-Transfer-Encoding': '8bit',
-    'Generated-By': f'Babel {VERSION}\n',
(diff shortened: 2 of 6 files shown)


akx:
Oh... At first sight, this looks like a whole lot more complex than the parser we have right now.

Could you please explain why this would be necessary, to begin with? (What prompted you to start writing this?)  What are the improvements over the current implementation (`PoFileParser`)? 

PRs are of course in general welcome, but a 1300 line PR out of the blue is a bit daunting! 😅

AA-Turner:
@akx FYI, there is discussion in https://github.com/sphinx-doc/sphinx-intl/issues/118. I'm equally confused about the motivation behind this series of PRs.

A

hoangduytran:
Thank you all for your comments and findings. On my machine, this is the code I'm using to load and write a PO file, taking from here -- https://translate.blender.org/projects/blender-ui/ui/vi/ -- click the Files->Download translation. The file is about 7.6MB (the version on my machine). The code is this: 
Screenshot 2025-03-19 at 17 26 48
I use my local pyenv, and swap in and out the old and the new when testing, this is the results I've got today, the top one, 4.68 seconds is the current load_po(), I then swapped out the old env with the new env, and reran the same test, and the code took only 3.09 seconds. Now this is without running the loading in multiprocessing mode. With the processing mode on, I got 2.26 seconds for both tasks, read and write. 

Screenshot 2025-03-19 at 17 36 55
Screenshot 2025-03-19 at 17 37 21
I am using Python 3.9.6 and running on MacOS Monterey 12.7.6. I have made some more corrections and adding a little more features to this file last few days. If you want, I can update this new code. I don't know why you have problem in running it, when I ran on my machine, it's working fine.

hoangduytran:
> I tried this locally using https://projects.blender.org/blender/blender-manual-translations/raw/branch/main/fi/LC_MESSAGES/blender_manual.po as the .po file to parse.
> 
> After making some changes (e.g. imports from a `local_logging` package I do not have, or a dependency on `sphinx_intl` (?)), this benchmarks to be about 3x slower than the current implementation on `master` on my machine, and prints out a whole lot of "Invalid continuation line for state 128" errors.
> 
> With `multiprocessing=True` passed to `load_po`, the resulting catalog is empty.
> 
> Secondly, I see there's a lot of reliance on shared global state, which is generally not a great idea – in particular, this couldn't necessarily be safely run in e.g. a web server context (if someone needed to e.g. extract messages from a PO file for, say, a translation service).
> 
> All in all, I might suggest that if you need this for some other project, you could maintain it separately – and on the other hand, now that I look at our PO file parsing code, there are things that we could optimize even without overhauling the entire thing.

Can you please tell me which os are you testing on, so I can see if I can replicate your testing environment, to see exactly what problem is. Please test with the latest source code I just update and let me know anything else needed to check for. Thank you so much. I know it's about collaborations and not individuality here. Thank you for your assistance.

hoangduytran:
Ok, I have corrected some more, using your test file (blender_manual.po) above, the result runs is this:

Linear Mode
% test.py
loaded 39271 records!
7.5229632010000005
Execution of run took 0:00:07.52.

Multiprocessing mode
% test.py
Using 11 batch(es) (cpu_count: 24)
number of batches: 11, batch_size: 5268
loaded 39271 records!
4.607690345
Execution of run took 0:00:04.61.

However, there is one issue and I haven't yet got an idea how to solve it correctly, that is the problem in the grammar of the header, the key: 

"Language-Team: Finnish <https://translate.blender.org/projects/blender-"
"manual/manual/fi/>\n"

has been split into two lines. Ideally, it must be joined with the previous line to make sense. With the current model of line by line processing, I haven't yet how to solve this and yet still maintain the correct grammar structure of the header. The common thought is to join with previous line when error occurs, assuming it's one part that has been broken down by the 'normalize' routine. but to me, only with human eyes that we could work this out clearly, whereas with computer, we're working on assumptions and that's dangerous. That's why I need to think it over, or if you have an idea how to solve this effectively but still maintain the correct logic then do please share.

# SOME DESCRIPTIVE TITLE.
# Copyright (C) : This page is licensed under a CC-BY-SA 4.0 Int. License
# This file is distributed under the same license as the Blender 2.92 Manual
# package.
# FIRST AUTHOR <EMAIL@ADDRESS>, 2020.
#
msgid ""
msgstr ""
"Project-Id-Version: Blender 2.92 Manual 2.92\n"
"Report-Msgid-Bugs-To: \n"
"POT-Creation-Date: 2025-03-17 16:53+0100\n"
"PO-Revision-Date: 2025-02-24 16:41+0000\n"
"Last-Translator: Anonymous <noreply@weblate.org>\n"
"Language: fi\n"
"Language-Team: Finnish <https://translate.blender.org/projects/blender-"
"manual/manual/fi/>\n" ******* >>>> the problem line here
"Plural-Forms: nplurals=2; plural=n != 1;\n"
"MIME-Version: 1.0\n"
"Content-Type: text/plain; charset=utf-8\n"
"Content-Transfer-Encoding: 8bit\n"
"Generated-By: Babel 2.17.0\n"

akx:
> Can you please tell me which os are you testing on, so I can see if I can replicate your testing environment, to see exactly what problem is.

Python 3.13.2, MacBook Pro (M2 Max).

The code I'm trying with is 

```
import os

from babel.messages.pofile import read_po
from babel.messages.poparse import load_po


def fun():
    method = os.environ["METHOD"]
    if method == "new":
        cat = load_po("blender_manual.po", multiprocessing=False)
    elif method == "old":
        with open("blender_manual.po", encoding="utf-8") as f:
            cat = read_po(f)
    else:
        raise NotImplementedError("...")
    assert cat
    print(cat, len(cat))


if __name__ == '__main__':
    for _x in range(3):
        fun()
```

The assert and print is there so I can verify that both methods result in the same number of messages.

With 3767daa7693e867ba7b2216d5519492df99045a2, running `env METHOD=old python readpobench_redesign_vs_orig.py` now gives me the continuation error you mentioned earlier.

Using the previous version, 4e1ee45a, with the `local_logging` and `sphinx_intl` imports removed, benchmarking with [hyperfine](https://github.com/sharkdp/hyperfine):

```
$ hyperfine --warmup 2 'env METHOD=new python readpobench_redesign_vs_orig.py' 'env METHOD=old python readpobench_redesign_vs_orig.py'
Benchmark 1: env METHOD=new python readpobench_redesign_vs_orig.py
  Time (mean ± σ):      4.204 s ±  0.029 s    [User: 4.003 s, System: 0.193 s]
  Range (min … max):    4.162 s …  4.264 s    10 runs

Benchmark 2: env METHOD=old python readpobench_redesign_vs_orig.py
  Time (mean ± σ):      1.402 s ±  0.005 s    [User: 1.327 s, System: 0.073 s]
  Range (min … max):    1.392 s …  1.413 s    10 runs

Summary
  env METHOD=old python readpobench_redesign_vs_orig.py ran
    3.00 ± 0.02 times faster than env METHOD=new python readpobench_redesign_vs_orig.py
```

akx:
> After using this model a while, for nearly two years, I just got worked up on the loading speed, and that pushed me to rewrite the `load_po()`. 

Sounds like your translation memory app could benefit from simply not loading PO files as much – as you've noticed, parsing text formats tends to be slow. I assume those .po files won't change very much for your app's use, so you could just store more optimized formats of them (and recreate the optimized format when the original has changed, naturally).

For instance, https://gist.github.com/akx/01a75f13324b7c36ad22c2b96c05df51 outputs

```
load 0.4362940788269043
opt and write 1.2938261032104492
load pickle 0.07299518585205078
``` 
on my machine, suggesting loading the catalog from a pickle file is 6 times faster than parsing the PO file. (Running that gist does require a small patch to Babel, since our 2007-vintage FixedOffsetTimezone objects don't pickle well – I'll make a PR to get rid of them.)

Even better, since your app sounds like it's basically doing database lookups, have you considered loading the catalog contents to a database? You could then do lookups like `SELECT * FROM messages WHERE msgstr LIKE '%foo%'` – I made a little proof of concept of this sort of thing, and on a totally unindexed SQLite database of 588,499 messages across 5,015 catalogs, that lookup takes about 0.1 seconds on my machine...

hoangduytran:
> import os
> 
> from babel.messages.pofile import read_po
> from babel.messages.poparse import load_po
> 
> 
> def fun():
>     method = os.environ["METHOD"]
>     if method == "new":
>         cat = load_po("blender_manual.po", multiprocessing=False)
>     elif method == "old":
>         with open("blender_manual.po", encoding="utf-8") as f:
>             cat = read_po(f)
>     else:
>         raise NotImplementedError("...")
>     assert cat
>     print(cat, len(cat))
> 
> 
> if __name__ == '__main__':
>     for _x in range(3):
>         fun()

Thank you for the test code. I'm learning new way of doing things from you. Thank you again.

hoangduytran:
> Summary
>   env METHOD=old python readpobench_redesign_vs_orig.py ran
>     3.00 ± 0.02 times faster than env METHOD=new python readpobench_redesign_vs_orig.py

This is impossible, I am using a rather old machine, 2010 MP 5,1 with a nvme drive, and I bet many people out there are still stuck with an old machines. It appears to me either there is some tricks or some buffering going on somewhere. I just use the @benchmark code like this: 

```
def benchmark(func: Callable[..., Any]) -> Callable[..., Any]:
    def format_td(seconds, digits=2):
        isec, fsec = divmod(round(seconds*10**digits), 10**digits)
        return f'{datetime.timedelta(seconds=isec)}.{fsec:0{digits}.0f}'

    def wrapper(*args: Any, **kwargs: Any) -> Any:
        start_time = perf_counter()
        value = func(*args, **kwargs)
        end_time = perf_counter()
        run_time = end_time - start_time
        # format_time = time.strftime('%H:%M:%S', time.localtime(seconds=run_time/1000.00))
        # execution_time = datetime.timedelta(seconds=run_time)
        print(run_time)
        time_str = format_td(run_time, 2)
        print(f'Execution of {func.__name__} took {time_str}.')
        return value
    return wrapper
```

and put

@benchmark
on top of my test main routine to measure the execution time.

hoangduytran:
> I assume those .po files won't change very much for your app's use

No, they are changing freqently, that's why I do not put them in redis. But your suggestion made me think about this again. I might try the redis db and see if the timing and easy of use are improved.

akx:
> No, they are changing freqently, 

How frequently? Seconds? Minutes? :)

Redis might not be a great storage here, since it's in-memory and quite simple. I would suggest SQLite, PostgreSQL, even DuckDB these days.

akx:
> > Summary
> > env METHOD=old python readpobench_redesign_vs_orig.py ran
> > 3.00 ± 0.02 times faster than env METHOD=new python readpobench_redesign_vs_orig.py
> 
> This is impossible, I am using a rather old machine, 2010 MP 5,1 with a nvme drive, and I bet many people out there are still stuck with an old machine.

I don't know what to tell you – with a script like this that adapts your `@benchmark` decorator:

```
from time import perf_counter
from typing import Any, Callable

from babel.messages.pofile import read_po
from babel.messages.poparse import load_po


def benchmark(func: Callable[..., Any]) -> Callable[..., Any]:
    def wrapper(*args: Any) -> Any:
        start_time = perf_counter()
        value = func(*args)
        end_time = perf_counter()
        run_time = end_time - start_time
        print(f'{func.__name__} {args}:'.ljust(40), f'{int(run_time*1000):5d}')
        return value

    return wrapper


@benchmark
def load_po_new(filename):
    return load_po(filename, multiprocessing=False)


@benchmark
def read_po_old(filename):
    with open(filename, encoding="utf-8") as f:
        return read_po(f)


def main():
    for _ in range(3):
        for filename in ("blender-ui-ui-vi.po", "blender_manual.po"):
            for fn in (load_po_new, read_po_old):
                fn(filename)


if __name__ == '__main__':
    main()
```

I get results like

```
(babel) ~/b/babel ((4e1ee45a) *) $ uv run --python=3.9 --no-project python bench1.py 2>/dev/null
load_po_new ('blender-ui-ui-vi.po',):      617
read_po_old ('blender-ui-ui-vi.po',):      528
load_po_new ('blender_manual.po',):       1473
read_po_old ('blender_manual.po',):        850
load_po_new ('blender-ui-ui-vi.po',):      930
read_po_old ('blender-ui-ui-vi.po',):      585
load_po_new ('blender_manual.po',):       1749
read_po_old ('blender_manual.po',):        986
load_po_new ('blender-ui-ui-vi.po',):     1129
read_po_old ('blender-ui-ui-vi.po',):      730
load_po_new ('blender_manual.po',):       1922
read_po_old ('blender_manual.po',):       1088
(babel) ~/b/babel ((4e1ee45a) *) $ uv run --python=3.13 --no-project python bench1.py 2>/dev/null
load_po_new ('blender-ui-ui-vi.po',):      451
read_po_old ('blender-ui-ui-vi.po',):      285
load_po_new ('blender_manual.po',):       1088
read_po_old ('blender_manual.po',):        535
load_po_new ('blender-ui-ui-vi.po',):      610
read_po_old ('blender-ui-ui-vi.po',):      398
load_po_new ('blender_manual.po',):       1386
read_po_old ('blender_manual.po',):        650
load_po_new ('blender-ui-ui-vi.po',):      623
read_po_old ('blender-ui-ui-vi.po',):      504
load_po_new ('blender_manual.po',):       1484
read_po_old ('blender_manual.po',):        736
```

hoangduytran:
Yes, the removal of the first part:

```
    # return poparse.load_po(filename, **kwargs)
    # pre-read to get charset
    with io.open(filename, 'rb') as f:
        cat = pofile.read_po(f)
    charset = cat.charset or 'utf-8'
```

from the 

`def load_po(filename, **kwargs):
`
in order just to get the charset from the file, in 

`sphinx_intl, `

has improved the running speed and my code is not faster, but I'm glad you're aware of the downfall of the old code. I hope this fix will benefit all users out there.

Another suggestion is to remove the normalise() or 'wrapping' the comment lines, especially the header strings, to avoid ambiguities. Do them for msgid and msgstr is OK. The result I got yesterday like this:

```
#: ../../manual/addons/3d_view/vr_scene_inspection.rst:16 ../../manual/addons/animation/copy_global_transform.rst:17 ../../manual/addons/import_export/anim_bvh.rst:17 ../../manual/addons/import_export/curve_svg.rst:17 ../../manual/addons/import_export/mesh_uv_layout.rst:13 ../../manual/addons/import_export/scene_fbx.rst:13 ../../manual/addons/import_export/scene_gltf2.rst:13 ../../manual/addons/node/node_wrangler.rst:16 ../../manual/addons/rigging/rigify/introduction.rst:53 ../../manual/addons/system/ui_translations.rst:10
#, fuzzy
msgid "Enabling Add-on"
msgstr "Edistynyt"
```
is really annoying, when the original like this:

```
#: ../../manual/addons/3d_view/vr_scene_inspection.rst:16
#: ../../manual/addons/animation/copy_global_transform.rst:17
#: ../../manual/addons/import_export/anim_bvh.rst:17
#: ../../manual/addons/import_export/curve_svg.rst:17
#: ../../manual/addons/import_export/mesh_uv_layout.rst:13
#: ../../manual/addons/import_export/scene_fbx.rst:13
#: ../../manual/addons/import_export/scene_gltf2.rst:13
#: ../../manual/addons/node/node_wrangler.rst:16
#: ../../manual/addons/rigging/rigify/introduction.rst:53
#: ../../manual/addons/system/ui_translations.rst:10
#, fuzzy
msgid "Enabling Add-on"
msgstr "Edistynyt"
```
is much cleaner.

hoangduytran:
> How frequently? Seconds? Minutes? :)
> 
> Redis might not be a great storage here, since it's in-memory and quite simple. I would suggest SQLite, PostgreSQL, even DuckDB these days.

I would say in minutes, so might be I will take a look at these DB. I cannot avoid to remember when I was working, my boss asked me to test the company's code with DB and flat files, I ran 10,000 samples over the codes to test both approaches and DB proved to be 7 times slower than flat files. This probably affected me and I tried to avoid using DB since.

hoangduytran:
I think simplify this code:


```
    def _format_comment(comment, prefix=''):
        for line in comment:
            yield f"#{prefix} {line.strip()}\n"

        # for line in comment_wrapper.wrap(comment):
        #     yield f"#{prefix} {line.strip()}\n"
```

in 

```
def generate_po(
    catalog: Catalog,
    *,
    ignore_obsolete: bool = False,
    include_lineno: bool = True,
    include_previous: bool = False,
    no_location: bool = False,
    omit_header: bool = False,
    sort_by: Literal["message", "location"] | None = None,
    width: int = 76,
) -> Iterable[str]:
```

of file 

`pofile.py`

and change this 

`yield from _format_comment(' '.join(locs), prefix=':')`

to this only:

`yield from _format_comment(locs, prefix=':')`

in the same 

`generate_po `

routine works.

I now got this back:

```
#: ../../manual/addons/3d_view/vr_scene_inspection.rst:16
#: ../../manual/addons/animation/copy_global_transform.rst:17
#: ../../manual/addons/import_export/anim_bvh.rst:17
#: ../../manual/addons/import_export/curve_svg.rst:17
#: ../../manual/addons/import_export/mesh_uv_layout.rst:13
#: ../../manual/addons/import_export/scene_fbx.rst:13
#: ../../manual/addons/import_export/scene_gltf2.rst:13
#: ../../manual/addons/node/node_wrangler.rst:16
#: ../../manual/addons/rigging/rigify/introduction.rst:53
#: ../../manual/addons/system/ui_translations.rst:10
msgid "Enabling Add-on"
msgstr "Trình Bổ Sung đã được Bật"
```

(for Vietnamese)

hoangduytran:
I think this can be closed now.

PR #1200, opened 2025-03-17
Optimizations for read_po

akx:
Inspired by #1199.

This PR improves parsing performance for a large PO file (https://projects.blender.org/blender/blender-manual-translations/raw/branch/main/fi/LC_MESSAGES/blender_manual.po) by approximately 1.4x on my machine.

On a corpus of 11,600 .po files (namely all that I had hanging around on my machine, heh) the speed-up is 1.30x:

```
$ hyperfine 'git checkout read-po-opt-nr && python readpobench_many.py' 'git checkout master && python readpobench_many.py'
Benchmark 1: git checkout read-po-opt-nr && python readpobench_many.py
  Time (mean ± σ):      8.176 s ±  0.075 s    [User: 7.683 s, System: 0.478 s]
  Range (min … max):    8.052 s …  8.311 s    10 runs

Benchmark 2: git checkout master && python readpobench_many.py
  Time (mean ± σ):     10.638 s ±  0.122 s    [User: 10.153 s, System: 0.469 s]
  Range (min … max):   10.491 s … 10.880 s    10 runs

Summary
  git checkout read-po-opt-nr && python readpobench_many.py ran
    1.30 ± 0.02 times faster than git checkout master && python readpobench_many.py
```

The only "breaking" change here is the one implemented by f988d79074b99bdb83f7096640292c92590f0384 – you can no longer pass a mixed iterable of byteses or strs into `read_po`. Typing-wise, you never could (`AnyStr` is either `bytes` or `str`, but not an union of them), but now it'll actually break.

Internally, `_NormalizedString` lost some of its lustre, but it's fine, since it is an internal helper no one should ever see.


babel/messages/catalog.py
```diff
@@ -286,9 +289,9 @@ def python_format(self) -> bool:
 
         :type:  `bool`"""
         ids = self.id
-        if not isinstance(ids, (list, tuple)):
-            ids = [ids]
-        return any(PYTHON_FORMAT.search(id) for id in ids)
+        if isinstance(ids, (list, tuple)):
+            return any(PYTHON_FORMAT.search(id) for id in ids)
```
AA-Turner, line 293:
```suggestion
            return any(filter(PYTHON_FORMAT.search, ids))
```

Though also the list comprehension inlining from [PEP 709](https://peps.python.org/pep-0709/) may be faster:

```suggestion
            return any([PYTHON_FORMAT.search(id) for id in ids])
```

babel/messages/catalog.py
```diff
@@ -301,9 +304,9 @@ def python_brace_format(self) -> bool:
 
         :type:  `bool`"""
         ids = self.id
-        if not isinstance(ids, (list, tuple)):
-            ids = [ids]
-        return any(_has_python_brace_format(id) for id in ids)
+        if isinstance(ids, (list, tuple)):
+            return any(_has_python_brace_format(id) for id in ids)
```
AA-Turner, line 308:
```suggestion
            return any(filter(_has_python_brace_format, ids))
```

akx:
@AA-Turner Thanks for the suggestions! I measured things and turns out Explicit Is Better Than Implicit – I just made them regular old for loops...

babel/util.py
```diff
@@ -38,6 +38,9 @@ def distinct(iterable: Iterable[_T]) -> Generator[_T, None, None]:
 
     :param iterable: the iterable collection providing the data
     """
+    if isinstance(iterable, (tuple, list)) and len(iterable) <= 1:
+        yield from iterable
+        return
     seen = set()
     for item in iter(iterable):
```
AA-Turner, line 45: The itertools recipe suggests 

```suggestion
for item in filterfalse(seen.__contains__, iterable):
```

Another alternative would be:

```
yield from dict.fromkeys(iterable)
```

The language now guarantees that dictionaries preserve order, but this approach is non-iterable.

akx: `dict.fromkeys(iterable)` is clever! It's 5% faster than this in a microbenchmark.

What do you mean with "but this approach is non-iterable"?

babel/messages/pofile.py
```diff
@@ -132,48 +138,14 @@ def __init__(self, message: str, catalog: Catalog, line: str, lineno: int) -> No
...
-        return self.__cmp__(other) <= 0
-
-    def __eq__(self, other: object) -> bool:
-        return self.__cmp__(other) == 0
-
-    def __ne__(self, other: object) -> bool:
-        return self.__cmp__(other) != 0
+        if not self:
+            return ""
+        return ''.join(unescape(s) for s in self)
```
AA-Turner, line 148: From memory the C loops used in `map` are still faster than Python code
```suggestion
        return ''.join(map(unescape, self))
```

AA-Turner (approved):

babel/util.py
akx: Yeah, we don't expect there to be very many items, and all of the internal uses are `list(distinct(...))` or `list(distinct(... + ...))` – TBH, I think we could micro-optimize that a bit too with an internal-use-only `_distinct_of(*iterables) -> list` function...

babel/messages/pofile.py
akx: Oh yeah, that reminds me we can do `super().__init__(map(str.strip, args))` instead of `super().__init__([s.strip() for s in args])`.

babel/util.py
akx: See https://github.com/python-babel/babel/pull/1200/commits/72e25cf81f9542fb0b502527456d84cd14ef74cc :)

tomasr8 (approved):

PR #1203, opened 2025-03-20
Use standard library `timezone` instead of `FixedOffsetTimezone`

akx:
Noticed `FixedOffsetTimezone`s didn't unpickle, and then realized we shouldn't use them at all.

### DIFF
--- a/babel/util.py
+++ b/babel/util.py
@@ -255,10 +255,20 @@ odict = dict
 
 
 class FixedOffsetTimezone(datetime.tzinfo):
-    """Fixed offset in minutes east from UTC."""
+    """
+    Fixed offset in minutes east from UTC.
+
+    DEPRECATED: Use the standard library `datetime.timezone` instead.
+    """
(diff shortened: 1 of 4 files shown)


babel/util.py
```diff
@@ -255,7 +255,12 @@ def wraptext(text: str, width: int = 70, initial_indent: str = '', subsequent_in
 
 
 class FixedOffsetTimezone(datetime.tzinfo):
-    """Fixed offset in minutes east from UTC."""
+    """
+    Fixed offset in minutes east from UTC.
+
+    DEPRECATED: Use the standard library `datetime.timezone` instead.
```
tomasr8, line 261: (Drive-by review, I'll try to review this thoroughly later) We could also add a module getter that raises a deprecation warning when this class is accessed/imported (or even when it's instantiated maybe?)

akx: Yeah, why not!

tomasr8 (approved):

PR #1204, opened 2025-03-20
Use pathlib.Path() in catalog frontend; improve test coverage

akx:
This is related to #1202 – the deeply nested `os.path.join()` would have formatted quite poorly, so it felt like a good idea to unnest it – and then I noticed the refactored codepaths weren't tested...

### DIFF
--- a/tests/messages/test_frontend.py
+++ b/tests/messages/test_frontend.py
@@ -9,6 +9,8 @@
 # This software consists of voluntary contributions made by many
 # individuals. For the exact contribution history, see the revision
 # history and logs, available at https://github.com/python-babel/babel/commits/master/.
+from __future__ import annotations
+
 import logging
 import os
 import re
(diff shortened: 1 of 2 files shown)


tests/messages/test_frontend.py
```diff
@@ -1601,6 +1602,79 @@ def test_update_catalog_boolean_args():
...
+        '--use-fuzzy',
+        '-i', str(po_file),
+        '-o', str(mo_file),
+        '-l', 'fi_FI',
+    ])
+    assert not cmdinst.run()
+    assert mo_file.exists()
+
+
+
```
tomasr8, line 1646:
```suggestion
```

tomasr8 (approved):
A bit delayed review :laughing: Anyway, we should be using pathlib wherever possible, so +1 for this :)

PR #1210, opened 2025-06-03
Use CLDR 47

tomasr8:
[CLDR 47](https://cldr.unicode.org/downloads/cldr-47) has been released.

Looking at the changes (which also include [46.1](https://cldr.unicode.org/downloads/cldr-46#461-changes)) there doesn't seem to be anything that requires changes on our side (but another pair of eyes would be helpful!)

More interesting are the upcoming changes in [v48](https://cldr.unicode.org/downloads/cldr-47#v48-advance-warnings) which might necessitate changes on our side, especially the part about `supplementalData.xml` and `supplementalMetadata.xml`:

> This will make it easier for implementations to filter out data that they don’t need, and make internal maintenance easier. This will not affect the data: just which file it is located in. Please plan to update XML and JSON parsers accordingly.

### DIFF
--- a/babel/dates.py
+++ b/babel/dates.py
@@ -1162,7 +1162,7 @@ def get_period_id(
     'midnight'
 
     >>> get_period_id(time(0), type="selection", locale="en_US")
-    'night1'
+    'morning1'
 
     :param time: The time to inspect.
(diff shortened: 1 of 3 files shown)


akx (approved):
Resolved conflict, good to go!

Issue #1218, opened 2025-07-03
get_global('zone_territories') content vs Wikipedia 'canonical' tz names


anentropic:
I noticed that `get_global('zone_territories').keys()` is missing some entries found in `pytz.common_timezones`, some of which are green 'canonical' rows on the table here https://en.wikipedia.org/wiki/List_of_tz_database_time_zones#List

for example `America/Argentina/Buenos_Aires`

or there is `Europe/Kiev` but no `Europe/Kyiv`, where Wikipedia says the latter is canonical and the former a link

Does babel want to be fixing these issues? Or it relies on some upstream source that needs fixing/updating?

Below are all the tz names from `pytz.common_timezones` that lack a key in babel zone territories:

```python
['Africa/Asmara',
 'America/Argentina/Buenos_Aires',
 'America/Argentina/Catamarca',
 'America/Argentina/Cordoba',
 'America/Argentina/Jujuy',
 'America/Argentina/Mendoza',
 'America/Atikokan',
 'America/Coyhaique',
 'America/Indiana/Indianapolis',
 'America/Kentucky/Louisville',
 'America/Nuuk',
 'Asia/Ho_Chi_Minh',
 'Asia/Kathmandu',
 'Asia/Kolkata',
 'Asia/Yangon',
 'Atlantic/Faroe',
 'Canada/Atlantic',
 'Canada/Central',
 'Canada/Eastern',
 'Canada/Mountain',
 'Canada/Newfoundland',
 'Canada/Pacific',
 'Europe/Kyiv',
 'GMT',
 'Pacific/Chuuk',
 'Pacific/Kanton',
 'Pacific/Pohnpei',
 'US/Alaska',
 'US/Arizona',
 'US/Central',
 'US/Eastern',
 'US/Hawaii',
 'US/Mountain',
 'US/Pacific',
 'UTC']
```
or if I extract the 'canonical' tz names from the Wikipedia page then the missing ones are:
```python
{'America/Argentina/Buenos_Aires',
 'America/Argentina/Catamarca',
 'America/Argentina/Cordoba',
 'America/Argentina/Jujuy',
 'America/Argentina/Mendoza',
 'America/Coyhaique',
 'America/Indiana/Indianapolis',
 'America/Kentucky/Louisville',
 'America/Nuuk',
 'Asia/Ho_Chi_Minh',
 'Asia/Kathmandu',
 'Asia/Kolkata',
 'Asia/Yangon',
 'Atlantic/Faroe',
[...]

PR #1220, opened 2025-07-08
Use canonical IANA zone names in zone_territories

akx:
Refs #1218, should fix it.

There's an `iana="..."` attribute in (at least recent) timezone XML files, but we historically never read that and instead just trusted the first alias to be the canonical tzid.


anentropic:
the results in `test_core` look good to me 👍

Issue #1219, opened 2025-07-08
Exception raised if the revision date is blank


cdfarrow:
I created a .po file with [Poedit](https://poedit.net/) and it left the PO-Revision-Date field blank in the header. pybabel (2.17.0) raises the following exception when updating or compiling the po file:

 File "C:\<>\Python312\Lib\site-packages\babel\messages\catalog.py", line 102, in _parse_datetime_header
    dt = datetime.datetime.strptime(match.group('datetime'), '%Y-%m-%d %H:%M')
         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "C:\<>\Python312\Lib\_strptime.py", line 554, in _strptime_datetime
    tt, fraction, gmtoff_fraction = _strptime(data_string, format)
                                    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "C:\<>\Python312\Lib\_strptime.py", line 333, in _strptime
    raise ValueError("time data %r does not match format %r" %
ValueError: time data '' does not match format '%Y-%m-%d %H:%M'

It would be nice to ignore the blank date, or at least, to give a graceful warning/error message.

PR #1222, opened 2025-07-26
Handle blank or malformed PO-Revision-Date in .po file headers

Aniketsy:
#1219
This PR fixes an issue where Babel would raise a valueError when encountering a blank or malformed field in .po file header.

Please provide feedback if my approach is not correct. I would love to improve and modify changes.

Thankyou!

### DIFF
--- a/babel/messages/catalog.py
+++ b/babel/messages/catalog.py
@@ -97,9 +97,18 @@ def _has_python_brace_format(string: str) -> bool:
 
 
 def _parse_datetime_header(value: str) -> datetime.datetime:
-    match = re.match(r'^(?P<datetime>.*?)(?P<tzoffset>[+-]\d{4})?$', value)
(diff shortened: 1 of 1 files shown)


Aniketsy:
hi @akx

Please review this PR and provide feedback ! If this needs any changes to fix the issue.

Aniketsy:
@tomasr8
Thankyou for the review .
I would love to work on this and fix this issue .

Thanks again .

akx:
Hi @Aniketsy, thank you for the contribution!

As Tomas also said, I don't think it's a great idea to simply ignore invalid dates and cause silent data loss.

This is somewhat similar to #1185, where "strange" plural templates would be handled differently. We may want to do something like that here too, i.e. that you could always access and manipulate the original string value if needed, and it would be passed through as-is, even when not parseable.

PR #1225, opened 2025-08-27
Implement long/narrow -> short fallback in babel.units

### LINKED ISSUE #1217: babel.units.format_unit does not fall back from long to short correctly
```py
>>> import babel.units
>>> babel.units.format_unit(1, 'duration-month', 'long', locale='et')
'1 duration-month'  # should be: '1 kuu'
```
In CLDR for estonian, duration-month is only defined for 'short' and 'narrow' lengths. Before CLDR 44, it was also defined for 'long', but since then it has been aliased to the short version. It seems that pybabel does not understand these aliases properly, cf #1075, but the fix there was only for `babel.dates`, not for `babel.units`.

This happens for other locales and other (non-duration) units too.

trillllian:
Also implemented fallback from the correct pluralization to "other", which was also missing.

Fixes #1217.

### DIFF
--- a/babel/units.py
+++ b/babel/units.py
@@ -121,7 +121,7 @@ def format_unit(
 
     .. versionadded:: 2.2.0
 
-    :param value: the value to format. If this is a string, no number formatting will be attempted.
+    :param value: the value to format. If this is a string, no number formatting will be attempted and the number is assumed to be singular for the purposes of unit formatting.
     :param measurement_unit: the code of a measurement unit.
                              Known units can be found in the CLDR Unit Validity XML file:
                              https://unicode.org/repos/cldr/tags/latest/common/validity/unit.xml
@@ -137,7 +137,6 @@ def format_unit(
     q_unit = _find_unit_pattern(measurement_unit, locale=locale)
     if not q_unit:
         raise UnknownUnitError(unit=measurement_unit, locale=locale)
-    unit_patterns = locale._data["unit_patterns"][q_unit].get(length, {})
 
     if isinstance(value, str):  # Assume the value is a preformatted singular.
         formatted_value = value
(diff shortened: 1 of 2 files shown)


trillllian:
hi, i noticed 2.18.0 was released recently. is there any chance someone could take a look at this PR? it is really not a complicated fix in my opinion, and this bug causes quite a few strings in our application to go untranslated. (well, currently it's not much of a problem since we are using babel 2.12 which has a much older CLDR, but it would be nice to upgrade.)

thanks for all the great work on babel regardless!

babel/units.py
```diff
@@ -146,8 +145,23 @@ def format_unit(
...
-        return unit_patterns[plural_form].format(formatted_value)
+    unit_patterns = locale._data["unit_patterns"][q_unit]
+
+    # We do not support `<alias>` tags at all while ingesting CLDR data,
+    # so these aliases specified in `root.xml` are hard-coded here:
+    # <unitLength type="long"><alias source="locale" path="../unitLength[@type='short']"/></unitLength>
+    # <unitLength type="narrow"><alias source="locale" path="../unitLength[@type='short']"/></unitLength>
+    lengths_to_check = [length]
+    if length in ("long", "narrow"):
+        lengths_to_check.append("short")
```
akx: Probably a tiny bit more performant to
```suggestion
    lengths_to_check = [length, "short"] if length in ("long", "narrow") else [length]
```

akx (approved):
LGTM, one tiny performance note.

Sorry reviewing this had fallen through the cracks!

trillllian:
Thank you so much!

Do you have any plans regarding when the next release will be? I have a project where I've been putting off updating Babel specifically because of this bug, so it would be nice to have a release with the fix included... I was thinking maybe you could do a 2.18.1 when you're done with all the small bugfix PRs? No rush if you have other plans though.

PR #1232, opened 2025-09-28
Rule based number formatting

blagasz:
Number spelling based on the pure Python implementation of the CLDR RBNF engine.

Currently only supports whole numbers (cardinal, ordinal, year) as fractional numbers are not well-defined and not universally implemented in CLDR.

Based on the original PR #660 with useful enhancements from @akx at #682 and additional work from the original author:
- extensive test cases are generated
- dataclass added to make parsing context more readable in code
- linting and debugging including the removal of recursion errors

Extensive test cases are added, but should be considered smoke tests until a native speaker reviews them for each language.

Based on an earlier discussion: https://github.com/python-babel/babel/pull/114 and referenced in https://github.com/python-babel/babel/issues/179

Supersedes #682


babel/numbers.py
```diff
@@ -1023,6 +1024,26 @@ def __init__(self, message: str, suggestions: list[str] | None = None) -> None:
 SPACE_CHARS_RE = re.compile('|'.join(SPACE_CHARS))
 
 
+def spell_number(number, locale=LC_NUMERIC, ruleset=None):
```
akx, line 1027: Could you add type annotations here, please?

babel/numbers.py
```diff
@@ -1023,6 +1024,26 @@ def __init__(self, message: str, suggestions: list[str] | None = None) -> None:
...
+
+    :param number: the number to format
+    :param locale: the `Locale` object or locale identifier
+    :param ruleset: the ruleset to use; defaults to regular numbers.
+    """
+    speller = RuleBasedNumberFormat.negotiate(locale)
+    return speller.format(number, ruleset=ruleset)
+
+
+def get_rbnf_rules(locale=LC_NUMERIC):
```
akx, line 1038: I'm not sure this needs to be a public API? (If it is, it also locks down the public API for RuleBasedNumberFormat.) Or... WDYT, what sort of app would consume the raw rules?

scripts/import_cldr.py
```diff
@@ -341,6 +343,15 @@ def parse_global(srcdir, sup):
                 'official_status': language.attrib.get('officialStatus'),
             }
         territory_languages[territory.attrib['type']] = languages
+
+
+    # To help the negotiation in `babel.numbers.spell_number`
```
akx, line 348: `spell_number` doesn't seem to directly read this; can you fix up the comment?

scripts/import_cldr.py
```diff
@@ -456,6 +467,16 @@ def _process_local_datas(sup, srcdir, destdir, force=False, dump_json=False):
                 f"{unsupported_number_systems_string}",
             )
 
+        # there will be no rbnf rules for all locales
+        # there could be a separate iteration for rbnf rule files
```
akx, line 471: I'm not sure I'm totally following this comment?

babel/rbnf.py
```diff
@@ -0,0 +1,814 @@
...
+# Dev notes
+#
+# Reloading cldr:
+# python ./scripts/import_cldr.py ./cldr/cldr-core-35.1/common/ -f
+#
+# Tokenization is inspired by Ka-Ping Yee's tokenize library
+
+# Undocumented syntax (←%rule-name←←)
+# Trac ticket filed for CLDR update PL rbnf
+#     http://unicode.org/cldr/trac/ticket/10544
```
akx, line 26: https://unicode-org.atlassian.net/browse/CLDR-10544 is saying this was resolved in https://unicode-org.atlassian.net/browse/CLDR-8909 and rules are flat now - can you check if these comments are still relevant?

babel/rbnf.py
```diff
@@ -0,0 +1,814 @@
...
+            SUBSTITUTION_TOKEN: self.SUBSTITUTION,
+        }[typ]
+
+
+def parse_rbnf_rules(data, tree):
+    """
+    Parse rules based on:
+    http://www.unicode.org/reports/tr35/tr35-47/tr35-numbers.html#Rule-Based_Number_Formatting
+    """
+    rbnf_rules = data.setdefault('rbnf_rules', {})
```
akx, line 797:
```suggestion
    rbnf_rules: dict[str, Ruleset] = data.setdefault('rbnf_rules', {})
```
to make it clear to the reader what's getting assigned in there.

babel/rbnf.py
```diff
@@ -0,0 +1,814 @@
...
+
+    # ElementTree.dump(tree)
+
```
akx, line 800:
```suggestion
```

babel/rbnf.py
```diff
@@ -0,0 +1,814 @@
...
+IMPROPER_FRACTION_RULE = 'x.x'
+PROPER_FRACTION_RULE = '0.x'
+MASTER_RULE = 'x.0'
+INFINITY_RULE = 'Inf'
+NOT_A_NUMBER_RULE = 'NaN'
+SPECIAL_FRACTION_RULE = 'x,x'  # there are other options but not existent in CLDR
+
+
+# locale.number_symbols['decimal']
+# normal rule means a number is specified
```
akx, line 85: Dead code/comment?

babel/rbnf.py
```diff
@@ -0,0 +1,814 @@
...
+
+class RuleMalformed(RBNFError):
+    pass
+
+
+class RulesetSubstitutionWarning(UserWarning):
+    pass
+
+
+TokenInfo = collections.namedtuple('TokenInfo', 'type reference optional')
```
akx, line 112: This could be

```suggestion
TokenInfo = collections.namedtuple('TokenInfo', ('type', 'reference', 'optional'))
```
to be less stringly typed, or better:


```suggestion
class TokenInfo(typing.NamedTuple):
   type: int
   reference: ...  # (tuple[int, str] | None?)
   optional: ...
```
(please fill in the right types :) )

babel/rbnf.py
```diff
@@ -0,0 +1,814 @@
...
+    Some of the tokens are referencing other rulesets by name.
+    This information is stored in the token along with the type
+    of reference.
+
+    """
+    # remove unnecessary syntax (only used in the non-xml form)
+    if text.endswith(";"):
+        text = text[:-1]
+    if text.startswith("'"):
+        text = text[1:]
```
akx, line 139:
```suggestion
    text = text.removesuffix(";").removeprefix("'")
```

babel/rbnf.py
```diff
@@ -0,0 +1,814 @@
...
+
+
+class RulesetSubstitutionWarning(UserWarning):
+    pass
+
+
+TokenInfo = collections.namedtuple('TokenInfo', 'type reference optional')
+
+
+def tokenize(text):
```
akx, line 115: AIUI, the tokenization, etc. code is only used during import time? It would maybe be nice if it lived outside the library proper, or at least in a separate module (`babel/rbnf/parser.py`?), to make things a little lighter to import (and easier to understand).

babel/rbnf.py
```diff
@@ -0,0 +1,814 @@
...
+    brackets. From regular expressions for the beginning and
+    end of the optional section no tokens are generated.
+    Instead, all the tokens inside the optional section are
+    flagged as optional.
+
+    Some of the tokens are referencing other rulesets by name.
+    This information is stored in the token along with the type
+    of reference.
+
+    """
```
akx, line 134: This reads more like a comment than a docstring?

babel/rbnf.py
```diff
@@ -0,0 +1,814 @@
...
+                stop = False
+                text = text[match.end():]
+                if tok == OPT_START:
+                    optional = True
+                elif tok == OPT_END:
+                    optional = False
+                else:
+                    token = _gen_token(tok, match, optional)
+                    if token:
+                        yield token
```
akx, line 157: Walrusable!
```suggestion
                    if token := _gen_token(tok, match, optional):
                        yield token
```

babel/rbnf.py
```diff
@@ -0,0 +1,814 @@
...
+                break  # always start searching with the first regex
+        if stop:
+            raise ValueError(f"Unable to tokentize '{text}'")
```
akx, line 160: Typo, and could just use the repr (so newlines and whatnot don't break the error):
```suggestion
            raise ValueError(f"Unable to tokenize {text!r}")
```

babel/rbnf.py
akx, line 115:
```suggestion
def tokenize(text: str) -> Iterable[TokenInfo]:
```
I guess?

babel/rbnf.py
```diff
@@ -0,0 +1,814 @@
...
+
+
+def _gen_token(tok, match, optional):
```
akx, line 163:
```suggestion
def _gen_token(tok: int, match: re.Match, optional: bool) -> TokenInfo | None:
```
(though `tok` could maybe be `token_type`)

babel/rbnf.py
```diff
@@ -0,0 +1,814 @@
...
+    # remove this if CLDR is updated based on ticket
+    # http://unicode.org/cldr/trac/ticket/10544
+    if tok == INTEGRAL_TOKEN and match.group(2) == '←':
+        warnings.warn('Unsupported syntax ←...←←', SyntaxWarning, stacklevel=2)
```
akx, line 167: Can you check if this is still relevant?

babel/rbnf.py
```diff
@@ -0,0 +1,814 @@
...
+    if tok == PLURAL_TOKEN:
+        _, plurals = match.group(1).split(',')
+        return TokenInfo(tok, dict(plural_rule_parser.findall(plurals)), optional)
+
+    if tok == PREVIOUS_TOKEN:
+        return TokenInfo(tok, None, optional)
+
+    if tok == TEXT_TOKEN:
+        return TokenInfo(tok, match.group(0), optional)
+
```
akx, line 183: Maybe an explicit `return None` here? When do we expect to not return a TokenInfo?

babel/rbnf.py
```diff
@@ -0,0 +1,814 @@
...
+
+def _parse_reference(string):
```
akx, line 185:
```suggestion
def _parse_reference(string: str) -> tuple[int, str]:
```

babel/rbnf.py
```diff
@@ -0,0 +1,814 @@
...
+from babel.core import Locale, get_global
+
+TEXT_TOKEN = 1
+INTEGRAL_TOKEN = 2
+REMAINDER_TOKEN = 3
+PREVIOUS_TOKEN = 4
+SUBSTITUTION_TOKEN = 5
+PLURAL_TOKEN = 6
+OPT_START = 7
+OPT_END = 8
```
akx, line 49: Would it make sense for these to be a `TokenType` [`IntEnum`](https://docs.python.org/3.14/library/enum.html#enum.IntEnum)?

babel/rbnf.py
```diff
@@ -0,0 +1,814 @@
...
+    ]
+]
+
+plural_rule_parser = re.compile(r"([^\{\}]+)\{([^\{\}]+)\}")
+
+INTERNAL_REF = 1
+PRIVATE_REF = 2
+PUBLIC_REF = 3
+# PLURAL_REF = 4  # not in use as plural is handled using separate token
+DECIMAL_REF = 5
```
akx, line 71: Would it make sense for these to be a `RefType` [`IntEnum`](https://docs.python.org/3.14/library/enum.html#enum.IntEnum)?

babel/rbnf.py
```diff
@@ -0,0 +1,814 @@
...
+    if string.startswith('%'):
+        return PUBLIC_REF, string[1:]
+    if string[0] in '0#':
+        return DECIMAL_REF, string
+    warnings.warn(f'Reference parsing error: {string}', SyntaxWarning, stacklevel=3)
+    return INTERNAL_REF, ""  # defaults to this
+
+
+def compute_divisor(value, radix):
+    # compute the highest exponent of radix less than or equal to the rule's base
```
akx, line 199: "the rule's base"? What's the rule here?

babel/rbnf.py
```diff
@@ -0,0 +1,814 @@
...
+    ctx = decimal.Context(prec=20)
+    if isinstance(value, int):
+        if value == 0:
+            return 1
+        exp = decimal.Decimal(value).ln(ctx) / decimal.Decimal(radix).ln(ctx)
+        return radix ** math.floor(exp)
+    else:
+        return None
```
akx, line 207: Maybe
```suggestion
    if not value:
        return 1
    if not isinstance(value, int):
        return None
    ctx = decimal.Context(prec=20)
    exp = decimal.Decimal(value).ln(ctx) / decimal.Decimal(radix).ln(ctx)
    return radix ** math.floor(exp)
```
so we don't construct a context just to throw it away next?

babel/rbnf.py
```diff
@@ -0,0 +1,814 @@
...
+
+        -   You can't have an empty substitution descriptor with
+            a == substitution.
+        -   If you omit the substitution descriptor in a >> substitution
+            in a fraction rule, format the result one digit at a time
+            using the rule set containing the current rule.
+        -   If you omit the substitution descriptor in a << substitution
+            in a rule in a fraction rule set, format the result using
+            the default rule set for this formatter.
+    """
```
akx, line 236: This also looks like it should be a comment, not a docstring describing the class?

babel/rbnf.py
```diff
@@ -0,0 +1,814 @@
...
+    group_types = ('SpelloutRules', 'OrdinalRules', 'NumberingSystemRules')
+    # currently only SpelloutRules is used
```
akx, line 238: This is never read? If so,


```suggestion
```

babel/rbnf.py
```diff
@@ -0,0 +1,814 @@
...
+
+    def __init__(self, locale, group='SpelloutRules'):
+        self._locale = locale
+        self._group = group
+        self.rulesets = self._locale._data['rbnf_rules'][self._group]
+        self.plural_rule = self._locale._data['plural_form']
```
akx, line 244: Let's not expose `group` if we don't support the rest in this version anyway.

```suggestion
    def __init__(self, locale: Locale) -> None:
        self._locale = locale
        self.rulesets = self._locale._data['rbnf_rules']['SpelloutRules']
        self.plural_rule = self._locale._data['plural_form']
```

babel/rbnf.py
```diff
@@ -0,0 +1,814 @@
...
+
+    @property
+    def available_rulesets(self):
+        """list available public rulesets"""
+        return [r.name for r in self.rulesets if not r.private]
```
akx, line 249: Should this be a public API, or would `_available_ruleset_names` be a better name?

babel/rbnf.py
```diff
@@ -0,0 +1,814 @@
...
+
+    def _find_matching_ruleset(self, prefix):
```
akx, line 251:
```suggestion
    def _find_matching_ruleset(self, prefix: str) -> tuple[str | None, bool]:
```

... and a docstring explaining the semantics of the return value? I guess it's "name, exact" but I can't be sure.

Also, `(None, False)` as the fallback seems ugly - maybe this should just internally raise a `RulesetNotFound` if there is no match, so the caller wouldn't need to, and the rtype would be `tuple[str, bool]`?

babel/rbnf.py
```diff
@@ -0,0 +1,814 @@
...
+        if prefix in available_rulesets:
+            return (prefix, True)
+        # sorting here avoids use of more specific ("spellout-ordinal-sinokorean-count")
+        # rulesets when a shorter one might be available.
+        for ruleset in sorted(available_rulesets):
+            if ruleset.startswith(prefix):
+                return (ruleset, False)
+        return (None, False)
+
+    def match_ruleset(self, ruleset):
```
akx, line 262:
```suggestion
    def match_ruleset(self, ruleset: str) -> Ruleset:
```

babel/rbnf.py
```diff
@@ -0,0 +1,814 @@
...
+        if ruleset == "year":
+            ruleset = "spellout-numbering-year"
+        elif ruleset == "ordinal":
+            ruleset, exact_match = self._find_matching_ruleset("spellout-ordinal")
+            if not ruleset:
+                raise RulesetNotFound(f"No ordinal ruleset is available for {self._locale}")
+            if not exact_match:
+                warnings.warn(f"Using non-specific ordinal ruleset {ruleset}", RulesetSubstitutionWarning, stacklevel=2)
+        if not ruleset.startswith("spellout-"):
+            ruleset = "spellout-" + ruleset
```
akx, line 275:
```suggestion
            ruleset = f"spellout-{ruleset}"
```

babel/rbnf.py
```diff
@@ -0,0 +1,814 @@
...
+        return ruleset_obj
+
+    def format(self, number, ruleset=None):
+        """Format a number (int/float/decimal) with spelling rules.
+
+        Ruleset may be an actual ruleset name for the locale,
+        or one of the aliases "year" or "ordinal".
+        """
+        if not ruleset:
+            ruleset = "spellout-numbering"
```
akx, line 291:
```suggestion
    def format(self, number: int | float | decimal.Decimal, ruleset: str = "spellout-numbering") -> str:
        """Format a number (int/float/decimal) with spelling rules.

        Ruleset may be an actual ruleset name for the locale,
        or one of the aliases "year" or "ordinal".
        """
```

babel/rbnf.py
```diff
@@ -0,0 +1,814 @@
...
+
+        try:
+            return ruleset.apply(number, self)
+        except RecursionError as e:
+            raise RBNFError(f"Infinite recursion formatting {number} with {ruleset.name}, potentially malformed ruleset!") from e
+
+    def get_ruleset(self, name):
+        for r in self.rulesets:
+            if r.name == name:
+                return r
```
akx, line 303: Is there a reason `rulesets` isn't a dict keyed by name (in the parsed data)? 🤔 

Feels like this is doing a whole lot of O(n)...

babel/rbnf.py
```diff
@@ -0,0 +1,814 @@
...
+            ruleset = "spellout-numbering"
+
+        ruleset = self.match_ruleset(ruleset)
+
+        try:
+            return ruleset.apply(number, self)
+        except RecursionError as e:
+            raise RBNFError(f"Infinite recursion formatting {number} with {ruleset.name}, potentially malformed ruleset!") from e
+
+    def get_ruleset(self, name):
```
akx, line 300:
```suggestion
    def get_ruleset(self, name: str) -> Ruleset | None:
```

... but it would be nicer if this just raised an exception instead of returning None if there is no match? That would come for free if `rulesets` was a dict (see comment below); this function wouldn't really even be needed if you would just `self.rulesets[name]`.

babel/rbnf.py
```diff
@@ -0,0 +1,814 @@
...
+        for r in self.rulesets:
+            if r.name == name:
+                return r
+
+    @classmethod
+    def negotiate(cls, locale):
```
akx, line 306:
```suggestion
    def negotiate(cls, locale: str | Locale) -> Self:
```
I guess?

babel/rbnf.py
```diff
@@ -0,0 +1,814 @@
...
+        """
+        Negotiate proper RBNF rules based on global data item `rbnf_locales`
+        Caching is not necessary the Locale object does that pretty well
```
akx, line 309: Reads like a comment, not a docstring.


```suggestion
        Find the RBNF rules for the given locale.
```
or something would be more appropriate, I think.

babel/rbnf.py
```diff
@@ -0,0 +1,814 @@
...
+        """
+        loc = Locale.negotiate([str(Locale.parse(locale))], get_global('rbnf_locales'))
+        if not loc:
+            raise RulesetNotFound(f"No RBNF rules available for {locale}")
+        return cls(loc)
+
+
+class Ruleset:
+    """
```
akx, line 318: This should also be a comment, not a docstring, please.

babel/rbnf.py
```diff
@@ -0,0 +1,814 @@
...
+            of the base value for the specified locale, which is
+            normally equivalent to the ←← value. This uses the ordinal
+            plural rules from PluralFormat. All strings used in the
+            plural format are treated as the same base value for parsing.
+
+    INFINITY_RULE = 'Inf'
+
+    NOT_A_NUMBER_RULE = 'NaN'
+
+    SPECIAL_FRACTION_RULE = 'x,x'  # there are other options but not existent in CLDR
```
akx, line 478:
```suggestion
```
These seem to mirror the global constants?

babel/rbnf.py
```diff
@@ -0,0 +1,814 @@
...
+    """
+
+    __slots__ = ("name", "private", "rules")
+
+    def __init__(self, name, private=False):
```
akx, line 483:
```suggestion
    def __init__(self, name: str, private: bool = False) -> None:
```

babel/rbnf.py
```diff
@@ -0,0 +1,814 @@
...
+            ruleset=self,
+            fractional=fractional,
+            SUBSTITUTION=number,
+        )
+        integral, remainder = divmod(number, 1)
+
+        # fractional rule (ruleset in fractional processing)
+        # the value should always be between 0 and 1
+        # not yet tested it needs clarification
+        # CURRENTLY ONLY WHOLE NUMBERS ARE AVAILABLE
```
akx, line 504: Does this mean that trying to format a number with a decimal part only formats the integer part? If so, we should probably instead refuse to format such a number at all instead of silently returning the wrong spelling.

babel/rbnf.py
```diff
@@ -0,0 +1,814 @@
...
+    """
+
+    __slots__ = ("name", "private", "rules")
+
+    def __init__(self, name, private=False):
+        self.name = name
+        self.private = private
+        self.rules = []
+
+    def apply(self, raw_number, parent, fractional=False, index=None):
```
akx, line 488:
```suggestion
    def apply(self, raw_number: ..., parent: ..., fractional: bool = False, index: ... = None) -> ...:
```
(please fill in the blanks)

babel/rbnf.py
```diff
@@ -0,0 +1,814 @@
...
+        # not yet tested it needs clarification
+        # CURRENTLY ONLY WHOLE NUMBERS ARE AVAILABLE
+        # if fractional:
+        #     index = self.get_rule_fractional(remainder)
+        #     if index is None:
+        #         raise RuleNotFound(f"rule for fractional processing of {remainder}")
+        #     rule = self.rules[index]
+        #     context.INTEGRAL = rule.value * remainder  # here remainder == number
+        #     context.omit_optional = (rule.value * number == 1)
+        #     return rule.apply(number, context)
```
akx, line 512: Commented-out code should probably go.


```suggestion
```

babel/rbnf.py
```diff
@@ -0,0 +1,814 @@
...
+        #             if rule is None:
+        #                 raise RuleNotFound(f"proper fraction rule ({PROPER_FRACTION_RULE})")
+
+        #         else:
+        #             rule = self.get_rule_special(IMPROPER_FRACTION_RULE)
+        #             if rule is None:
+        #                 raise RuleNotFound(f"improper fraction rule ({IMPROPER_FRACTION_RULE})")
+        #             context.omit_optional = 0 < number < 1  # between 0 and 1
+
+        #     return rule.apply(number, context)
```
akx, line 545: Commented-out code should go.
```suggestion
```

babel/rbnf.py
```diff
@@ -0,0 +1,814 @@
...
+class ParsingContext:
+    speller: RuleBasedNumberFormat = field(repr=False)
+    ruleset: Ruleset = field(repr=False)
+    fractional: bool = False
+    omit_optional: bool = False  # no default value is defined in the spec
+    remainder_as_fractional: bool = False
+    SUBSTITUTION: decimal.Decimal = None
+    INTEGRAL: decimal.Decimal = None
+    REMAINDER: decimal.Decimal = None
+    previous_rule_index: int = None  # get rule using ruleset
```
akx, line 779:
```suggestion
    SUBSTITUTION: decimal.Decimal | None = None
    INTEGRAL: decimal.Decimal | None = None
    REMAINDER: decimal.Decimal | None = None
    previous_rule_index: int | None = None  # None = get rule using ruleset
```

babel/rbnf.py
```diff
@@ -0,0 +1,814 @@
...
+            # not coming from a PREVIOUS TOKEN
+            index = self.get_rule_integral(integral)
+
+        rule = self.rules[index]
+        integral2, remainder2 = divmod(integral, rule.divisor)
+        context.REMAINDER = remainder2
+        context.INTEGRAL = integral2
+        context.previous_rule_index = max(0, index - 1)  # get rule using ruleset
+        context.omit_optional = (remainder2 != 0)  # only if not even multiple
+        return rule.apply(number, context)
```
akx, line 558: Hmm, mutating the context in-place sounds dangerous. Is it? (Should the context be `frozen=True`, and you'd use `dataclasses.replace()` to "fork" new copies of it?)

babel/rbnf.py
```diff
@@ -0,0 +1,814 @@
...
+
+    def get_rule_special(self, val, strict=False):
```
akx, line 560:
```suggestion
    def get_rule_special(self, val: ..., strict: bool = False) -> ...:
```

babel/rbnf.py
```diff
@@ -0,0 +1,814 @@
...
+
+class Rule:
+    """
+    base value, a divisor, rule text, and zero, one, or two substitutions.
+    """
+    specials = {
+        NEGATIVE_NUMBER_RULE, IMPROPER_FRACTION_RULE,
+        PROPER_FRACTION_RULE, MASTER_RULE, INFINITY_RULE,
+        NOT_A_NUMBER_RULE, SPECIAL_FRACTION_RULE,
+    }
```
akx, line 660: Let's make this a module-level constant (`SPECIAL_RULES`?) instead of hiding it here as a cvar?

babel/rbnf.py
```diff
@@ -0,0 +1,814 @@
...
+
+    def get_rule_special(self, val, strict=False):
+        if val in Rule.specials:
+            for r in self.rules:
+                if r.value == val:
+                    return r
+
+        # return last rule if no match occurred and strict is false
+        if not strict:
+            return self.rules[-1]
```
akx, line 568: Looks like this function is currently only ever called with `strict=True`, so it should probably be strict all the way:
```suggestion
    def get_rule_special(self, val):
        if val in Rule.specials:
            for r in self.rules:
                if r.value == val:
                    return r
        raise LookupError("No special rule {val!r}")
```
or something?

babel/rbnf.py
```diff
@@ -0,0 +1,814 @@
...
+        its base value is not an even multiple of its divisor, and the number
+        is an even multiple of the rule's divisor, use the rule that precedes
+        it in the rule list. Otherwise, use the rule itself.
+        """
+        # automatically return last rule if no range matched
+        ret = len(self.rules) - 1
+
+        for i in range(len(self.rules) - 1):
+            if self.rules[i].value in Rule.specials:
+                continue
```
akx, line 585: Do these special rules have some sort of intrinsic order, or should they be kept in a separate dict for ease of lookup, and so we wouldn't need to do this all the time?

IOW, `self.special_rules[NEGATIVE_NUMBER_RULE]` would be a simple dict lookup, and this loop for integral rules wouldn't need to check this specifically.

babel/rbnf.py
```diff
@@ -0,0 +1,814 @@
...
+                    return r
+
+        # return last rule if no match occurred and strict is false
+        if not strict:
+            return self.rules[-1]
+
+    def get_rule_integral(self, val):
+        """
+        Binary-search the rule list for the rule with the highest base value
+        less than or equal to the number.
```
akx, line 573: I'm not sure I'm seeing binary search here?

babel/rbnf.py
```diff
@@ -0,0 +1,814 @@
...
+        if val in Rule.specials:
+            for r in self.rules:
+                if r.value == val:
+                    return r
+
+        # return last rule if no match occurred and strict is false
+        if not strict:
+            return self.rules[-1]
+
+    def get_rule_integral(self, val):
```
akx, line 570:
```suggestion
    def get_rule_integral(self, val: ...) -> ...:
```

babel/rbnf.py
```diff
@@ -0,0 +1,814 @@
...
+
+    SPECIAL_FRACTION_RULE = 'x,x'  # there are other options but not existent in CLDR
+    """
+
+    __slots__ = ("name", "private", "rules")
+
+    def __init__(self, name, private=False):
+        self.name = name
+        self.private = private
+        self.rules = []
```
akx, line 486:
```suggestion
        self.rules: list[...] = []
```
(fill in the blank?)

babel/rbnf.py
```diff
@@ -0,0 +1,814 @@
...
+    #     # get the index of the closest 0 match
+    #     bst = min(dists, key=lambda x: x[1])[0]
+
+    #     # there is a following rule
+    #     if len(self.rules) > bst + 1 and \
+    #             self.rules[bst].value == self.rules[bst + 1].value and \
+    #             val * self.rules[bst].value > 1:
+    #         bst += 1
+
+    #     return bst
```
akx, line 641: Commented-out code should go :)
```suggestion
```

babel/rbnf.py
```diff
@@ -0,0 +1,814 @@
...
+
+    def __str__(self):
+        rules_str = '\n'.join(['\t' + str(r) for r in self.rules])
+        return f'Ruleset {self.name} {"private" if self.private else " "}\n{rules_str}\n'
+
+
+class Rule:
+    """
+    base value, a divisor, rule text, and zero, one, or two substitutions.
+    """
```
akx, line 655: This doesn't add much as a docstring?
```suggestion
```

babel/rbnf.py
```diff
@@ -0,0 +1,814 @@
...
+    specials = {
+        NEGATIVE_NUMBER_RULE, IMPROPER_FRACTION_RULE,
+        PROPER_FRACTION_RULE, MASTER_RULE, INFINITY_RULE,
+        NOT_A_NUMBER_RULE, SPECIAL_FRACTION_RULE,
+    }
+
+    __slots__ = ("value", "divisor", "tokens", "substitutions")
+
+    def __init__(self, value, text, radix=None):
```
akx, line 664:
```suggestion
    def __init__(self, value: ..., text: str, radix: ... | None = None) -> None:
```
(please fill in the blanks)

babel/rbnf.py
```diff
@@ -0,0 +1,814 @@
...
+    #         bst += 1
+
+    #     return bst
+
+
+    def __repr__(self):
+        return f"<Ruleset {self.name} {'private' if self.private else ' '}{len(self.rules)} rules>"
+
+    def __str__(self):
+        rules_str = '\n'.join(['\t' + str(r) for r in self.rules])
```
akx, line 648:
```suggestion
        rules_str = '\n'.join(f'\t{r}' for r in self.rules)
```

babel/rbnf.py
```diff
@@ -0,0 +1,814 @@
...
+        self.tokens = list(tokenize(text))
+        # could not decide if number of substitutions counted with or without optional ones
+        # counting optional causes infinite recursion in the `lt` locale
+        self.substitutions = len([t for t in self.tokens if t.type in REFERENCE_TOKENS if not t.optional])
+
+    def apply(self, number, context):
+        """
+        """
+        # print(f"RULE {self.value} - divisor: {self.divisor}")
+
```
akx, line 684:
```suggestion
```

babel/rbnf.py
```diff
@@ -0,0 +1,814 @@
...
+    Parse rules based on:
+    http://www.unicode.org/reports/tr35/tr35-47/tr35-numbers.html#Rule-Based_Number_Formatting
+    """
+    rbnf_rules = data.setdefault('rbnf_rules', {})
+
+    # ElementTree.dump(tree)
+
+    for ruleset_grouping in tree.findall('.//rbnf/rulesetGrouping'):
+        group_name = ruleset_grouping.attrib['type']
+        rbnf_rules[group_name] = []  # TODO check for overwrite
```
akx, line 803: Should this TODO be heeded now?

babel/rbnf.py
```diff
@@ -0,0 +1,814 @@
...
+        return the value of one of the all-caps params selected by the type of reference
+        """
+        return {
+            INTEGRAL_TOKEN: self.INTEGRAL,
+            REMAINDER_TOKEN: self.REMAINDER,
+            SUBSTITUTION_TOKEN: self.SUBSTITUTION,
+        }[typ]
+
+
+def parse_rbnf_rules(data, tree):
```
akx, line 792: Yeah, this definitely looks like it should be in `import_cldr.py`, not here :)

babel/rbnf.py
```diff
@@ -0,0 +1,814 @@
...
+
+    def _return_value_by_type(self, typ: int):
+        """
+        return the value of one of the all-caps params selected by the type of reference
+        """
+        return {
+            INTEGRAL_TOKEN: self.INTEGRAL,
+            REMAINDER_TOKEN: self.REMAINDER,
+            SUBSTITUTION_TOKEN: self.SUBSTITUTION,
+        }[typ]
```
akx, line 789: Probably shouldn't construct a dict just to pick out a value from it.


```suggestion
    def _return_value_by_type(self, typ: int) -> decimal.DECIMAL:
        if typ == INTEGRAL_TOKEN:
            return self.INTEGRAL
        if typ == REMAINDER_TOKEN:
            return self.REMAINDER
        if typ == SUBSTITUTION_TOKEN:
            return self.SUBSTITUTION
        raise ValueError("Invalid token type")
```

babel/rbnf.py
```diff
@@ -0,0 +1,814 @@
...
+                PUBLIC_REF: f"←%{ref}←",
+                PRIVATE_REF: f"←%%{ref}←",
+                DECIMAL_REF: f"←{ref}←",
+            }[ref_type]
+        elif t.type == PREVIOUS_TOKEN:
+            return "→→→"
+        elif t.type == PLURAL_TOKEN:
+            return f"$({','.join([f'{k}{{{v}}}' for k,v in t.reference.items()])})$"
+        else:
+            return ''
```
akx, line 767: Shouldn't `_str_token` be `Token`'s `__str__`? (Could be done if you use `typing.NamedTuple` for it, as commented above).

babel/rbnf.py
```diff
@@ -0,0 +1,814 @@
...
+        if t.type == TEXT_TOKEN:
+            return t.reference
+        elif t.type in REFERENCE_TOKENS:
+            ref_type, ref = t.reference
+            return {
+                INTERNAL_REF: f"←{ref}←",
+                PUBLIC_REF: f"←%{ref}←",
+                PRIVATE_REF: f"←%%{ref}←",
+                DECIMAL_REF: f"←{ref}←",
+            }[ref_type]
```
akx, line 761: Should be an if-else, please.

babel/rbnf.py
```diff
@@ -0,0 +1,814 @@
...
+        PROPER_FRACTION_RULE, MASTER_RULE, INFINITY_RULE,
+        NOT_A_NUMBER_RULE, SPECIAL_FRACTION_RULE,
+    }
+
+    __slots__ = ("value", "divisor", "tokens", "substitutions")
+
+    def __init__(self, value, text, radix=None):
+        """
+        divisor : iterator of literal, back_sub, fwd_sub, lit_exact elements parsed from rule
+        """
```
akx, line 667: Not a constructor docstring.
```suggestion
```

Could be a comment next to `self.divisor = ...`?

babel/rbnf.py
```diff
@@ -0,0 +1,814 @@
...
+        if value in self.specials:
+            self.value = value
+        else:
+            self.value = int(value)
+
+        self.divisor = compute_divisor(self.value, int(radix or 10))
+        self.tokens = list(tokenize(text))
+        # could not decide if number of substitutions counted with or without optional ones
+        # counting optional causes infinite recursion in the `lt` locale
+        self.substitutions = len([t for t in self.tokens if t.type in REFERENCE_TOKENS if not t.optional])
```
akx, line 678: Is... `n_references` a better name for this?

babel/rbnf.py
```diff
@@ -0,0 +1,814 @@
...
+
+    def __init__(self, name, private=False):
+        self.name = name
+        self.private = private
+        self.rules = []
+
+    def apply(self, raw_number, parent, fractional=False, index=None):
+        number = decimal.Decimal(str(raw_number))
+        # str is needed to avoid unnecessary precision
+        # decimal is necessary for exact representation in fraction rules
```
akx, line 491: If we don't support fraction rules right now, I think we shouldn't do the decimal roundtripping?

babel/rbnf.py
```diff
@@ -0,0 +1,814 @@
...
+    def get_rule_integral(self, val):
+        """
+        Binary-search the rule list for the rule with the highest base value
+        less than or equal to the number.
+
+        If that rule has two substitutions,
+        its base value is not an even multiple of its divisor, and the number
+        is an even multiple of the rule's divisor, use the rule that precedes
+        it in the rule list. Otherwise, use the rule itself.
+        """
```
akx, line 579: Not exactly a docstring, can you make it a comment?

babel/rbnf.py
```diff
@@ -0,0 +1,814 @@
...
+        divisor : iterator of literal, back_sub, fwd_sub, lit_exact elements parsed from rule
+        """
+        # TODO handle specials separatelly?
+        if value in self.specials:
+            self.value = value
+        else:
+            self.value = int(value)
+
+        self.divisor = compute_divisor(self.value, int(radix or 10))
```
akx, line 674: Some special handling anyway, I think, like this? Does it make sense at all to try and compute the divisor for a special value?
```suggestion
        if value in self.specials:
            self.value = value
            self.divisor = None
        else:
            self.value = int(value)
            self.divisor = compute_divisor(self.value, int(radix or 10))
```

babel/rbnf.py
```diff
@@ -0,0 +1,814 @@
...
+        self.tokens = list(tokenize(text))
+        # could not decide if number of substitutions counted with or without optional ones
+        # counting optional causes infinite recursion in the `lt` locale
+        self.substitutions = len([t for t in self.tokens if t.type in REFERENCE_TOKENS if not t.optional])
+
+    def apply(self, number, context):
```
akx, line 680:
```suggestion
    def apply(self, number: ..., context: ParsingContext):
```
(please fill in the blanks)

babel/rbnf.py
```diff
@@ -0,0 +1,814 @@
...
+        """
+        """
+        # print(f"RULE {self.value} - divisor: {self.divisor}")
+
+        res = []
```
akx, line 685: It could make sense to make this a generator function over the string atoms instead of gathering them into a list (unless we need to peek at the list!).

That way the recursion becomes a beautiful `yield from ruleset.apply(...)` too?

babel/rbnf.py
```diff
@@ -0,0 +1,814 @@
...
+                form = context.speller.plural_rule(number)
+                # no need to test plural rules here, other should always be present
+                # if form not in t.reference and "other" not in t.reference:
+                #     raise RuleMalformed(f"Plural form {form} not in {self} and no fallback option ('other') either!")
+
+                res.append(t.reference[form if form in t.reference else 'other'])
+
+            # when the tokenisation succeeds this never happens
+            # else:
+            #     raise ValueError(f'unknown token {t}', t)
```
akx, line 736: This dead code should be made live! 😄 🧟 


```suggestion
            else:  # pragma: no cover
                # If tokenization succeeded, we should never reach this
                raise ValueError(f'unknown token {t}', t)
```

babel/rbnf.py
```diff
@@ -0,0 +1,814 @@
...
+                    # if t.type == REMAINDER_TOKEN and context.remainder_as_fractional:
+                    #     fractional = True
+                    # else:
+                    #     fractional = context.fractional
+                    fractional = context.fractional
+                    res.append(ruleset.apply(
+                        context._return_value_by_type(t.type),  # number
+                        context.speller,  # parent
+                        fractional,
+                    ))
```
akx, line 716: The guard should be unnecessary; if `ruleset` is None, we'll probably want to die hard here instead of skipping formatting stuff?
```suggestion
                res.append(ruleset.apply(
                    context._return_value_by_type(t.type),
                    parent=context.speller,
                    fractional=context.fractional,
                ))
```

babel/rbnf.py
```diff
@@ -0,0 +1,814 @@
...
+                    index=context.previous_rule_index,
+                ))
+
+            elif t.type == PLURAL_TOKEN:
+                form = context.speller.plural_rule(number)
+                # no need to test plural rules here, other should always be present
+                # if form not in t.reference and "other" not in t.reference:
+                #     raise RuleMalformed(f"Plural form {form} not in {self} and no fallback option ('other') either!")
+
+                res.append(t.reference[form if form in t.reference else 'other'])
```
akx, line 732: What's happening here...? Wasn't the type of `t.reference` a 2-tuple? 🤔

babel/rbnf.py
```diff
@@ -0,0 +1,814 @@
...
+
+    def __repr__(self):
+        return f"<Rule {self.value} - {self.divisor} {len(self.tokens)} tokens>"
+
+    def __str__(self):
+        tokens_str = ''.join([
+            f"[{self._str_token(t)}]"
+            if t.optional else
+            f"{self._str_token(t)}"
+        for t in self.tokens])
```
akx, line 748:
```suggestion
        tokens_str = ''.join(f"[{self._str_token(t)}]" if t.optional else f"{self._str_token(t)}" for t in self.tokens)
```

tests/test_number_spelling.py
```diff
@@ -0,0 +1,126 @@
...
+soft_hyphen = '\xad'
+
+
+def test_basic():
+    x = rbnf.RuleBasedNumberFormat.negotiate('hu_HU')
+    assert str(x._locale) == 'hu'
+    assert 'spellout-numbering' in x.available_rulesets
+
+
+@pytest.mark.all_rbnf_locales
```
akx, line 17: Please add the mark to the pytest configuration file to quiesce the warning :)

tests/test_number_spelling.py
```diff
@@ -0,0 +1,126 @@
...
+        negotiated_speller.match_ruleset("nonexistent")
+
+
+def test_tokenization():
+    x = list(rbnf.tokenize("text[opt];"))
+    res = [
+        rbnf.TokenInfo(type=1, reference='text', optional=False),
+        rbnf.TokenInfo(type=1, reference='opt', optional=True),
+    ]
+    assert x == res
```
akx, line 38:
```suggestion
    assert list(rbnf.tokenize("text[opt];")) == [
        rbnf.TokenInfo(type=1, reference='text', optional=False),
        rbnf.TokenInfo(type=1, reference='opt', optional=True),
    ]
```

tests/test_number_spelling.py
```diff
@@ -0,0 +1,126 @@
...
+        list(rbnf.tokenize("==="))
+
+    with pytest.warns(SyntaxWarning, match=r"Reference parsing error.*"):
+        list(rbnf.tokenize("←bad←;"))
+
+
+@pytest.mark.all_rbnf_locales
+def test_xml_parsing(locale):
+    """
+    All the rues implicitly go through the arsing during CLDR import.
```
akx, line 52: ... What? 😄 

Either way, I don't think this test needs to be here, especially if/when the parse code is moved to import_cldr as commented there.

tests/rbnf_test_cases/_utils.py
```diff
@@ -0,0 +1,105 @@
+import os
+
+import tomllib
+
+from babel.core import Locale, get_global
+from babel.rbnf import RBNFError, RuleBasedNumberFormat
+
+print(os.system("pwd"))
+
+
```
akx, line 10:
```suggestion
```

tests/rbnf_test_cases/_utils.py
```diff
@@ -0,0 +1,105 @@
...
+            with open("tests/rbnf_test_cases/_template.toml", "rb") as f:
+                test_cases_temp = tomllib.load(f)
+                test_cases = test_cases_temp
+        else:
+            with open(template_toml_path, "rb") as f:
+                test_cases_temp = tomllib.load(f)
+    else:
+        test_cases_temp = test_cases
+
+    return test_cases_temp[get_mapped_ruleset_name(ruleset_name)].items()
```
akx, line 31: Looks like this simplifies to something like
(import `functools.cache` and `pathlib.Path`) 
```suggestion

@cache
def _get_test_cases_template(path: str = "tests/rbnf_test_cases/_template.toml"):
    return tomllib.load(Path(path).read_text())


def get_test_cases(ruleset_name: str):
    return _get_test_cases_template()[get_mapped_ruleset_name(ruleset_name)]
```

tests/rbnf_test_cases/_utils.py
```diff
@@ -0,0 +1,105 @@
...
+    mapping = {
+        "spellout-numbering-year": "year",
+        "spellout-numbering": "numbering",
+        "spellout-ordinal": "ordinal",
+        "spellout-cardinal": "cardinal",
+    }
+    for k, v in mapping.items():
+        if ruleset.startswith(k):
+            return v
+    return 'numbering'  # default fallback
```
akx, line 45:
```suggestion
mapping = {
    "spellout-numbering-year": "year",
    "spellout-numbering": "numbering",
    "spellout-ordinal": "ordinal",
    "spellout-cardinal": "cardinal",
}

def get_mapped_ruleset_name(ruleset: str) -> str:
    for k, v in mapping.items():
        if ruleset.startswith(k):
            return v
    return 'numbering'  # default fallback
```

tests/rbnf_test_cases/_utils.py
```diff
@@ -0,0 +1,105 @@
...
+
+
+def generate_test_for_locale(
+        locale: Locale,
+        output_toml_path: str,
+        test_cases: dict = None,
+
+
+) -> None:
```
akx, line 54:
```suggestion
def generate_test_for_locale(
    locale: Locale,
    output_toml_path: str,
) -> None:
```

tests/rbnf_test_cases/_utils.py
```diff
@@ -0,0 +1,105 @@
...
+
+        for k, _ in get_test_cases(ruleset_name=ruleset):
+
+            try:
+                v2 = speller.format(k, ruleset=ruleset)
+                print(f"    {k} : '{v2}'")
+                lines.append(f'{k} = "{v2}"')
+            except RBNFError as e:
+                print(k, locale, ruleset, e)
+                input()
```
akx, line 80:
```suggestion
            v2 = speller.format(k, ruleset=ruleset)
            lines.append(f'{k} = "{v2}"')
```

tests/rbnf_test_cases/_utils.py
```diff
@@ -0,0 +1,105 @@
...
+
+
+def generate_all_tests(
+        test_cases_toml_path: str = "tests/rbnf_test_cases/_template.toml",
+        output_dir: str = "tests/rbnf_test_cases/",
+) -> None:
+
+    for locale in list(get_global('rbnf_locales')):
+        output_toml_path = os.path.join(output_dir, f"{locale}.toml")
+        generate_test_for_locale(locale, output_toml_path)
```
akx, line 95:
```suggestion
def generate_all_tests() -> None:
    for locale in get_global('rbnf_locales'):
        generate_test_for_locale(locale, "tests/rbnf_test_cases/{locale}.toml")
```

akx (changes requested):
Thanks for reviving the number spelling challenge, and sorry it took me so long to get to reviewing this.

Beyond the comments within (mostly lack of type hints etc, but some other things too), please remove the non-human-audited (were there audited ones?) `rbnf_test_cases`? We don't want to add all of that to the repo here (especially if/when we don't know if they're correct). The script to generate them should naturally stay there (some comments for the script too).

I did take a quick look at the test cases for my native Finnish, and they look correct :)

Issue #1250, opened 2026-02-03
ImportError: cannot import name 'number_re' from 'babel.numbers' since 2.18.0


m-stefanski:
Hi everyone, 

It seems that 2.18.0 has broken xlsx2html, as described in this issue: https://github.com/Apkawa/xlsx2html/issues/57

```
    import xlsx2html as x2h
  File "/home/app/.local/lib/python3.10/site-packages/xlsx2html/__init__.py", line 3, in <module>
    from .core import xlsx2html
  File "/home/app/.local/lib/python3.10/site-packages/xlsx2html/core.py", line 12, in <module>
    from xlsx2html.format import format_cell
  File "/home/app/.local/lib/python3.10/site-packages/xlsx2html/format.py", line 12, in <module>
    from babel.numbers import (
ImportError: cannot import name 'number_re' from 'babel.numbers' (/home/app/.local/lib/python3.10/site-packages/babel/numbers.py)
```

I do not see any explicit `number_re` deprecation warning in changelog, is this change intentional?

I have also found the breaking change: https://github.com/akx/babel/commit/019ba6c799b1caddab2864aaa4d6037f0065cc92#diff-1f5648fadf1d08f1e2848f5ddc0fb038340f64e5dd566d8213b72b7372e9818eR1204

My undestanding is that non-private objects should not be "privatized" without at least some deprecation warning. 
I intend to fix it by renaming it back to historic name.

m-stefanski:
I have created a small PR fixing this, kindly approve it.

akx:
Hmm... well, the pattern isn't a [documented](https://babel.pocoo.org/en/latest/api/numbers.html) API, and its internals (the group names, or how it might work in the future – after all, this change in #1213 made things more efficient, and we might want to do more of that in the future) don't have a stability guarantee either – and the group names are evidently something `xlsx2html` relies on too. I didn't expect third-party code to be using it, but I was wrong :)

I'm a little hesitant to just "un-private" the API that is really meant to be private and malleable by Babel maintainers. (If there are multiple downstream users of that pattern and this turns out to be more of an issue, we could then re-expose `number_re` as a documented API.)

For now, the better solution would be for xlsx2html to vendor a copy of the pattern for its uses – and if you just need to run something with Babel 2.18 and `xlsx2html` right now, you could do something like

```python
import babel.numbers as bn
bn.number_re = bn._number_pattern_re
```

early enough in your code. Would that help?

m-stefanski:
[...]

PR #1251, opened 2026-02-03
Revert numbers.py breaking change of _number_pattern_re to number_re

m-stefanski:
Solves https://github.com/python-babel/babel/issues/1250


akx:
Closing since #1250 was closed, see rationale there.

PR #1253, opened 2026-02-11
Remove double newline at EOF of po files

### LINKED ISSUE #799: Two <BLANKLINE> in generated PO EOFs
As is documented, [`write_po` function renders two blank lines at the end of the file](https://github.com/python-babel/babel/blob/96acf8688c0e4b96788fbc35bb6d1640ef46a091/babel/messages/pofile.py#L497-L498). GNU gettext [documentation about PO files](https://www.gnu.org/software/gettext/manual/html_node/PO-Files.html) does not specify that this is mandatory.

I think that this makes no sense, but maybe something could explain why this has been the choice. It creates inconsistent output in diffs, uneeded extra space in disk, and so on... it is just a bad practice for text files.

CorentinDeBoisset:
Hello !

This is my first PR, so I'm not fully sure about everything. I used the unit tests to check for regressions, but are there other untested edge-cases we should worry about ?

Fixes #799

### DIFF
--- a/babel/messages/pofile.py
+++ b/babel/messages/pofile.py
@@ -665,10 +665,19 @@ def generate_po(
             yield f"{prefix}msgid {normalize(message.id, prefix=prefix, width=width)}\n"
             yield f"{prefix}msgstr {normalize(message.string or '', prefix=prefix, width=width)}\n"
 
+    first_message = True
+
     for message in _sort_messages(catalog, sort_by=sort_by):
-        if not message.id:  # This is the header "message"
-            if omit_header:
-                continue
+        is_header = not message.id
+        if is_header and omit_header:
+            continue
+
+        if first_message:
+            first_message = False
+        else:
+            yield '\n'
+
+        if is_header:
             comment_header = catalog.header_comment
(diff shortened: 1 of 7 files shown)


babel/messages/pofile.py
```diff
@@ -665,10 +665,19 @@ def _format_message(message, prefix=''):
...
+        is_header = not message.id
+        if is_header and omit_header:
+            continue
+
+        if first_message:
+            first_message = False
+        else:
+            yield '\n'
+
+        if is_header:
```
akx, line 680: Would this become simpler (not sure if it would!) if `first_message` was instead something like `emitted_messages` (boolean or integer)? IOW, the `if not message.id: if omit_header: continue` check would be as is in the preimage, and then something like

```
   if emitted_messages:
       # Emit a separator between this and the previous message
       yield '\n'
   emitted_messages = True
```
?

akx:
Thanks for the contribution! Please see the review comment; I'm not sure if it'd simplify things, but to me it looks like there's maybe more `if`s than necessary :)

PR #1254, opened 2026-02-12
Cache validity of a locale name

xmo-odoo:
Currently `_cache` is only populated on `load`, this is an issue with code which do a lot of locale parsing / instantiation but for one reason or an other rarely end up needing to actually load the locale data (e.g. date formatting with non-locale-dependent patterns) because every cache miss is an `os.path.exists`.

This change takes advantage of semantics differences between `exists` and `load`: `exists` just needs the entry to exist in the cache, while `load` needs the entry to be truthy. Thus `exists` can cache its lookup by setting an empty dict (or even `None` but that seems a bit too dodgy).

Alternatively we could remove the `os.path.exists` "slow path" and call `normalize_locale` every time, as `locale_identifiers` gets cached on first call, the initial call would be a lot more expensive but then no further `exists` would need to touch the filesystem.

For a third alternative, `exists` could be decorated with `functools.cache`, in order to have its own cache completely independent of `load`. `load` could also be migrated to `functools.cache` as, as far as I can tell, the `localedata._cache` is never invalidated anyway, with the drawback that a locale's data might be resolved multiple times if concurrent calls are tight enough (`functools.cache` synchronises the cache but not the resolution of the value).

### DIFF
--- a/babel/localedata.py
+++ b/babel/localedata.py
@@ -60,6 +60,7 @@ def resolve_locale_filename(name: os.PathLike[str] | str) -> str:
     return os.path.join(_dirname, f"{name}.dat")
 
 
+@lru_cache(maxsize=None)
 def exists(name: str) -> bool:
     """Check whether locale data is available for the given locale.
 
@@ -72,7 +73,7 @@ def exists(name: str) -> bool:
     if name in _cache:
         return True
     file_found = os.path.exists(resolve_locale_filename(name))
-    return True if file_found else bool(normalize_locale(name))
+    return file_found or bool(normalize_locale(name))
 
 
 @lru_cache(maxsize=None)
--- a/tests/test_localedata.py
+++ b/tests/test_localedata.py
@@ -109,10 +109,10 @@ def test_locale_argument_acceptance():
     assert normalized_locale is None
     assert not localedata.exists(None)
 
-    # Testing list input.
+    # Testing tuple input.
     normalized_locale = localedata.normalize_locale(['en_us', None])
     assert normalized_locale is None
-    assert not localedata.exists(['en_us', None])
+    assert not localedata.exists(('en_us', None))
 
 
 def test_locale_identifiers_cache(monkeypatch):


babel/localedata.py
```diff
@@ -72,7 +72,10 @@ def exists(name: str) -> bool:
     if name in _cache:
         return True
     file_found = os.path.exists(resolve_locale_filename(name))
-    return True if file_found else bool(normalize_locale(name))
+    if file_found or normalize_locale(name):
+        _cache.setdefault(name, {})
```
akx: This is clever!

Clever enough that it would definitely at least need a comment here and also somewhere around `_cache`'s declaration, and likely also in `load()`, so future readers can understand the semantics. In a "less type confusion" sense, `None` could be a better marker...

akx (changes requested):

babel/localedata.py
akx: (Aside, but just a thought... why do we guard _all_ of `load()` in the RLock...)

akx (changes requested):
On second thought, the additional cleverness is a little too clever. I think just `functools.cache` on `def exists()` is better.

babel/localedata.py
xmo-odoo: > (Aside, but just a thought... why do we guard _all_ of `load()` in the RLock...)

Checking the blame the reason seems lost to time, [the lock was already in the initial import](https://github.com/python-babel/babel/commit/00f16f91a878676a55aa1e6e85b5e1092f7cf8e9#diff-45c6f4d6218c9b1de24d7196e0b5f977bb4d0061b8aa45c8db0ffbc4ad315ce4R69). 

I would assume either something like "we can change it if it becomes an issue", or (and?) that given the gil there should be essentially no contention on the lock once all the relevant locales have been initialised, it would only occur if a release occurred in the small handful of instructions of the "fast path". This might be an assumption to reconsider given GIL-less runtimes, but at the same time unless someone's complained it might not be worth the risk?

If you can get in contact with Edgewall they might still have the old history stashed away somewhere but that seems unlikely.

xmo-odoo:
> On second thought, the additional cleverness is a little too clever. I think just `functools.cache` on `def exists()` is better.

After checking for safety, `functools.cache` was added in 3.9 but `python-requires` and the classifiers specify support for 3.8, so I used `lru_cache(maxsize=None)`. It's already used by the function right below as well as in extract so it's consistent.

I simplified the return expression as well, afaik `os.path.exists` already returns a genuine `bool` so there's no reason to use a ternary, looks like it was just leftover from previous changes.

xmo-odoo:
Ah looks like it's not happy because lru_cache blows up immediately on non-hashable content, and there's a test which passes a list to `exists`...

Should I just create a second separate manual cache instead then? Or update the test and `exists` can't be called with non-hashable content anymore? (technically it is typed as taking an `str`)

akx:
> After checking for safety, functools.cache was added in 3.9 but python-requires and the classifiers specify support for 3.8, so I used lru_cache(maxsize=None). It's already used by the function right below as well as in extract so it's consistent.

I think the next version of Babel will be the last to support 3.8 (which is EOL), so please add a `# TODO: Use @cache when dropping Python 3.8 support` style comment?

> Ah looks like it's not happy because lru_cache blows up immediately on non-hashable content, and there's a test which passes a list to `exists`...

Added in 1903a2bc514f9b5cf2559be65dff965cf5284bd0... Bit of a silly test.

> Or update the test and `exists` can't be called with non-hashable content anymore? (technically it is typed as taking an `str`)

Yeah, let's just change that test to pass in a tuple or something. 😅  Blowing up in `lru_cache` when the input isn't hashable is fine, no one should be doing that anyway (and they'll find out the hard way here!)

xmo-odoo:
> > After checking for safety, functools.cache was added in 3.9 but python-requires and the classifiers specify support for 3.8, so I used lru_cache(maxsize=None). It's already used by the function right below as well as in extract so it's consistent.
> 
> I think the next version of Babel will be the last to support 3.8 (which is EOL), so please add a `# TODO: Use @cache when dropping Python 3.8 support` style comment?

There's a few other locations where it's used the same way so might make more sense to go in a task / todo list for modernization than a spot comment just for that one? Dunno if you already have a task / milestone for that.

> > Or update the test and `exists` can't be called with non-hashable content anymore? (technically it is typed as taking an `str`)
> 
> Yeah, let's just change that test to pass in a tuple or something. 😅 Blowing up in `lru_cache` when the input isn't hashable is fine, no one should be doing that anyway (and they'll find out the hard way here!)

I'll do that then, thanks :+1:

akx (approved):

PR #1259, opened 2026-02-22
Skip byte strings in extract_python instead of crashing

### LINKED ISSUE #1190: `pybabel extract` crashes when it encounters a byte string
(Happens on current master)

Sample file:

```python
_(b'foo')
```

now run `pybabel extract test.py -o messages.pot` and you'll get this error:

```python
> pybabel extract test.py -o messages.pot
extracting messages from test.py
  File "/home/dev/babel/venv/bin/pybabel", line 8, in <module>
    sys.exit(main())
             ^^^^^^
  File "/home/dev/babel/babel/messages/frontend.py", line 998, in main
  File "/home/dev/babel/babel/messages/frontend.py", line 924, in run
    return cmdinst.run()
           ^^^^^^^^^^^^^
  File "/home/dev/babel/babel/messages/frontend.py", line 522, in run
                                                        ^^^^^^^^^
  File "/home/dev/babel/babel/messages/extract.py", line 282, in check_and_call_extract_file
    for message_tuple in extract_from_file(
                         ^^^^^^^^^^^^^^^^^^
  File "/home/dev/babel/babel/messages/extract.py", line 324, in extract_from_file
    return list(extract(method, fileobj, keywords, comment_tags,
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/dev/babel/babel/messages/extract.py", line 451, in extract
                                                ^^^^^^^
  File "/home/dev/babel/babel/messages/extract.py", line 562, in extract_python
    messages.append(''.join(buf))
                    ^^^^^^^^^^^^
TypeError: sequence item 0: expected str instance, bytes found
```

It seems like we're trying to call `.join()` on a byte string. I suggest just fixing the crash but keeping byte strings disallowed - they won't work with `gettext` anyway because gettext uses normal strings internally.

bysiber:
When `pybabel extract` encounters a byte string like `_(b'foo')`, `_parse_python_string` returns a `bytes` object which then causes `''.join(buf)` to crash with `TypeError: sequence item 0: expected str instance, bytes found`.

Since byte strings are not valid translation strings (gettext uses normal strings internally), this just skips them by returning `None` from `_parse_python_string` when it encounters bytes.

Added a test that confirms byte strings are silently skipped while normal strings in the same file are still extracted correctly.

Fixes #1190

### DIFF
--- a/babel/messages/extract.py
+++ b/babel/messages/extract.py
@@ -710,6 +710,8 @@ def _parse_python_string(value: str, encoding: str, future_flags: int) -> str | 
     if isinstance(code, ast.Expression):
         body = code.body
         if isinstance(body, ast.Constant):
+            if isinstance(body.value, bytes):
+                return None  # byte strings are not translatable
             return body.value
         if isinstance(body, ast.JoinedStr):  # f-string
             if all(isinstance(node, ast.Constant) for node in body.values):
--- a/tests/messages/test_extract.py
+++ b/tests/messages/test_extract.py
@@ -180,3 +180,17 @@ foof = _(
         'NOTE: This should still be considered, even if',
         'the text is far away',
     ]
+
+
+def test_extract_byte_string_does_not_crash():
+    """Byte strings like _(b'foo') should be skipped without crashing."""
+    buf = BytesIO(b"""\
+_(b'foo')
+_('hello')
+_(b"bar")
+_('world')
+""")
+    messages = list(extract.extract('python', buf, {'_': None}, [], {}))
+    assert len(messages) == 2
+    assert messages[0][1] == 'hello'
+    assert messages[1][1] == 'world'


babel/messages/extract.py
```diff
@@ -710,6 +710,8 @@ def _parse_python_string(value: str, encoding: str, future_flags: int) -> str |
     if isinstance(code, ast.Expression):
         body = code.body
         if isinstance(body, ast.Constant):
+            if isinstance(body.value, bytes):
+                return None  # byte strings are not translatable
```
akx, line 714: Like for the f-string case below, let's add a note for our future selves that we could warn in this case.
```suggestion
            if isinstance(body.value, bytes):
                # TODO: we could raise an error or warning when this occurs
                return None  # byte strings are not translatable
```

akx (changes requested):
LGTM, tiny addition request!

PR #1261, opened 2026-03-12
fix(mofile): decode msgctxt to str in read_mo

### LINKED ISSUE #1147: `read_mo` does not decode `msgctxt`
When reading an MO file, the message context is kept as bytes unlike the `msgid` and `msgstr` which are decoded to utf8. This is surprising because in order to use for example `Catalog.get`, one must pass the context encoded as bytes. I wonder if this is intentional or just an omission?

Example:
When I read an MO file compiled from this PO file:
```po
msgctxt "fooctxt"
msgid "foo"
msgstr "bar"
```
The message key is `('foo', b'fooctxt')` (notice the that `fooctxt` is a bytes object):
```python
>>> f = open('messages.mo', 'rb')
>>> cat = read_mo(f)
>>> cat._messages
OrderedDict({('foo', b'fooctxt'): <Message 'foo' (flags: [])>})
```

Dhi13man:
## Summary

- `read_mo` keeps `msgctxt` as bytes while `msgid`/`msgstr` are decoded to strings, causing `Catalog.get(id, context='...')` to fail for messages with context
- Added `ctxt.decode(catalog.charset)` after splitting on the `\x04` separator, consistent with how `msgid`/`msgstr` are decoded and how `write_mo` expects `context` to be `str`
- Added two regression tests: context round-trip through write_mo → read_mo, and non-ASCII context handling

Closes #1147

## Test plan

- [x] `test_read_mo_decodes_context` — verifies context is str, messages retrievable by string context, and no-context messages unaffected
- [x] `test_read_mo_decodes_non_ascii_context` — verifies non-ASCII context (German umlauts) round-trips correctly
- [x] Full test suite passes (7305 passed, pre-existing timezone failures excluded)
- [x] `ruff check` and `ruff format --check` clean

Generated with [Dhiman's Agentic Suite](https://github.com/Dhi13man)

### DIFF
--- a/babel/messages/mofile.py
+++ b/babel/messages/mofile.py
@@ -84,6 +84,7 @@ def read_mo(fileobj: SupportsRead[bytes]) -> Catalog:
 
         if b'\x04' in msg:  # context
             ctxt, msg = msg.split(b'\x04')
+            ctxt = ctxt.decode(catalog.charset)
         else:
             ctxt = None
 
--- a/tests/messages/test_mofile.py
+++ b/tests/messages/test_mofile.py
@@ -20,8 +20,7 @@ data_dir = os.path.join(os.path.dirname(__file__), 'data')
 
 
 def test_basics():
-    mo_path = os.path.join(data_dir, 'project', 'i18n', 'de',
-                           'LC_MESSAGES', 'messages.mo')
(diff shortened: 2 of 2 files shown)


Dhi13man:
I had leftover Claude Max and wanted to contribute to backlog clearing in open source projects before it reset, but I see that this could be a nuisance. My apologies.

PR #1264, opened 2026-04-10
Try to hold `_cache_lock` less in `load()`

akx:
Shouldn't be a need to acquire the lock just to return the already-cached data.


akx:
cc @xmo-odoo (re #1254).

PR #1271, opened 2026-05-09
Fix Locale subclass __str__ issue

### LINKED ISSUE #1252: Locale class uses str(...) in __init__
Hello!

I noticed that the Locale class under `babel/core.py` uses str(...) when getting it's identifiers.
This can pose an issue for overriding classes that wish to modify the str representation of the object.

Looking at the class, all other points in the class (and str dunder itself) use a different function (`get_locale_identifier`), that can be used in __init__ too, and free __str__ to modifications. As of right now modifying it causes a bug on creating the Locale object properly and limiting usage.

```python
def __init__(
      self,
      language: str,
      territory: str | None = None,
      script: str | None = None,
      variant: str | None = None,
      modifier: str | None = None,
  ) -> None:
      """Initialize the locale object from the given identifier components.

      >>> locale = Locale('en', 'US')
      >>> locale.language
      'en'
      >>> locale.territory
      'US'

      :param language: the language code
      :param territory: the territory (country or region) code
      :param script: the script code
      :param variant: the variant code
      :param modifier: a modifier (following the '@' symbol, sometimes called '@variant')
      :raise `UnknownLocaleError`: if no locale data is available for the
                                   requested locale
      """
      #: the language code
      self.language = language
      #: the territory (country or region) code
      self.territory = territory
      #: the script code
      self.script = script
      #: the variant code
      self.variant = variant
      #: the modifier
      self.modifier = modifier
      self.__data: localedata.LocaleDataDict | None = None

      # identifier = str(this) <-- this is the old part, the following is the proposed solution to free __str__ up for inheriting classes
      identifier = get_locale_identifier(
          (self.language, self.territory, self.script, self.variant, self.modifier),
      )
      identifier_without_modifier = identifier.partition('@')[0]
      if localedata.exists(identifier):
          self.__data_identifier = identifier
      elif localedata.exists(identifier_without_modifier):
          self.__data_identifier = identifier_without_modifier
      else:
          raise UnknownLocaleError(identifier)

# Current implementation of str in the Locale class
def __str__(self) -> str:
      return get_locale_identifier(
          (self.language, self.territory, self.script, self.variant, self.modifier),
      )
```

Chiifouu:
Fixes #1252 

I noticed that Locale.__init__ was using str(self) to build the identifier. This breaks when a subclass overrides __str__ with a custom format.

This PR replaces it with get_locale_identifier() and adds a small test to make sure subclassed Locale objects work correctly.

### DIFF
--- a/babel/core.py
+++ b/babel/core.py
@@ -211,7 +211,9 @@ class Locale:
         self.modifier = modifier
         self.__data: localedata.LocaleDataDict | None = None
 
-        identifier = str(self)
+        identifier = get_locale_identifier(
+            (self.language, self.territory, self.script, self.variant, self.modifier),
+        )
         identifier_without_modifier = identifier.partition('@')[0]
         if localedata.exists(identifier):
             self.__data_identifier = identifier
--- a/tests/test_core.py
+++ b/tests/test_core.py
@@ -83,6 +83,14 @@ class TestLocaleClass:
         locale = Locale('en', 'US')
         assert locale.language == 'en'
         assert locale.territory == 'US'
+    
+    def test_locale_subclass_custom_str(self):
+        class MyLocale(Locale):
+            def __str__(self):
+                return f"[{self.language}]"
+        locale = MyLocale("en")
+        assert locale.language == "en"
+        assert str(locale) == "[en]"
 
     def test_default(self, monkeypatch):
         for name in ['LANGUAGE', 'LC_ALL', 'LC_CTYPE', 'LC_MESSAGES']:


nathaniel-kotzrenko-hs (approved):

akx (changes requested):
Instead of duplicating the `get_locale_identifier` invocation from `__str__`, maybe we should add a new `@property` called `_identifier` and use it in both places?

nathaniel-kotzrenko-hs:
@akx imo that would change the standard usage of the existing ```get_locale_identifier``` function.
In this case it makes more sense to modify all call sites (there are several in this class) to use the ```@property``` but won't it just mean another layer of indirection?

I think @Chiifouu's solution is good for this case, it doesn't modify the existing standard but does decouple ```__str__``` behavior from affecting the class, clean and straightforward, no need to modify the existing standard.

akx:
@nathaniel-kotzrenko-hs 

> imo that would change the standard usage of the existing get_locale_identifier function.

I don't see how - you'd use `get_locale_identifier()` to build an identifier from raw parts as before.

> In this case it makes more sense to modify all call sites (there are several in this class) 

As far as I can see, there's only one call site that uses the Locale instance's values, and that's `__str__`. Other call sites (`Locale.parse`, `default_locale()`) don't.

> but won't it just mean another layer of indirection?

In my eyes, it'll be the one canonical way to get the instance's full locale identifier, even if `__str__` is overridden.

To be clear, I mean `Locale.__str__()` should just return `self._identifier`, and places where `str(locale)` are being currently used should use `locale._identifier`. (I'm slightly hesitant about making it `identifier`, i.e. a public API.)

PR #1272, opened 2026-05-11
harden localedata loading with restricted unpickler

uwezkhan:
this replaces the direct pickle.load() usage in babel.localedata with a restricted unpickler.

right now locale data files are loaded using pickle without restricting what objects can be created during deserialization. this change adds a _SafeUnpickler which only allows the small set of classes and builtin types that babel actually needs for locale data loading.

also added security tests to make sure:

normal locale data still loads correctly
malicious pickle payloads are rejected
unsafe globals cannot be loaded during deserialization

the existing `.dat` files continue to work and no changes are needed for the locale generation process.

this is mainly a defense in depth hardening change to make locale data loading safer against malicious or corrupted pickle data.

### DIFF
--- a/babel/localedata.py
+++ b/babel/localedata.py
@@ -118,6 +118,19 @@ def _is_non_likely_script(name: str) -> bool:
     return False
 
 
+class _SafeUnpickler(pickle.Unpickler):
+    def find_class(self, module, name):
+        if module == 'babel.localedata' and name == 'Alias':
+            return Alias
+        if module == 'babel.plural' and name == 'PluralRule':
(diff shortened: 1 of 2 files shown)


uwezkhan:
any update ?

akx (commented):
Thank you for the PR, but I'm not sure it's necessary or complete.

* Did you try this with the current Babel data?
* Can you construct a scenario where `def load()` can load an arbitrary attacker-constructed pickle, or how an attacker could modify a pickle in the data directory to be malicious (without having permissions to similarly modify `babel/localedata.py`)?

PR #1276, opened 2026-05-27
Fix timezone display indianapolis 1260

yjaafri:
- Fix timezone display for "America/Indiana/Indianapolis" by resolving the correct locale-backed timezone data instead of falling back to GMT offsets
- add a regression test for the "zoneinfo" case.


yjaafri:
Hi, sorry for the delay. I was busy with university work and I accidentally left my pull request unmerged. I hope this fix looks good to you. I will be glad to contribute to more issues in the future !

PR #1278, opened 2026-06-09
decode message context to str in read_mo

rootvector2:
`read_mo` splits the context off a message but never decodes it, so a `.mo` carrying a msgctxt yields a `Message.context` of `bytes` that mis-keys the entry and crashes re-export through `write_mo`; found by round-tripping a catalog with a context.


rootvector2:
gentle ping

PR #1281, opened 2026-06-24
Apply exponent grouping for scientific notation patterns

### LINKED ISSUE #1026: Incorrect expected value in exponent grouping test
## Overview Description

According to [specifications for scientific notation in UNICODE LOCALE DATA MARKUP LANGUAGE (LDML) - PART 3: NUMBERS](https://unicode.org/reports/tr35/tr35-numbers.html#sci):

> The maximum number of integer digits, if present, specifies the exponent grouping. The most common use of this is to generate engineering notation, in which the exponent is a multiple of three, for example, "##0.###E0". The number 12345 is formatted using "##0.####E0" as "12.345E3".

https://github.com/python-babel/babel/blob/9ef53c6a6ab5fc604b58ccb19dc63ebcf8edd28b/tests/test_numbers.py#L93-L94

lin-hongkuan:
Fixes #1026.

LDML scientific notation treats the maximum number of integer digits in a scientific pattern as the exponent grouping size. Babel only shifted by the minimum integer digit count, so patterns such as `##0.####E0` formatted `12345` as `1.2345E4` instead of `12.345E3`.

This changes the scientific notation normalization to use the maximum integer digit count when it is larger than the minimum, while preserving the existing fixed-minimum behavior for patterns without exponent grouping.

Checks run locally:
- `python setup.py import_cldr`
- `python -m pytest tests\test_numbers.py tests\test_numbers_format_decimal.py -q` (140 passed)
- `python -m pre_commit run --files babel\numbers.py tests\test_numbers.py tests\test_numbers_format_decimal.py --show-diff-on-failure --color=never`
- `PYTHONUTF8=1 python -m pytest -q` (7478 passed, 8 skipped, 2 xfailed)

### DIFF
--- a/babel/numbers.py
+++ b/babel/numbers.py
@@ -1424,10 +1424,14 @@ class NumberPattern:
         value = value * get_decimal_quantum(exp)
         assert value.adjusted() == 0
 
-        # Shift exponent and value by the minimum number of leading digits
-        # imposed by the rendering pattern. And always make that number
-        # greater or equal to 1.
-        lead_shift = max([1, min(self.int_prec)]) - 1
+        # Shift exponent and value by the number of leading digits imposed by
+        # the rendering pattern. If a maximum integer digit count is present
+        # and larger than the minimum, it specifies exponent grouping.
+        min_int, max_int = self.int_prec
+        if max_int > min_int and max_int > 1:
+            lead_shift = exp % max_int
+        else:
+            lead_shift = max([1, min_int]) - 1
         exp = exp - lead_shift
         value = value * get_decimal_quantum(-lead_shift)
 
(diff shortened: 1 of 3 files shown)


babel/numbers.py
```diff
@@ -1424,10 +1424,14 @@ def scientific_notation_elements(
...
-        # greater or equal to 1.
-        lead_shift = max([1, min(self.int_prec)]) - 1
+        # Shift exponent and value by the number of leading digits imposed by
+        # the rendering pattern. If a maximum integer digit count is present
+        # and larger than the minimum, it specifies exponent grouping.
+        min_int, max_int = self.int_prec
+        if max_int > min_int and max_int > 1:
+            lead_shift = exp % max_int
+        else:
+            lead_shift = max([1, min_int]) - 1
```
akx, line 1434: Useless list construction, might as well do
```suggestion
            lead_shift = max(1, min_int) - 1
```

akx (changes requested):
Could you show another implementation of the CLDR/LDML doing this? With this sort of change I'd really love to be sure.

PR #1282, opened 2026-06-24
Fix parse_date and parse_time to raise ParseError instead of ValueError

### LINKED ISSUE #1178: parse_date raises `ValueError: day is out of range for month` for iso8601 string
```
$ /venv/py311/bin/python
Python 3.11.11 (main, Dec  4 2024, 08:55:08) [GCC 9.4.0] on linux
>>> import babel
>>> babel.__version__
'2.16.0'
>>> from babel.dates import format_date
>>> from babel.dates import parse_date
>>> parse_date('2005-12-31T01:02:03Z')
  File "/venv/py311/lib/python3.11/site-packages/babel/dates.py", line 1252, in parse_date
    return datetime.date(year, month, day)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
ValueError: day is out of range for month
```

binggao1230:
## Problem

`parse_date` and `parse_time` both catch `ParseError` as their documented error type, but when the parsed numbers produce an out-of-range date or time value, the internal `datetime.date` / `datetime.time` constructor raises a raw `ValueError` that escapes to the caller uncaught.

Reported in #1178. Example:

```python
>>> from babel.dates import parse_date
>>> parse_date('2005-12-31T01:02:03Z')
# Before: ValueError: day is out of range for month
# After: ParseError: String '2005-12-31T01:02:03Z' does not match format 'MMM d, y' (day is out of range for month)
```

```python
>>> from babel.dates import parse_time
>>> parse_time('25:00')
# Before: ValueError: hour must be in 0..23
# After: ParseError: String '25:00' does not match format 'h:mm:ss a' (hour must be in 0..23)
```

## Fix

Wrap the final `datetime.date(year, month, day)` call in `parse_date` and `datetime.time(hour, minute, second)` call in `parse_time` with `try/except ValueError`, re-raising as `ParseError` with an informative message that includes the input string and the expected format pattern.

All 2245 existing tests pass (2 xfailed unchanged).

Fixes #1178.

### DIFF
--- a/babel/dates.py
+++ b/babel/dates.py
@@ -1330,7 +1330,10 @@ def parse_date(
     day = int(numbers[indexes['D']])
     if month > 12:
         month, day = day, month
-    return datetime.date(year, month, day)
+    try:
+        return datetime.date(year, month, day)
+    except ValueError as e:
+        raise ParseError(f"String '{string}' does not match format '{fmt.pattern}' ({e})") from None
 
 
 def parse_time(
@@ -1395,7 +1398,10 @@ def parse_time(
         minute = int(numbers[indexes['M']])
         if len(numbers) > 2:
             second = int(numbers[indexes['S']])
-    return datetime.time(hour, minute, second)
+    try:
+        return datetime.time(hour, minute, second)
+    except ValueError as e:
+        raise ParseError(f"String '{string}' does not match format '{fmt.pattern}' ({e})") from None
 
 
 class DateTimePattern:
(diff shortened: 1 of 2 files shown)


akx (changes requested):
This would need a test.

Also, I think it would be useful to raise from the ValueError, not `None`.

And... maybe the message needs some work? The exception is, after all, that the numbers we parsed out of the string don't represent a Gregorian date.

binggao1230:
All three applied in 0e92abc:

- both errors now chain the underlying `ValueError` (`from e`) instead of `from None`
- the message says the parsed numbers do not represent a valid Gregorian date/time
- `test_parse_date_out_of_range` and `test_parse_time_out_of_range` cover the new path (day 30 in February, hour 25); 2279 pass.

PR #1293, opened 2026-07-18
Don't crash on blank PO-Revision-Date / POT-Creation-Date headers

### LINKED ISSUE #1219: Exception raised if the revision date is blank

agu2347:
Fixes #1219.

Some tools (e.g. Poedit) can leave the `PO-Revision-Date` header blank in a generated `.po` file instead of omitting it or using the conventional `YEAR-MO-DA HO:MI+ZONE` placeholder. Parsing such a file crashes with:
```
```
because `_parse_datetime_header()` is called unconditionally with the blank value, as reported in the issue.

**Fix**: guard both call sites (`po-revision-date` and, for the same reason, `pot-creation-date`) to skip parsing when the header value is blank, keeping the previous/default value instead -- the same treatment already given to the `YEAR-MO-DA HO:MI+ZONE` placeholder value right above it, and to a blank `Language` header elsewhere in this same method (`self._set_locale(value or None)`).

**Testing**: added a regression test that parses a message block with both headers blank and asserts: no exception is raised, `creation_date`/`revision_date` fall back to their defaults rather than being corrupted, and no literal `None` leaks into the round-tripped `mime_headers` output. I confirmed the test fails with the original `ValueError` without the fix, and passes with it.

I also verified end-to-end with an actual `.po` file parse + write round-trip (using an installed release build with CLDR data, since a shallow clone here can't build that data on its own), confirming no `None` leaks into the output and that normal (non-blank) header values are completely unaffected. Ran the full existing `test_catalog.py` and `test_pofile.py` suites: all pass, no regressions.

### DIFF
--- a/babel/messages/catalog.py
+++ b/babel/messages/catalog.py
@@ -582,10 +582,16 @@ class Catalog:
                 self._num_plurals = int(params.get('nplurals', 2))
                 self._plural_expr = params.get('plural', '(n != 1)')
             elif name == 'pot-creation-date':
-                self.creation_date = _parse_datetime_header(value)
+                # Some tools (e.g. Poedit) may leave this header blank.
+                # Keep the previous/default value rather than crashing.
+                if value.strip():
+                    self.creation_date = _parse_datetime_header(value)
             elif name == 'po-revision-date':
-                # Keep the value if it's not the default one
-                if 'YEAR' not in value:
+                # Keep the value if it's not the default one.
+                # Some tools (e.g. Poedit) may also leave this header
+                # blank; keep the previous/default value in that case
+                # too, rather than crashing.
+                if 'YEAR' not in value and value.strip():
                     self.revision_date = _parse_datetime_header(value)
 
     mime_headers = property(
--- a/tests/messages/test_catalog.py
+++ b/tests/messages/test_catalog.py
@@ -346,6 +346,36 @@ def test_catalog_stores_datetime_correctly():
             assert value == '2009-03-09 15:47-0700'
 
 
+def test_catalog_handles_blank_revision_and_creation_date():
+    """Regression test for
+    https://github.com/python-babel/babel/issues/1219
+
+    Some tools (e.g. Poedit) can leave PO-Revision-Date and/or
+    POT-Creation-Date blank in the header. Parsing such a header must
+    not raise, and must not silently store an unparseable value that
+    would corrupt the header on write-out.
+    """
+    cat = catalog.Catalog()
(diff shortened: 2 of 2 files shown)


akx:
See https://github.com/python-babel/babel/pull/1222 for previous discussion.

agu2347:
Thanks for the pointer -- I read through the #1222 discussion. To make sure this PR doesn't repeat the concern raised there ("I don't think we should be silently ignoring malformed dates" / silent data loss on unparseable values): #1222 was about *any* unparseable date string being silently swallowed, discarding a value the user actually wrote. This PR is narrower and only triggers on a genuinely *blank* header value (`POT-Creation-Date:` / `PO-Revision-Date:` with nothing after the colon) -- something Poedit itself sometimes writes. In that case there's no user-authored value to lose; falling back to the existing default (or leaving the previous value, for revision-date) doesn't discard any real information the way silently accepting-and-mangling an actually malformed non-blank date would. Any other malformed-but-non-blank value still goes through `_parse_datetime_header` unchanged and will raise/behave exactly as before -- this PR doesn't touch that path. Happy to adjust the approach further if this distinction doesn't fully address the earlier concern.

Mukller:
Heads up for maintainers: there are currently **three open PRs** addressing this exact problem (blank `POT-Creation-Date`/`PO-Revision-Date` headers from tools like Poedit, crash in `_parse_datetime_header`, ref. #1219):

| PR | Approach | Blank value | Malformed non-blank value |
|---|---|---|---|
| #1293 (this one) | guard at call site (`value.strip()`) | skipped, default kept | **still raises ValueError** (current strict behavior) |
| #1311 | `with suppress(ValueError)` at call site | skipped | **also silently ignored** — broader behavior change than the stated goal |
| #1317 | parser returns `datetime \| None` | `None` returned, caller skips | raises |

Notes on the differences:

- This PR (#1293) is the most conservative: only *blank* values are ignored, genuinely malformed dates keep raising. The test exercises the real header-parsing path rather than calling `_set_mime_headers` directly, which I like.
- #1311's `suppress(ValueError)` silently accepts any unparseable date, which could mask real corruption — probably not what we want by default.
- #1317 moves the handling into the parser and changes its return contract to `datetime | None`; that's arguably cleaner architecturally, but callers of `_parse_datetime_header` outside `_set_mime_headers` (if any) would need auditing.

It would be great if @akx (or whoever picks this up) can choose a preferred approach so two of the three can be closed — happy to re-review whichever direction wins. All three reference the same motivating case, and all include tests, which is nice to see.

PR #1296, opened 2026-07-29
Add support for Python 3.15

akx:
The [Python 3.15 beta](https://discuss.python.org/t/python-3-15-0-beta-4-is-here) is out.


akx:
> `platform linux -- Python 3.15.0b4`

It's some newer code, but it checks out.

PR #1287, opened 2026-06-27
fix: prevent Hebrew locale from corrupting subsequent locale formatting

C1-BA-B1-F3:
## Summary

Fixes #1234

After using Hebrew locale (), subsequent calls to  with other locales (like , , ) returned Hebrew-formatted text instead of the requested locale.

## Root Cause

 was mutating the cached locale data when resolving aliases. Hebrew locale's  dict was the **same object** as root's  dict (due to shallow copying in ). When the alias at  was resolved for Hebrew, the resolved Hebrew month names were written back into the shared dict, corrupting the root locale data that all other locales inherit from.

## Reproduction

```python
from babel.dates import format_date
from datetime import datetime

date_obj = datetime(2025, 10, 15)

print(format_date(date_obj, 'LLLL', 'de'))  # Oktober
print(format_date(date_obj, 'LLLL', 'he'))  # אוקטובר
print(format_date(date_obj, 'LLLL', 'no'))  # אוקטובר (BUG: should be 'oktober')
print(format_date(date_obj, 'LLLL', 'fr'))  # אוקטובר (BUG: should be 'octobre')
```

## Fix

Remove the write-back in . The resolved values are no longer stored back into the original data dict, preventing mutation of shared/cached locale data.

```python
# Before (buggy):
if val is not orig:
    self._data[key] = val  # Mutates shared cache!
return val

# After (fixed):
return val  # No mutation
```

## Tests Added

- `test_locale_data_isolation_hebrew`: Verifies month names are not corrupted after Hebrew usage
- `test_locale_data_isolation_format_date`: Verifies format_date output is correct after Hebrew
- `test_locale_data_cache_not_mutated`: Verifies root data integrity after alias resolution

All 7325 existing tests pass with this change.

PR #1299, opened 2026-07-30
Memoize resolved locale data per instance, share LocaleDataDicts

### LINKED ISSUE #1234: Hebrew Locale Corrupts Subsequent Locale Formatting

akx:
Fixes #1234.

This measured to be a little faster than `master`, too.

This is somewhat related to #31 and #305.

## Duplicates

Closes #1263.
Closes #1274.
Closes #1287.
Closes #1294.

### DIFF
--- a/babel/localedata.py
+++ b/babel/localedata.py
@@ -174,6 +174,14 @@ def load(name: os.PathLike[str] | str, merge_inherited: bool = True) -> dict[str
         _cache_lock.release()
 
 
+def clear_caches() -> None:
+    """
+    Clear locale data caches.
+    """
+    with _cache_lock:
+        _cache.clear()
+
+
 def merge(dict1: MutableMapping[Any, Any], dict2: Mapping[Any, Any]) -> None:
     """Merge the data from `dict2` into the `dict1` dictionary, making copies
     of nested dictionaries.
--- a/babel/localedata.py
+++ b/babel/localedata.py
@@ -25,6 +25,7 @@ from itertools import chain
 from typing import Any
 
 _cache: dict[str, Any] = {}
+_dict_cache: dict[str, LocaleDataDict] = {}
 _cache_lock = threading.RLock()
 _dirname = os.path.join(os.path.dirname(__file__), 'locale-data')
 _windows_reserved_name_re = re.compile("^(con|prn|aux|nul|com[0-9]|lpt[0-9])$", re.I)
@@ -180,6 +181,21 @@ def clear_caches() -> None:
     """
     with _cache_lock:
         _cache.clear()
+        _dict_cache.clear()
+
+
+def get_locale_data(name: str) -> LocaleDataDict:
+    """Return an alias-resolving `LocaleDataDict` over the merged data for
+    the given locale.
+
+    The wrapper is cached and shared: repeated requests for the same locale
+    return the same object. Alias resolutions memoized within it are
+    locale-specific.
+    """
+    try:
+        return _dict_cache[name]
+    except KeyError:
+        return _dict_cache.setdefault(name, LocaleDataDict(load(name)))
 
 
 def merge(dict1: MutableMapping[Any, Any], dict2: Mapping[Any, Any]) -> None:
@@ -248,6 +264,9 @@ class Alias:
         return data
 
 
+_sentinel = object()
+
+
 class LocaleDataDict(abc.MutableMapping):
     """Dictionary wrapper that automatically resolves aliases to the actual
     values.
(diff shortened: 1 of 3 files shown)


babel/localedata.py
```diff
@@ -272,13 +304,19 @@ def __getitem__(self, key: str | int | None) -> Any:
...
-            self._data[key] = val
+            # Only resolved/wrapped values are memoized.
+            # Scalars are always read from `self._data`, so that
+            # manual writes into the backing data (possibly shared)
+            # stay visible.
+            self._resolved[key] = val
         return val
 
     def __setitem__(self, key: str | int | None, value: Any) -> None:
+        self._resolved.pop(key, None)
```
tomasr8, line 316: I was wondering why not also `self._resolved[key] = val`, but I guess it' so that the value is properly resolved next time it's accessed?

babel/localedata.py
```diff
@@ -174,6 +175,28 @@ def load(name: os.PathLike[str] | str, merge_inherited: bool = True) -> dict[str
...
 
+def clear_caches() -> None:
+    """
+    Clear locale data caches.
+    """
+    _cache.clear()
+    _dict_cache.clear()
+
+
+def get_locale_data(name: str) -> LocaleDataDict:
```
tomasr8, line 187: IIUC this is a separate optimization and not part of the main fix right?

babel/localedata.py
tomasr8, line 324: Now that `self._data` is shared, do we still need to make a copy of it here or can we just `LocaleDataDict(self._data, ...)`?

tomasr8 (approved):
Some questions and comments, otherwise looks good!

babel/localedata.py
akx: It's related: without this, performance would regress from `master` since resolutions were being done over and over for the common case of `..., locale="fi"`.

babel/localedata.py
akx: Yeap, because writes target `self._data`.

tomasr8: Right, I missed that

babel/localedata.py
akx: And now that #1300 landed, I rebased this on `master`, and: 

> **Merging this PR will improve performance by 11.63%**

Yay!

PR #1302, opened 2026-07-30
CLDR 48.2

akx:
I had apparently implemented this in the spring but forgor about it? Well.


akx:
Good catch by codspeed: we don't need to slow down all unit operations, just those that would fail due to deprecated aliases.

We'll just have to accept _some_ degradation here.

PR #1306, opened 2026-07-30
Support CLDR 48 relative date formats

akx:
New data in CLDR 48, let's make use of it!

### DIFF
--- a/babel/dates.py
+++ b/babel/dates.py
@@ -355,6 +355,57 @@ def get_era_names(
     return Locale.parse(locale or LC_TIME).eras[width]
 
 
+def get_relative_name(
+    field: str,
+    offset: int,
+    length: Literal['long', 'short', 'narrow'] = 'long',
+    locale: Locale | str | None = None,
+) -> str | None:
+    """Return the locale's literal name for a relative offset of a date field.
+
+    These are the CLDR ``<relative>`` strings such as "yesterday", "today",
+    "tomorrow" for ``field='day'``, or "last Monday"/"this Monday"/"next Monday"
+    for ``field='mon'``.
+
+    Returns ``None`` when the locale does not define a literal name for the
+    requested field/offset (in which case a phrase like "in 2 days" built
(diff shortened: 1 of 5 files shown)


tests/test_dates.py
```diff
@@ -583,6 +583,156 @@ def test_format_datetime(timezone_getter):
...
+    ) == 'tomorrow at 3:30:00\u202fPM UTC'
+    # German 'long': "um" joiner
+    assert dates.format_relative_datetime(
+        'morgen', time(15, 30), format='long', locale='de',
+    ) == 'morgen um 15:30:00 UTC'
+    # French 'full': "à" joiner
+    assert dates.format_relative_datetime(
+        'demain', time(15, 30), format='full', locale='fr',
+    ) == 'demain à 15:30:00 temps universel coordonné'
+    # Czech has no relative pattern, so falls back to datetime_formats
```
tomasr8: Surprised there are no relative patterns since they should work fine here, but I guess they'll add them eventually..

babel/dates.py
```diff
@@ -355,6 +355,59 @@ def get_era_names(
...
+    :param length: "long", "short", or "narrow".
+    :param locale: a `Locale` object or a locale identifier.  Defaults to the
+                   system time locale.
+
+    .. versionadded:: 2.19
+    """
+    locale = Locale.parse(locale or LC_TIME)
+    date_fields = locale._data['date_fields']
+    key = field if length == 'long' else f'{field}-{length}'
+    data = date_fields.get(key) or date_fields.get(field)
```
tomasr8: So if the specified length does not exist, you get the long version by default?

tomasr8 (approved):

tests/test_dates.py
akx: Mmmm, I think this is related to #1076...

akx: Let's see how this changes after #1310 lands :)

PR #1303, opened 2026-07-30
Update messages' plural rules from CLDR data

akx:
Our `_GettextCompiler` emits suboptimal expressions, and the fix to generate more optimal ones isn't completely trivial, so this uses the ones generated by https://github.com/php-gettext/Languages instead.

Fixes #495
Fixes #711
Fixes #816
Fixes #817
Closes #1289 (duplicate; this one has more thorough coverage)
Closes #768 (duplicate; this one has more thorough coverage)

PR #1309, opened 2026-07-31
Update messages' plural rules from CLDR 48.1 data (correctly)

akx:
Follows up on #1303 to fix the script and re-add some mistakenly omitted entries.

I'll have to be more careful next time 😅


tests/messages/frontend/test_cli.py
```diff
@@ -341,7 +341,7 @@ def test_init_singular_plural_forms(cli):
 "Last-Translator: FULL NAME <EMAIL@ADDRESS>\n"
 "Language: ja_JP\n"
 "Language-Team: ja_JP <LL@li.org>\n"
-"Plural-Forms: nplurals=2; plural=(n != 1);\n"
+"Plural-Forms: nplurals=1; plural=0;\n"
```
akx, line 344: @jun66j5 Fixed back to the correct form here (re https://github.com/python-babel/babel/pull/1303#discussion_r3686731086). Thank you for checking! ❤️

PR #1310, opened 2026-07-31
Add support for other systemic aliases specified in CLDR `root.xml`

### LINKED ISSUE #1076: Add support for other systemic aliases specified in CLDR `root.xml`

akx:
We had some manual fallbacks for missing field/unit lengths, and ignored many `<alias>`es in CLDR data entirely.

This PR imports those aliases and uses them to resolve missing lengths, improving output in corner cases and fixes some rare KeyError crashes too.

There's a quirk related to person-variant unit aliases, which are only expressed in CLDR for the short length. If one follows the spec to the letter, a request for a long/narrow person-variant unit will route through the short length and lose the requested length.

We use the same "workaround" as ICU to preserve the length, by importing these as whole-unit aliases. Not totally spec-compliant, but the end result is simply better.

Fixes #1076.

### DIFF
--- a/babel/localedata.py
+++ b/babel/localedata.py
@@ -217,6 +217,8 @@ def merge(dict1: MutableMapping[Any, Any], dict2: Mapping[Any, Any]) -> None:
                 if val1 is None:
                     val1 = {}
                 if isinstance(val1, Alias):
+                    # A dict overriding an alias becomes an `(alias, overrides)` tuple
+                    # resolved in `Alias.resolve` or `LocaleDataDict.__getitem__`.
                     val1 = (val1, val2)
                 elif isinstance(val1, tuple):
                     alias, others = val1
@@ -260,6 +262,10 @@ class Alias:
         elif isinstance(data, tuple):
             alias, others = data
             data = alias.resolve(base)
+            if others:  # Apply overrides on a copy.
+                data = data.copy()
+                merge(data, others)
+
         return data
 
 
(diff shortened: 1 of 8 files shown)


babel/localedata.py
```diff
@@ -260,6 +262,10 @@ def resolve(self, data: Mapping[str | int | None, Any]) -> Mapping[str | int | N
         elif isinstance(data, tuple):
             alias, others = data
             data = alias.resolve(base)
+            if others:  # Apply overrides on a copy.
```
tomasr8, line 265: Sorry, I'm not that familiar with this part of the codebase - why is this needed? And should it be tested?

tomasr8 (approved):
Not super familiar with this, but nothing jumps out at me. The comments were very helpful though :)

babel/localedata.py
akx: `others` were simply ignored, which caused some chains of override/alias to fail (which are needed here).

There's an explicit test now :)

PR #1311, opened 2026-08-10
Don't crash on blank PO-Revision-Date/POT-Creation-Date headers

### LINKED ISSUE #1219: Exception raised if the revision date is blank

vjymisal0:
## Bug

Closes #1219

Some tools (e.g. Poedit) leave `PO-Revision-Date` or `POT-Creation-Date` blank in a `.po` file's header, instead of using the `YEAR-MO-DA HO:MI+ZONE` placeholder or omitting the header entirely.

`Catalog._set_mime_headers()` passes that value straight to `_parse_datetime_header()`, which does:

```python
```

With an empty string this raises:

```
```

which propagates all the way up and crashes `pybabel extract`/`update`/`compile` (see the traceback in #1219).

## Fix

Wrap both call sites (`pot-creation-date` and `po-revision-date`) in `contextlib.suppress(ValueError)`, so a blank or otherwise malformed date header is simply ignored (keeping whatever default/previous value the `Catalog` already had) instead of aborting the whole command. This matches the "ignore the blank date, or at least give a graceful message instead of crashing" request in the issue.

## Testing

- Added `test_catalog_set_mime_headers_ignores_blank_dates` in `tests/messages/test_catalog.py`, which reproduces the crash against the old code (fails with the reported `ValueError` before the fix) and passes after it.
- Ran the full test suite: `pytest tests/messages` — 353 passed.
- Ran `ruff check` on the changed files — no issues.

### DIFF
--- a/babel/messages/catalog.py
+++ b/babel/messages/catalog.py
@@ -13,6 +13,7 @@ from __future__ import annotations
 import datetime
 import re
 from collections.abc import Iterable, Iterator
+from contextlib import suppress
 from copy import copy
 from difflib import SequenceMatcher
 from email import message_from_string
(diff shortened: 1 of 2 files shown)


vjymisal0:
Hi maintainers! Just following up gently to see if anyone could take a look at this patch preventing crashes on blank PO/POT date headers. Tests are passing and branch is up to date with main. Thank you for your time and guidance!

PR #1313, opened 2026-08-12
Fix extract_messages dry-run behavior

### LINKED ISSUE #910: --dry-run does not work as expected
## Overview Description
We would like to add a command to our tests and check if we are frailer-free for message extractions.
found this `--dry-run` command on the [docs](https://babel.pocoo.org/en/latest/setup.html?highlight=--dry-run#extract-messages), but this does not work and I still see changes to my `.pot` file.

## Steps to Reproduce

1. See the CLI help:
```bash
$ ./setup.py extract_messages --help
Global options:
  --verbose (-v)  run verbosely (default)
  --quiet (-q)    run quietly (turns verbosity off)
  --dry-run (-n)  don't actually do anything
  --help (-h)     show detailed help message

Options for 'extract_messages' command:
   ...
```

2. extract with `--dry-run` command.
```bash
$ ./setup.py extract_messages --dry-run
```

## Actual Results

``` diff
# *.pot
-"POT-Creation-Date: 2022-10-12 11:37+0000\n"
+"POT-Creation-Date: 2022-10-14 11:06+0200\n"
```

## Expected Results
`don't actually do anything`  The `.pot` file should not have been updated or touched.

## Additional Information
Please let me know if we have forgotten to add some options, or if its a bug in the code, feel free to ask questions. and thanks in advance.

nightcityblade:
Fixes #910

The setuptools `extract_messages` command now honors the distutils dry-run flag and returns before creating or overwriting its output catalog. A regression test verifies that an existing POT file remains unchanged.

Tests:
- `pytest -q tests/messages` (353 passed)
- `uvx ruff==0.14.10 check babel/messages/setuptools_frontend.py tests/messages/test_setuptools_frontend.py`

### DIFF
--- a/babel/messages/frontend.py
+++ b/babel/messages/frontend.py
@@ -488,7 +488,7 @@ class ExtractMessages(CommandMixin):
 
     def run(self):
         mappings = self._get_mappings()
-        with open(self.output_file, 'wb') as outfile:
+        with self._open_output_file() as outfile:
             catalog = Catalog(
                 project=self.project,
                 version=self.version,
@@ -549,6 +549,9 @@ class ExtractMessages(CommandMixin):
                 sort_output=self.sort_output,
                 width=self.width,
             )
+
+    def _open_output_file(self) -> BinaryIO:
+        return open(self.output_file, 'wb')
 
     def _get_mappings(self):
         mappings = []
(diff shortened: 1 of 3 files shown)


akx (changes requested):
Erm. Doesn't this mean that passing `--dry-run` is equivalent to not calling the command at all? That's not the point.

nightcityblade:
Thanks — you're right. I reworked the implementation so dry-run now performs mapping, extraction, and serialization into an in-memory buffer while leaving the output catalog untouched. The regression test also asserts that extraction actually occurred. The full messages test suite passes (353 tests), and Ruff passes on all changed files.

PR #1161, opened 2024-12-09
Implementation of msgcat and msgmerge utilities from GNU gettext

soft-suroleb:
#### pybabel concat

- Allows concatenating multiple .po files into one.
- In case of translation conflicts, the translation from the first file is taken, as if the use-first option is set to true. Flags, locations, and other information in messages are combined.
- Implemented options:
  - output-file
  - less-than
  - more-than
  - unique
  - use-first
  - no-location
  - width
  - no-wrap
  - sort-output
  - sort-by-file

#### pybabel merge

- Enables merging files using a compendium for translation memory.
- The compendium can be used in two modes:
  - Default mode: Translations from the compendium are taken if absent in the output file.
  - Compendium overwrite mode: With the compendium-overwrite option, translations in the compendium are considered primary and overwrite those in the output file. If a translation is taken from the compendium, a comment is added specifying the source.
- Implemented options:
  - input-files
  - compendium
  - compendium-overwrite
  - no-compendium-comment
  - update
  - output-file
  - backup
  - suffix
  - no-fuzzy-matching
  - no-location
  - width
  - no-wrap
  - sort-output
  - sort-by-file

PR #1316, opened 2026-08-13
Concatenate/merge followups

akx:
Follows up on #1161 with some bugfixes and clarifications.


akx:
cc @soft-suroleb, please take a look :)

PR #1321, opened 2026-08-17
Downgrade TranslationError to warning(frontend.py) in compile_catalog

goodluck-ry:
Hi! Based on your wish, I modified frontend.py to fix this: python_format false positives no longer cause the command to exit with status 1 — instead they're logged as a warning, and the command exits 0 as normal.

One thing to flag though: since this downgrades TranslationError as a whole, genuine placeholder mismatches (not just false positives) will now also exit with status 0. This means they might not get caught automatically anymore.


goodluck-ry:
Hi！I've opened  a PR to fix this: #1321

akx:
You've linked to this same PR in your comment.

What is the "this" is this meant to fix? Whose wish are you referring to?

The thing you flagged: isn't that pretty critical?

goodluck-ry:
> You've linked to this same PR in your comment.
> 
> What is the "this" is this meant to fix? Whose wish are you referring to?
> 
> The thing you flagged: isn't that pretty critical?

Sorry for the confusion! This is my first time submitting a PR, so I messed up the links earlier. Allow me to clarify:

1. Link issue: I've now updated the main description to correctly link the original discussion post.
2. Context on "wish": The "wish" mentioned in the original post was to downgrade the `babel compile` error status to a warning so builds wouldn't fail on false positives.
3. Regarding your point on critical safety: You're completely right. Simply downgrading all errors (Option A) is too brute-force and unsafe because it masks real `TranslationError`s.

I will go with Option B: target only the specific "false positive" scenarios (like unmatched `% d` patterns) to issue a warning, while keeping standard `TranslationError`s as status 1 so real bugs aren't missed. 

I'll refactor the code and update this PR shortly!

goodluck-ry:
Closing in favor of #1325 as this refactored version fixes the root cause in catalog.py.

PR #1324, opened 2026-08-17
Report extra PO plural forms without changing parse

### LINKED ISSUE #1323: `num_plurals` checker is toothless on parsed pofiles
I was looking at `test_checkers.py`, and noticed that if I just mess with the pofile templates, e.g. add `msgstr[67]`, the tests would happily pass.

Looks like 7dfe6f9e01a4c157f69b6495ca1f293def3acc76 added the "msg has more translations than num_plurals of catalog" check in pofile reading - the additional translations have been skipped in parsing, so the checker doesn't do anything for catalogs parsed from pofiles...

adarshx01:
Fixes #1323.

`read_po` still drops `msgstr[N]` entries where `N >= nplurals` and pads `message.string` to `catalog.num_plurals`, same as it has for a decade. Extra forms are counted on `message.discarded_plural_forms` so `num_plurals()` can fail on a parsed catalog instead of always seeing a matching length.

### DIFF
--- a/babel/messages/catalog.py
+++ b/babel/messages/catalog.py
@@ -192,6 +192,9 @@ class Message:
             self.previous_id = []
         self.lineno = lineno
         self.context = context
+        # msgstr[N] entries dropped while reading a PO file because
+        # N was >= catalog.num_plurals. Does not affect message.string.
+        self.discarded_plural_forms = 0
 
     def __repr__(self) -> str:
         return f"<{type(self).__name__} {self.id!r} (flags: {list(self.flags)!r})>"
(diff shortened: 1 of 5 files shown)


akx:
This would change how Babel has parsed files for a decade. #1323 didn't yet include analysis as to what to do about the issue, just that it exists.

adarshx01:
Fair point — dropping extra `msgstr[N]` on read has been the parse behaviour for a long time, and #1323 only showed that the checker is a no-op after that, not what to do about it.

I went this way because `num_plurals` already runs on `Catalog` objects, so extra forms have to survive `read_po` for the check to mean anything. Happy to close this if you'd rather keep the historical parse and treat extras only at check time, or leave the checker as-is.

adarshx01:
Pushed a follow-up that keeps the historical parse.

`read_po` still drops `msgstr[N]` when `N >= nplurals` and still pads `message.string` to `catalog.num_plurals`, so catalogs and `write_po` look the same as they have for a decade. The extra entries are counted on `message.discarded_plural_forms` so `num_plurals()` can report them instead of always seeing a matching length after the drop.

Issue #865, opened 2022-04-26
Issues with PYTHON_FORMAT pattern and the comment blocks


hoangduytranuk:
## Overview Description
As mentioned in this bug report [Message.locations duplicate unnecessary #10104](https://github.com/sphinx-doc/sphinx/issues/10104), I found several problem with the current babel code, specifically with PYTHON_FORMAT pattern, mentioned in the [catalog.py](https://github.com/python-babel/babel/blob/master/babel/messages/catalog.py)  and the way that comments are written, as mentioned in the [pofile.py](https://github.com/python-babel/babel/blob/master/babel/messages/pofile.py). 
1. The PYTHON_FORMAT pattern recognise lines with percentage signs as a python-format and wrongly flagging the message being a python-format, but when loading into editors such as POEdit, the application shown that this is WRONG. I'm translating for Blender.org at the moment and this has caused so much time correcting the file that I ended up taking a look into the Python code and debugging it to find out the reason why. It turned out to be that this pattern, PYTHON_FORMAT, in the [catalog.py](https://github.com/python-babel/babel/blob/master/babel/messages/catalog.py) is currently not smart enough and erroneously creating havoc on the read/write of the PO file. This all started due to the fact I needed to write my own merging tool using Python, rather than relying on the buggy xgettext's msgmerge. 
2. The comment blocks are wrapping with the message undesirably, making changes to the source file when there are NO changes whatsoever has been made by the translator. Very annoying, especially when you are observing changes in the source file.

## Steps to Reproduce

1. Down load the PO translation file from the Blender github ([vi.po](https://github.com/blender/blender-translations/blob/master/po/vi.po)). This file is the one I'm working on at the moment in my own repository ([blender_ui](https://github.com/hoangduytran/blender_ui))
2. Make a simple python read/write this file to a new one, and observe the result:
```
from sphinx_intl import catalog as c
home = os.environ['HOME']
file_path = os.path.join(home, 'vi.po')
out_file = os.path.join(home, 'test_vi.po')
data = c.load_po(file_path)
c.dump_po(out_file, data, line_width=4096)    # simulate the --no-wrap flag on the msgmerge of xgettext
```
3. Open the test_vi.po file in POEdit and you will see right at the top lines being flagged as WRONG use of python-format.
4. Notice also that comment lines above the msgid are also WRAPPED.

## Expected Results
1. NO Python format ERRORS in the POEdit
[...]

PR #1325, opened 2026-08-17
fix(catalog): fix PYTHON_FORMAT regex false positives for percentage text

### LINKED ISSUE #1268: Reduce `compile` translation checks from ERROR to WARNING
# Problem

Some translation warnings are false positives.
E.g. `10% of` will be recognised as a placeholder (`% d`). If the translation is `10% der`, babel `compile` will say `incompatible format for placeholder 1: 'o' and 'd' are not compatible`.

This is then raised as `error` and the command exits with `status: 1`. This prevents e.g. running the command during docker image build.

# Wish

Even though the exit status is 1 the command succeeds in correctly compiling the `.po` to a valid `.mo` file.
The compile command should therefore exit with status 0 and only issue a `warning` about the problematic translations.

# Workaround

Append `; exit 0` to the command to force success.

goodluck-ry:
Fixes #1268 
Supersedes #1321

Hi @akx,

I've reverted the changes in frontend.py and refactored the PR to focus entirely on the regex rule in catalog.py.

The root cause of the false positive (such as '10% of' breaking builds) was that line 72 of catalog.py matched the space following '%' as a Python format flag in PYTHON_FORMAT. Removing the space flag from the regex prevents strings like '10% of' from being recognized as placeholders in the first place.

The original issue suggested downgrading errors from ERROR to WARNING. However, applying a downgrade in frontend.py would silently swallow real format mismatches. Fixing the regex at the source ensures that true syntax errors are still safely caught and exit with status 1, while the false positives are eliminated without touching error severity.

Trade-off note: Since explicit space-padded specifiers (like '% d') and text like '10% der' are textually identical at the static regex level, removing the space flag means '% d' won't be flagged. However, space-padded specifiers are extremely rare in translations compared to percentage text, making this a clean and pragmatically minimal fix.

I have updated the branch with corresponding tests for this case. Let me know what you think!


Mukller (commented):
Thanks for tackling this — the \100 % done\ false positives are real and annoying.

I verified the change empirically against both regex versions:

| input | master | this PR |
|---|---|---|
| \100 % done\ (prose) | matched (FP) | no match |
| \50 % off\ (prose) | matched (FP) | no match |
| \\% d\ (valid placeholder) | matched | **no match** |
| \\% 5d\, \\% f\, \\% e\ | matched | **no match** |

The problem: the space character is a legitimate printf flag (\'% d' % 3 == ' 3'\, [Python docs](https://docs.python.org/3/library/stdtypes.html#printf-style-string-formatting)), so dropping \\\ \\\ from the flags class turns current false positives into false negatives for translations that legitimately use spaced placeholders.

A middle ground that keeps both cases working: allow the space flag only when explicit width/precision digits follow (bare \% <letter>\ is almost always prose):

\\\python
PYTHON_FORMAT = re.compile(
    r'''
    \\%
        (?:\\(([\\w]*)\\))?
        (
            (?:[-#0+]?(?:\\*|[\\d]+)?(?:\\.(?:\\*|\\[\\d]+))?)
            |
            (?:\\ +[\\d]+(?:\\.[\\d]+)?)
        )
        [hlL]?
        ([diouxXeEfFgGcrs%])
''',
    re.VERBOSE,
)
\\\

With this variant: \% done\ / \% off\ don't match, while \% 5d\, \% f\-style placeholders still do. Related open issue about PYTHON_FORMAT quirks: #865. Could you add test cases for both directions to the PR? Happy to re-review.

Mukller:
*(Reposting with fixed formatting — the review above got mangled by my tooling, apologies.)*

Thanks for tackling this — the `100 % done` false positives are real and annoying.

I verified the change empirically against both regex versions:

| input | master | this PR |
|---|---|---|
| `100 % done` (prose) | matched (FP) | no match ✅ |
| `50 % off` (prose) | matched (FP) | no match ✅ |
| `% d` (valid placeholder) | matched | **no match ❌** |
| `% 5d`, `% f`, `% e` | matched | **no match ❌** |

**The problem:** the space character is a legitimate printf flag (`'% d' % 3 == ' 3'`, see [Python docs](https://docs.python.org/3/library/stdtypes.html#printf-style-string-formatting)), so dropping `\ ` from the flags class turns current false positives into false negatives for translations that legitimately use spaced placeholders.

**A middle ground** that keeps both cases working — allow the space flag only when explicit width digits follow (a bare `% <letter>` is almost always prose):

```python
PYTHON_FORMAT = re.compile(
    r'''
    \%
        (?:\(([\w]*)\))?
        (
            (?:[-#0+]?(?:\*|[\d]+)?(?:\.(?:\*|[\d]+))?)
            |
            (?:\ +[\d]+(?:\.[\d]+)?)
        )
        [hlL]?
        ([diouxXeEfFgGcrs%])
''',
    re.VERBOSE,
)
```

With this variant: `% done` / `% off` don't match, while `% 5d`-style placeholders still do.

Related open issue about PYTHON_FORMAT quirks: #865.

Could you add test cases covering both directions (prose false positives *and* spaced valid placeholders) to the PR? Happy to re-review.

Mukller:
Self-correction before anyone wastes time on my earlier suggestion: I verified my proposed pattern empirically and it had a gap — the second branch required digits after the space flag, so `% .2f` (space + precision, no width) would **not** match, even though `'% .2f' % 1.5 == ' 1.50'` is valid Python.

Corrected alternative — space flag is allowed only when followed by width digits or a precision dot:

```python
(
    (?:[-#0+]?(?:\*|[\d]+)?(?:\.(?:\*|[\d]+))?)   # current form, minus space flag
    |
    (?:\ +(?:\.[\d]+|[\d]+(?:\.[\d]+)?))          # space flag requires explicit width/precision
)
```

Verified against: `100 % done` ✗ / `50 % off` ✗ (prose stays out), `% 5d` ✓ / `% .2f` ✓ / `% 5.2f` ✓ / `%s`, `%5d` ✓.

Remaining known limitation, which I think is fundamental rather than fixable: bare `% f` / `% e` (space flag + conversion, no width/precision) stay unmatched — they're indistinguishable from prose like `% done` at the regex level. Master matches them today, so this trade-off should be an explicit maintainer decision; if losing bare `% f` is unacceptable, the PR's current approach (drop the space flag entirely) vs. this middle ground vs. status quo are the three real options.

goodluck-ry:
> Self-correction before anyone wastes time on my earlier suggestion: I verified my proposed pattern empirically and it had a gap — the second branch required digits after the space flag, so `% .2f` (space + precision, no width) would **not** match, even though `'% .2f' % 1.5 == ' 1.50'` is valid Python.
> 
> Corrected alternative — space flag is allowed only when followed by width digits or a precision dot:
> 
> ```python
> (
>     (?:[-#0+]?(?:\*|[\d]+)?(?:\.(?:\*|[\d]+))?)   # current form, minus space flag
>     |
>     (?:\ +(?:\.[\d]+|[\d]+(?:\.[\d]+)?))          # space flag requires explicit width/precision
> )
> ```
> 
> Verified against: `100 % done` ✗ / `50 % off` ✗ (prose stays out), `% 5d` ✓ / `% .2f` ✓ / `% 5.2f` ✓ / `%s`, `%5d` ✓.
> 
> Remaining known limitation, which I think is fundamental rather than fixable: bare `% f` / `% e` (space flag + conversion, no width/precision) stay unmatched — they're indistinguishable from prose like `% done` at the regex level. Master matches them today, so this trade-off should be an explicit maintainer decision; if losing bare `% f` is unacceptable, the PR's current approach (drop the space flag entirely) vs. this middle ground vs. status quo are the three real options.

Thanks so much for the detailed analysis, empirical validation, and the refined regex logic!

I have updated catalog.py using the proposed middle-ground pattern (requiring explicit width/precision for space flags) and added comprehensive test coverage:

1. Code & Test Changes

catalog.py: Updated PYTHON_FORMAT to handle space-flagged specifiers only when followed by explicit width digits or a precision dot.

test_catalog.py: Added test_message_python_format_prose_percent to ensure prose like 100 % done, 50 % off, 10% of, and 10% der are ignored, while space-padded specifiers with explicit width/precision continue to match.

test_checkers.py: Added test cases ensuring translations with these prose strings pass validation cleanly.

2. Trade-offs & Edge Cases Observed
As discussed, static regex has fundamental limits with text homographs. For transparency, here are the remaining trade-offs with this pattern:

Bare space flags: As noted, bare % f or % d (without width/precision) remain unmatched to avoid catching prose.

Prose adjacent to digits: Phrases like 100 % 5off or offer % 5x will still match as % 5o / % 5x because 5 triggers the width rule followed by a valid format specifier character.

Dynamic width (% *d): Space-padded specifiers using * for dynamic width (e.g., % *d) are not matched by the explicit digit branch.

All local tests are passing cleanly. Thanks again for guiding this to a much cleaner solution! Ready for re-review whenever you have time.

Mukller:
Verified your updated branch locally — the middle-ground pattern behaves exactly as specified:

| input | expected | on branch |
|---|---|---|
| `100 % done` / `50 % off` / `10% of` / `10% der` | no match | ✓ |
| `% 5d` / `% .2f` / `% 5.2f` / `%s` / `%5d` | match | ✓ |
| `Name: %s, %d years` (mixed) | match | ✓ |
| bare `% f` (documented limitation) | no match | ✓ |

`pytest tests/messages/test_catalog.py -k python_format` — 3/3 green, including the new prose-regression test.

I also re-checked the checker-side tests you mentioned: prose strings pass validation cleanly while real format-specifier mismatches still raise. The trade-off table you wrote (bare `% f`/`% e` staying unmatched) is accurately stated — that's indeed the irreducible ambiguity at regex level, and leaving it as an explicit maintainer decision is the right call.

Nothing further from my side; the implementation matches everything we discussed.

goodluck-ry:
Hi @akx, just wanted to check in on this PR — it's been about three
weeks since the last update. In the meantime, @Mukller kindly did a
thorough review and verified the regex approach against both the
original false-positive cases and some edge cases (see the discussion
above) — really appreciate the help there.

Happy to make further changes whenever you get a chance to take a
look. Thanks for your time!

tests/messages/test_catalog.py
```diff
@@ -38,6 +38,18 @@ def test_message_python_format():
...
+def test_message_python_format_prose_percent():
+    # literal percent signs in prose must not be mistaken for placeholders
+    assert not catalog.PYTHON_FORMAT.search('100 % done')
+    assert not catalog.PYTHON_FORMAT.search('50 % off')
+    assert not catalog.PYTHON_FORMAT.search('10% of')
+    assert not catalog.PYTHON_FORMAT.search('10% der')
+    # space-flag placeholders with explicit width/precision are still valid
+    assert catalog.PYTHON_FORMAT.search('% 5d')
+    assert catalog.PYTHON_FORMAT.search('% .2f')
+    assert catalog.PYTHON_FORMAT.search('% 5.2f')
```
akx, line 50: Hmm. `"% d"` is a perfectly valid formatting specifier.

```python
>>> '100 % done' % 4.2
'100  4one'
```

Mukller:
LGTM. The regex restructure correctly separates the two valid space-flag patterns (% 5d with width, % .2f with precision) from prose usage (100 % done, 50 % off).

The split (?:[-#0+]?(?:\*|\d+)?(?:\.(?:\*|\d+))?) | (?:\ +(?:\.[\d]+|\d+(?:\.[\d]+)?)) is clean: first branch handles standard flags, second branch handles space-flag + width/precision only. Moving [hlL]? outside both alternatives avoids duplicating it.

Test coverage is thorough — prose percent, space-flag with width, space-flag with precision, space-flag with both.

Approved.

akx:
@Mukller Please see my comment above.

tests/messages/test_catalog.py
goodluck-ry: > Hmm. `"% d"` is a perfectly valid formatting specifier.
> 
> ```python
> >>> '100 % done' % 4.2
> '100  4one'
> ```

Fair point, your example does hold. I went with treating it as prose
because a bare space-flag placeholder with no width/precision is
quite rare in practice, while "100% done" / "50% off" style prose is
far more common — so it's a trade-off between the two.

I'd lean toward keeping the current behavior if that's acceptable to
you. If you'd rather minimize false negatives instead (at the cost of
reintroducing some false positives), I can look into a stricter
condition. Which direction do you prefer?

PR #1326, opened 2026-08-18
Add a --check mode to the extract command

### LINKED ISSUE #900: Check mode for `exctract` command
## Overview Description

For the `update` command there is a `--check` mode which is really nice to use in CI to verify that the translations are up to date. I'd also like a check mode for the `extract` command to verify if the `.pot` file is up to date.

ChrisJr404:
Closes #900.

The `update` command already has a `--check` mode that verifies catalogs are up to date without writing them, which is very handy in CI. This adds the equivalent for `extract`.

With `--check`, messages are extracted from the source as usual, but the output file is not written. Instead the freshly extracted catalog is compared against the existing `--output-file`, and a non-zero exit code (a `BaseError`) is raised if:

- the output file does not exist, or
- the extracted template would differ from what is on disk.

If nothing would change, the command logs that the file is up to date and returns normally.

### POT-Creation-Date

A fresh extraction always regenerates the `POT-Creation-Date` header, so a naive comparison would always report a difference. The check therefore ignores that header alone (it copies the existing file's creation date onto the freshly extracted catalog before comparing), mirroring how `update --check` neutralises the volatile revision date. Any real change to the messages or the other headers is still reported.

The comparison round-trips the extracted catalog through `write_po`/`read_po` with the same options that would be used to write it, so the check reflects exactly what would end up on disk.

### Usage

```console
$ pybabel extract --check -o messages.pot mypackage/
```

### Tests

Added `test_extract_check` and `test_extract_check_ignores_pot_creation_date` in `tests/messages/frontend/test_cli.py`, covering the up-to-date, changed, and missing-file cases, that the check never writes/creates the output file, and that a differing `POT-Creation-Date` alone does not trigger a failure. The docs for the `extract` command are updated as well.

`pytest tests/messages` passes (389 tests) and `ruff check` is clean.

### DIFF
--- a/babel/messages/frontend.py
+++ b/babel/messages/frontend.py
@@ -23,7 +23,7 @@ import tempfile
 import warnings
 from collections import Counter, defaultdict
 from configparser import RawConfigParser
-from io import StringIO
+from io import BytesIO, StringIO
 from typing import TYPE_CHECKING, Any, BinaryIO, Iterable, Literal
 
 if TYPE_CHECKING:
@@ -355,6 +355,10 @@ class ExtractMessages(CommandMixin):
          'header comment for the catalog'),
         ('last-translator=', None,
          'set the name and email of the last translator in output'),
+        ('check', None,
+         "don't write the output file, just check whether it is up to date. "
+         "Return code 0 means the file is up to date, return code 1 means "
+         "that it would be changed."),
     ]  # fmt: skip
     boolean_options = [
         'no-default-keywords',
@@ -364,6 +368,7 @@ class ExtractMessages(CommandMixin):
         'sort-output',
         'sort-by-file',
         'strip-comments',
+        'check',
     ]
     as_args = 'input-paths'
     multiple_value_options = (
@@ -407,6 +412,7 @@ class ExtractMessages(CommandMixin):
         self.ignore_dirs = None
         self.header_comment = None
         self.last_translator = None
+        self.check = False
 
     def finalize_options(self):
         if self.input_dirs:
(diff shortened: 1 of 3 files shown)


babel/messages/frontend.py
```diff
@@ -497,67 +503,100 @@ def callback(filename: str, method: str, options: dict):
...
-                include_lineno=self.include_lineno,
-                no_location=self.no_location,
-                omit_header=self.omit_header,
-                sort_by_file=self.sort_by_file,
-                sort_output=self.sort_output,
-                width=self.width,
-            )
+            self._write_catalog(outfile, catalog)
+
+    def _write_catalog(self, fileobj, catalog):
```
akx, line 563: Please add type annotations here.

babel/messages/frontend.py
```diff
@@ -497,67 +503,100 @@ def callback(filename: str, method: str, options: dict):
...
+            catalog,
+            include_lineno=self.include_lineno,
+            no_location=self.no_location,
+            omit_header=self.omit_header,
+            sort_by_file=self.sort_by_file,
+            sort_output=self.sort_output,
+            width=self.width,
+        )
+
+    def _run_check(self, catalog):
```
akx, line 575: Please add type annotations here.

babel/messages/frontend.py
```diff
@@ -497,67 +503,100 @@ def callback(filename: str, method: str, options: dict):
...
+        if not os.path.exists(self.output_file):
+            self.log.warning('PO template file %s does not exist.', self.output_file)
+            raise BaseError(f"POT file {self.output_file} is out of date.")
```
akx, line 578: Is it useful to consider a missing template file that would be newly written out-of-date (and fail on that)? What do you think are the semantics?

akx (changes requested):

PR #1327, opened 2026-08-24
Extract catalog compilation into a compile_message_catalog() function

### LINKED ISSUE #760: Extract the compile functionality into an API function
Catalogs can currently only be compiled using the command line. To allow applications to leverage Babel to compile translations themselves, the relevant code can be extracted from `babel.messages.frontend.extract_messages()` into its own API function, separating it from the command line.

ChrisJr404:
Closes #760.

This pulls the compile logic out of the `CompileCatalog` command into a plain `compile_message_catalog()` function, so applications can compile catalogs from their own code without shelling out to the CLI (that was the use case in the betty PR linked from the issue). The command just calls the function now, so the command line behavior is unchanged.

I kept the signature close to the compile options (directory/locale/domain, or an explicit input and output file), added tests covering the file and directory modes plus the error counting, and dropped a short note under the compile docs.


docs/cmdline.rst
```diff
@@ -70,6 +70,11 @@ If neither the ``input_file`` nor the ``locale`` option is set, this command
 looks for all catalog files in the base directory that match the given domain,
 and compiles each of them to MO files in the same directory.
 
+The same operation is available as a function, so applications can compile
+catalogs from their own code without shelling out to the command line:
+
+.. autofunction:: babel.messages.frontend.compile_message_catalog
```
akx, line 76: Doesn't sound like a great import path, does it? I'd move this one module up.

babel/messages/frontend.py
```diff
@@ -210,78 +210,128 @@ def finalize_options(self):
...
+    statistics: bool = False,
+    log: logging.Logger | None = None,
+) -> int:
+    """Compile message catalogs from PO files into binary MO files.
+
+    This performs the same work as the ``pybabel compile`` command, exposed so
+    that applications can compile catalogs without shelling out to the command
+    line. Either point it at a base ``directory`` (optionally narrowed to a
+    single ``locale``), or pass an explicit ``input_file`` and ``output_file``.
+
```
akx, line 269: I'm not sure this needs to refer to `pybabel compile` as such?

babel/messages/frontend.py
```diff
@@ -210,78 +210,128 @@ def finalize_options(self):
...
+    :param directory: the base directory containing the catalogs
+    :param locale: the locale to compile, or ``None`` to compile every locale
+                   found under ``directory``
+    :param domain: the catalog domain, or an iterable of domains, to compile
+    :param input_file: the PO file to read, instead of walking ``directory``
+    :param output_file: the MO file to write when ``input_file`` is given
+    :param use_fuzzy: also compile fuzzy translations
+    :param statistics: log translation statistics for each catalog
+    :param log: the logger to report progress to, defaulting to Babel's own
+    :return: the number of catalogs that had errors
```
akx, line 283: For programmatic use, this sounds rather useless. I'd rather like it to return a namedtuple, result dataclass, typeddict, something like that, with more detail than just an integer.

`statistics` should also go in there instead of get logged implicitly here.

The original code already returned a `catalogs_and_errors`; this sounds worse already...

babel/messages/frontend.py
```diff
@@ -210,78 +210,128 @@ def finalize_options(self):
...
+
+def compile_message_catalog(
+    directory: str | os.PathLike[str] | None = None,
+    locale: str | Locale | None = None,
+    domain: str | Iterable[str] = "messages",
+    input_file: str | os.PathLike[str] | None = None,
+    output_file: str | os.PathLike[str] | None = None,
+    use_fuzzy: bool = False,
+    statistics: bool = False,
+    log: logging.Logger | None = None,
```
akx, line 261: If you're introducing a programmatic API like this, most of these should probably be Explicit Is Better Than Implicit, so no defaults. The flags and logger should also preferably be kwarg-only.

Also, we may call `domain` `domains` here and not accept a single string, that way the code below doesn't need to `listify_value()`.

babel/messages/frontend.py
```diff
@@ -210,78 +210,128 @@ def finalize_options(self):
...
+    if log is None:
+        log = logging.getLogger('babel')
+
+    n_errors = 0
+    for current_domain in listify_value(domain):
+        triples = list(_get_catalogs_to_compile(
+            directory, locale, current_domain, input_file, output_file,
+        ))
+        if not triples:
+            raise OptionError(f'no message catalogs found for domain {current_domain!r}')
```
akx, line 294: This sounds like a FileNotFoundError that the command class can translate to an `OptionError`.

akx (changes requested):
Some initial change requests within.

PR #1328, opened 2026-08-25
core: Add get_official_territories()

### LINKED ISSUE #904: Provide a list of current ISO 3166 country codes?
More of a feature question/suggestion: would it be possible to generate and provide a list of [ISO 3166-2](https://www.iso.org/iso-3166-country-codes.html) country codes? Quick glance at `common/main/*.xml` ([cldr-common.zip](http://unicode.org/Public/cldr/41/)) would indicate that the 2-letter codes are provided as
```
ldml/identity/language@type
```
I’m just not familiar enough with the standard, but [Country/Region (Territory) Names](https://cldr.unicode.org/translation/displaynames/countryregion-territory-names) mentions that they’re related:

> Country and region names (referred to as Territories in the Survey Tool) may be used as part of [Language/Locale Names](https://cldr.unicode.org/translation/displaynames/languagelocale-names), […]

Or am I misinterpreting it?

The Babel Core already has a few hardwired country codes for language aliasing:

https://github.com/python-babel/babel/blob/33d1dd738af8da30c7f24efe6b76ff8f56d154fc/babel/core.py#L79-L88

Basically, what I’m suggesting is something like [python-iso3166](https://github.com/deactivated/python-iso3166) but generated from the CLDR data, and perhaps as simple strings.

ChrisJr404:
Closes #904.

Adds `get_official_territories()` to `babel.core`, returning a frozenset of the ISO 3166-1 alpha-2 territory codes. As mentioned in the issue, the data is pulled from the CLDR supplemental code mappings during the import, so it isn't a strict copy of the ISO list. I keep the two-letter codes that have a numeric ISO code, and drop the deprecated aliases and the user-assigned ranges, which lands on 250 codes.

A couple of territories the CLDR doesn't carry are missing, and a few exceptionally reserved codes like EU come along, so the docstring calls that out.


scripts/import_cldr.py
```diff
@@ -98,6 +98,12 @@ def _parse_currency_date(s):
     return tuple(map(int, parts + [1] * (3 - len(parts))))
 
 
+def _is_user_assigned_territory(code):
+    # ISO 3166-1 reserves these code ranges for private use, so they never
+    # denote an actual territory.
+    return code in ('AA', 'ZZ') or 'QM' <= code <= 'QZ' or 'XA' <= code <= 'XZ'
```
akx, line 104: Which ones of these appear in the CLDR, I'm curious?

akx (changes requested):
> A couple of territories the CLDR doesn't carry are missing, and a few exceptionally reserved codes like EU come along, so the docstring calls that out.

This seems unacceptable if the intent is to provide the official ISO 3166-1 territory list. _If_ this were called `get_cldr_territories()`, that'd be much better already.

However, then I'd like that this was `Refs #904`, not fixes.

PR #1329, opened 2026-08-25
Fix Catalog.is_identical() for messages with msgctxt

abhijeet117:
## Summary
Catalog.is_identical() returned False for any catalog containing a context-specific message, even when compared with itself, so pybabel update --check reported such catalogs as out of date on every run. The internal (msgid, msgctxt) keys were resolved through get(), which strips the context and looks the message up by msgid only. The lookup now uses the internal keys directly.

## Testing
Reproduced on master: a catalog with one pgettext-style message returned False from is_identical(cat). Added catalog tests covering self-comparison, equal catalogs, differing context, differing string and context-free cases; the context test fails without the fix. Full suite: 7828 passed with only two pre-existing timezone-name failures identical on a clean tree.

## Checklist
- [x] Bug reproduced before the fix
- [x] Root cause identified
- [x] Bug fixed
- [x] Tests passed


akx:
Please see #1308.

PR #1336, opened 2026-09-04
Consider square brackets when building call_stack

### LINKED ISSUE #1335: Python extractor treats dict arguments as literal strings
Consider the following string:
```
pgettext(context, dictionary["key"])
```

Both of the arguments are variables, so Babel's extractor should pick nothing. However, it will pick `"key"`.

denyshon:
- Resolves https://github.com/python-babel/babel/issues/1335

This PR adds `[` and `]` to values for which `call_stack.append` and `call_stack.pop` are called in `extract_python`.


akx (changes requested):
I'd love to see some tests, please!

denyshon:
@akx I've added a couple of test cases to the existing `test_invalid_filter`. Hopefully, this should cover the changes in this PR

PR #1337, opened 2026-09-08
Fix JavaScript extraction of escaped Unicode surrogate pairs

WenJing95:
JavaScript literals such as `gettext("\uD83D\uDE00")` currently extract two surrogate code points, producing an incorrect PO message ID and a `UnicodeEncodeError` when compiling an MO catalog. Combine valid UTF-16 pairs during string unquoting so escaped and literal characters produce the same message ID, while preserving unmatched surrogates.

Add regression coverage for ordinary strings, template strings, pair boundaries, and escaped backslashes.

Validation: `pytest tests/messages -q` (405 passed), `pre-commit run --all-files --show-diff-on-failure`, and an extraction → PO → MO translation round-trip.

### DIFF
--- a/babel/messages/extract.py
+++ b/babel/messages/extract.py
@@ -849,7 +849,12 @@ def extract_javascript(
             elif token.type in ('string', 'template_string'):
                 new_value = unquote_string(token.value)
                 if concatenate_next:
-                    last_argument = (last_argument or '') + new_value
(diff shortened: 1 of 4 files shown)


akx:
Hi! The JavaScript parser is pretty wonky in general. Did this stem from an actual issue you've had?

PR #1340, opened 2026-09-10
Accept C and POSIX in Locale.parse

### LINKED ISSUE #1236: Babel fails to handle C locale properly
The "C" locale is the only one that is guaranteed to be available on all systems, however Babel does not handle it correctly:

```
13:12:25:   File "/home/xxxx/venv/lib/python3.12/site-packages/babel/core.py", line 427, in parse
13:12:25:     raise UnknownLocaleError(input_id)
13:12:25: babel.core.UnknownLocaleError: unknown locale 'c'
```

This is problematic if a python program that uses Babel is running with the C locale set.

gyanu2507:
Fixes #1236.

`default_locale()` maps `C`, `POSIX`, and `C.UTF-8` to `en_US_POSIX`. `Locale.parse()` did not, so `Locale.parse("C")` raised `UnknownLocaleError: unknown locale 'c'`. That is the frame in the report.

`parse()` now applies the same mapping (case-insensitive, optional encoding suffix) before `parse_locale`. `en_US_POSIX` with a POSIX variant is unchanged.


babel/core.py
```diff
@@ -361,6 +361,12 @@ def parse(
         if not isinstance(identifier, str):
             raise TypeError(f"Unexpected value for identifier: {identifier!r}")
 
+        # C/POSIX is not a CLDR language. Same mapping default_locale uses.
+        posix_stem = identifier.split(".")[0].split("@")[0]
+        posix_stem = posix_stem.replace("-", "_")
```
akx: Why is this `replace` necessary? `"C"` and `"POSIX"` don't include dashes or underscores.

babel/core.py
```diff
@@ -361,6 +361,12 @@ def parse(
         if not isinstance(identifier, str):
             raise TypeError(f"Unexpected value for identifier: {identifier!r}")
 
+        # C/POSIX is not a CLDR language. Same mapping default_locale uses.
+        posix_stem = identifier.split(".")[0].split("@")[0]
```
akx: What's the `@` splitting for? The test case doesn't seem to exercise it, and in any case a `@` modifier would be in the postfix already split out by `.split(".")`.

Additionally, please use `.partition(".")` in cases like this for performance.

akx (changes requested):

gyanu2507:
Dropped the dash replace and the `@` split. `C` / `POSIX` never have those on the language stem, and the encoding is already off the first `.`. Switched that check to `partition(".")`.

akx (approved):
LGTM, thanks!

PR #1341, opened 2026-09-12
Ignore quoted literals when selecting date and time fields

Str0k:
`parse_date` and `parse_time` currently search the raw format pattern for field letters. Letters inside quoted text can therefore change the inferred field order or enable an AM/PM adjustment even though the pattern has no day-period field.

For example, `parse_date('2024 day 04 01', format="yyyy 'day' MM dd")` returns January 4 instead of April 1. Likewise, `parse_time('at 03:30 pm', format="'at' HH:mm 'pm'")` returns 15:30 instead of 03:30.

Use the existing pattern tokenizer to consider only field tokens when selecting date/time components. This also respects escaped apostrophes within quoted text. The existing numeric extraction and supported field set are unchanged.

Validation on Windows with Python 3.12 and the project's CLDR 48.2 import:

- Six regression cases fail before the fix and pass after it: quoted date fields, escaped apostrophes, quoted minute/second fields, and a quoted day-period marker.
- Full suite: 7,834 passed, 9 skipped, 2 expected failures.
- Ruff lint passes for both modified files.

AI assistance: Codex helped develop the patch and regression tests. The diff and verification results are provided for maintainer review.


akx (changes requested):
The pattern was just parsed (including tokenization) and now it's getting tokenized again? I don't quite like that idea.

Could you see if `parse_pattern()` could do this operation internally and keep the information on the cached parsed pattern?

PR #1343, opened 2026-09-12
extract: Warn when skipping an f-string with substitutions

### LINKED ISSUE #715: Pybabel should raise an error when it encounters f-strings

ColinHouse:
`_(f"Hello {name}")` cannot be translated, so extraction skips it. Since #915
replaced `eval` with `ast`, it is skipped silently: `pybabel extract` exits 0,
writes no message, and prints nothing. The string never reaches translators and
ships untranslated, with no signal to the author.

This warns instead, in the same format as the existing "Empty msgid" warning:

    app.py:2: warning: Skipping f-string with substitutions: it cannot be
    translated as written.  Extract a format string instead,
    e.g. _("Hello {name}").format(name=name).

There are two code paths, because f-string tokenisation changed in 3.12:

- before 3.12 the whole f-string is one `STRING` token and `_parse_python_string`
  returns `None`;
- from 3.12 (PEP 701) a substitution arrives as a non-`FSTRING_*` token and the
  f-string is dropped where `current_fstring_start` is reset.

Both are covered, and the suite was run on 3.11 and 3.12.

It warns rather than raising, as the issue body asks ("warn the user about the
incorrect usage instead of simply crashing"). Raising would break builds that
extract successfully today. Happy to make it an error if you prefer.

A constant f-string such as `_(f"spam" f"eggs")` is still extracted and does not
warn; an f-string outside a translation call is untouched.

The stale `# TODO: we could raise an error or warning when not all nodes are
constants` is replaced, since this implements it.

fixes #715


akx (approved):
LGTM, thanks!

ColinHouse:
Sorry — this closed itself because I deleted my fork by accident, minutes after your review. Nothing to do with the review.

Reopened as #1350 with the same commit (6f30522) and a byte-identical diff.

PR #1334, opened 2026-08-29
fix(issue #1219): Exception raised if the revision date is blank

arunsoman:
Fixes https://github.com/python-babel/babel/issues/1219

## What
autonomous fix via `atomic-forge fix` — CIE (code graph over MCP) localized the bug, generated a failing regression test, and forge's repair loop fixed the source against it.

I created a .po file with [Poedit](https://poedit.net/) and it left the PO-Revision-Date field blank in the header. pybabel (2.17.0) raises the following exception when updating or compiling the po file:

```
 File "C:\<>\Python312\Lib\site-packages\babel\messages\catalog.py", line 102, in _parse_datetime_header
    dt = datetime.datetime.strptime(match.group('datetime'), '%Y-%m-%d %H:%M')
         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "C:\<>\Python312\Lib\_strptime.py", line 554, in _strptime_datetime
    tt, fraction, gmtoff_fraction = _strptime(data_string, format)
                                    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "C:\<>\Python312\Lib\_strptime.py", line 333, in _strptime
    raise ValueError("time data %r does not match format %r" %
ValueError: time data '' does not match format '%Y-%m-%d %H:%M'
```

It would be nice to ignore the blank date, or at least, to give a graceful warning/error message.

## How it was verified
- regression test: `tests/test_forge_1219.py` (CIE-generated; fails on the pre-fix code, passes on the fix)
- repair rounds: 1  failures: 2 -> 0
- repaired file(s): babel/messages/catalog.py

---
🏛️ Generated end-to-end by [atomic-forge](https://github.com/arunsoman/atomic-forge) — an autonomous, test-driven issue→PR repair engine (`atomic-forge fix <issue-url>`).

PR #1269, opened 2026-04-27
Ignore blank PO revision dates

ShipItAndPray:
## Summary
- ignore blank `PO-Revision-Date` header values when loading catalog MIME headers
- preserve the existing placeholder revision date instead of trying to parse an empty string
- add a focused regression test for reading a `.po` header with a blank revision date

Closes #1219.

## Validation
- E1 baseline repro exists: yes
  - On `master`, `pofile.read_po()` raises `ValueError: time data '' does not match format '%Y-%m-%d %H:%M'` when the header contains `PO-Revision-Date:` with an empty value.
- E2 new regression test fails on baseline: yes
  - In a detached `origin/master` worktree with only the new test applied:
  - `python3 -m pytest tests/messages/test_pofile_read.py -k 'blank_revision_date' -q`
  - Result: `1 failed, 31 deselected`
- E3 regression test passes after fix: yes
  - `python3 -m pytest tests/messages/test_pofile_read.py -k 'blank_revision_date' -q`
  - Result: `1 passed, 31 deselected`
- E4 nearby targeted behavior passes: yes
  - Direct issue repro now returns a catalog with `revision_date == 'YEAR-MO-DA HO:MI+ZONE'`
  - `python3 -m pytest tests/messages/test_catalog.py -k 'datetime_parsing' -q`
  - Result: `1 passed, 43 deselected`
- E5 diff stays localized: yes
  - Changes are limited to `babel/messages/catalog.py` and `tests/messages/test_pofile_read.py`.

Note: I did not use the full `tests/messages/test_pofile_read.py` file as a validation gate because this raw source checkout does not have Babel's locale data bundle built, and many unrelated tests in that file fail before reaching this issue's code path.

PR #1258, opened 2026-02-21
Handle blank PO-Revision-Date and POT-Creation-Date headers

bysiber:
When a PO file has a blank `PO-Revision-Date` header (e.g. generated by Poedit), `_parse_datetime_header` crashes with a `ValueError` because it tries to parse an empty string with `strptime`.

```
ValueError: time data '' does not match format '%Y-%m-%d %H:%M'
```

This adds a `value.strip()` check before calling `_parse_datetime_header` for both `PO-Revision-Date` and `POT-Creation-Date` headers. When the value is blank, the catalog keeps its existing defaults instead of crashing.

Fixes #1219

PR #1347, opened 2026-09-17
Don't crash on a blank PO-Revision-Date/POT-Creation-Date header

### LINKED ISSUE #1219: Exception raised if the revision date is blank

cristianchiriac:
Fixes #1219.

## Bug

Some tools (e.g. Poedit) leave `PO-Revision-Date` or `POT-Creation-Date` blank in a `.po` file's header, instead of eliding the header or using the `YEAR-MO-DA HO:MI+ZONE` placeholder. `_parse_datetime_header()` passes that blank value straight to `datetime.strptime()`, which crashes:

```
```

This takes down `pybabel` entirely when updating or compiling such a file, as reported.

## Fix

Treat a blank (or whitespace-only) header value the same as an unset one: return `None` instead of raising. This matches the existing precedent a few lines up in the same function for the `Language` header ("if the header's value is an empty string, which is what some tools generate").

## Test plan

- Added `test_datetime_parsing_blank_value_returns_none`, parametrized over `''` and `'   '`.
- Reproduced the exact reported crash via `read_po()` on a `.po` file with blank `PO-Revision-Date`/`POT-Creation-Date` headers, confirmed it's gone after the fix (both attributes become `None`).
- `pytest tests/messages/test_catalog.py` — 21 passed (up from 19), same 25 pre-existing failures unrelated to this change (missing CLDR data files in this source checkout — `RuntimeError: The babel data files are not available`, identical failure list with and without this fix).
- `ruff check` / `ruff format --check` clean on the changed lines.


akx:
Please see https://github.com/python-babel/babel/pull/1222, https://github.com/python-babel/babel/pull/1258, https://github.com/python-babel/babel/pull/1269, https://github.com/python-babel/babel/pull/1293, https://github.com/python-babel/babel/pull/1311, #1334 for prior art and some discussion.

Was this PR prepared with AI assistance?

PR #1348, opened 2026-09-18
Preserve nonconsecutive repeated comment lines

### LINKED ISSUE #417: Don't remove duplicated comment lines.
When we extract messages, the long comment lines will be wrapped in 76 columns.  While this behavior, some common words can be duplicated on multiple lines.  But when we read a catalog, duplicated comment lines are removed automatically.  So the catalog before writing and after reading are different for each other.  I think that is a bug.

For example, if we extract `#. KEY: foo.bar.baz...<long words more>...` and `#. KEY: waldo.fred.plugh...<long words more>...` they can be spread on 4 lines by `write_po`:

```po
#. KEY:
#. foo.bar.baz...<long words more>...
#. KEY:
#. waldo.fred.plugh...<long words more>...
msgid "Example"
msgstr ""
```

Now we will read the example catalog by `read_po`.  It removes duplicated comment lines.  So `#. KEY:` appears just once instead of twice:

```python
>>> catalog = read_po('example.pot')
>>> msg = list(catalog)[0]
>>> print('\n'.join(msg.auto_comments))
#. KEY:
#. foo.bar.baz...<long words more>...
#. waldo.fred.plugh...<long words more>...
```

berkay-byte:
Closes #417.

Two long comments beginning with `KEY:` can wrap onto separate lines when written to a PO file. Reading that file removes the second `KEY:` because `Message.__init__()` deduplicates comments across the entire list.

Use `itertools.groupby` in `Message.__init__()` to collapse only consecutive identical comments. Repeated lines separated by other text keep their positions. Both extracted and translator comments follow this rule, including when messages are cloned. PO parsing continues to use the constructor normally.

Tests cover consecutive and separated repeats, empty comments, iterator inputs, cloning, obsolete PO entries, and wrapped-comment round trips. Existing catalog-merge deduplication remains unchanged.

Prepared with OpenAI Codex assistance.

Validation: 7,844 tests passed, 9 skipped, 2 expected failures on Python 3.14. Repository pre-commit checks passed for the changed files.


akx (changes requested):
While I don't know the rationale for the "only write unique comments" commit from nearly 20 years ago fdb346848ea3d0c8c2c3e60114919d395890007e, I think there may have been something to it.

To fix the issue in #417, namely

```
#. KEY:
#. foo.bar.baz...<long words more>...
#. KEY:
#. waldo.fred.plugh...<long words more>...
```
losing the second `KEY:`, I think a better patch would be to adjust `Message.__init__()` to keep only one copy of the same line repeated. What do you think?

PR #1353, opened 2026-09-19
Specify encoding when running tests

### LINKED ISSUE #1352: Some tests fail on Windows with UnicodeDecodeError
When running `make test`:
- `test_input_paths_is_treated_as_list` fails with `UnicodeDecodeError: 'charmap' codec can't decode byte 0x81 in position 617: character maps to <undefined>`
- `test_compendium_with_comment` fails with `UnicodeDecodeError: 'charmap' codec can't decode byte 0x81 in position 711: character maps to <undefined>`

That's because the tests expect UTF-8, but Windows defaults to ANSI. To prevent this from happening, we should explicitly specify the encoding when running tests.

denyshon:
- Resolves https://github.com/python-babel/babel/issues/1352


akx (approved):
This is fine. Btw, [UTF-8 Mode will become default in 3.15](https://peps.python.org/pep-0686/).

